PR #66, opened 2021-02-21
Add support for S3-compatible object storage.

### LINKED ISSUE #41: Add support for S3-compatible services
- Backblaze B2
- Minio
- GCS

benbjohnson:
## Overview

This commits adds support for non-AWS S3-compatible storage such as MinIO, Backblaze B2, & Google Cloud Storage (GCS). Other backends should also work but some code has been added to make URL-based configurations work more easily.

I also removed the region & endpoint from global config settings so that only the access key and secret key are available globally.

Fixes #41 

### MinIO

Start a MinIO server in a separate window:

```
docker run -p 9000:9000 minio/minio server /data
```

Then go to http://localhost:9000 and click on the "+" button and click "Create Bucket" and give the bucket a name such as "testbkt".

Set environment variables for the credentials:

```
export AWS_ACCESS_KEY_ID=minioadmin
export AWS_SECRET_ACCESS_KEY=minioadmin
```

Then you can replicate your database:

```
litestream replicate /path/to/db s3://testbkt.localhost:9000/db
```

The replica URL format for a local MinIO machine is `s3://<bucket>.localhost:9000/<remote-path>`.

To replicate to a non-localhost machine, you'll need to use the `litestream.yml` configuration file with the following configuration:

```yml
dbs:
  - path: /path/to/db
    replicas:
      - type: s3
        bucket: testbkt
        path: db
        endpoint: http://HOST:PORT
        forcePathStyle: true
```

### Backblaze B2

After you have signed up for an account, create a bucket (such as "testbkt"). B2 buckets must be globally unique so you will have to find a unique bucket name.

Set environment variables for the credentials. The B2 "key id" maps to the "AWS_ACCESS_KEY_ID" and  the "application key" maps to the "AWS_SECRET_ACCESS_KEY".

```
export AWS_ACCESS_KEY_ID=
export AWS_SECRET_ACCESS_KEY=
```

Then you can replicate your database:

```
litestream replicate /path/to/db s3://testbkt.s3.us-west-000.backblazeb2.com/db
```

The replica URL format for B2 is `s3://<bucket>.<endpoint>/<remote-path>`.

You can also specify a B2 replica in the configuration file:

```yml
dbs:
  - path: /path/to/db
    replicas:
      - type: s3
        bucket: testbkt
        path: db
        endpoint: https://s3.us-west-000.backblazeb2.com
        forcePathStyle: true
```

### Google Cloud Storage (GCS)

After you have signed up for an account, create a bucket (such as "testbkt"). GCS buckets must be globally unique so you will have to find a unique bucket name. Buckets that use dots may also require you to verify your domain with Google.

You will need to go to `Storage > Settings > Interoperability` in the Google Cloud console and create an access key for your service account. This should return an access key & a secret key. Set those in your environment variables:

```
export AWS_ACCESS_KEY_ID=
export AWS_SECRET_ACCESS_KEY=
```

Then you can replicate your database:

```
litestream replicate /path/to/db s3://testbkt.storage.googleapis.com/db
```

The replica URL format for GCS is `s3://<bucket>.storage.googleapis.com/<remote-path>`.

You can also specify a GCS replica in the configuration file:

```yml
dbs:
  - path: /path/to/db
    replicas:
      - type: s3
        bucket: testbkt
        path: db
        endpoint: https://storage.googleapis.com
        forcePathStyle: true
```

### DIFF
--- a/cmd/litestream/databases.go
+++ b/cmd/litestream/databases.go
@@ -22,7 +22,7 @@ func (c *DatabasesCommand) Run(ctx context.Context, args []string) (err error) {
 	if err := fs.Parse(args); err != nil {
 		return err
 	} else if fs.NArg() != 0 {
-		return fmt.Errorf("too many argument")
+		return fmt.Errorf("too many arguments")
 	}
 
 	// Load configuration.
@@ -40,7 +40,7 @@ func (c *DatabasesCommand) Run(ctx context.Context, args []string) (err error) {
 
 	fmt.Fprintln(w, "path\treplicas")
 	for _, dbConfig := range config.DBs {
-		db, err := newDBFromConfig(&config, dbConfig)
+		db, err := NewDBFromConfig(dbConfig)
 		if err != nil {
 			return err
 		}
--- a/cmd/litestream/generations.go
+++ b/cmd/litestream/generations.go
@@ -35,7 +35,7 @@ func (c *GenerationsCommand) Run(ctx context.Context, args []string) (err error)
 	var r litestream.Replica
 	updatedAt := time.Now()
 	if isURL(fs.Arg(0)) {
-		if r, err = NewReplicaFromURL(fs.Arg(0)); err != nil {
+		if r, err = NewReplicaFromConfig(&ReplicaConfig{URL: fs.Arg(0)}, nil); err != nil {
 			return err
 		}
 	} else if configPath != "" {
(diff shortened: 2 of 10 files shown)


oderwat:
Using a global or local defined `access-key-id` and `secret-access-key` in the YAML file does not work for me. I get `NoCredentialProviders: no valid providers in chain`. Using them as ENV variables on the command-line works. But this is not what I think of being secure at all.

benbjohnson:
@oderwat Thanks for taking a look and reporting the issue. Looks like I had environment variables set while I was doing testing which masked the problem. I fixed it in https://github.com/benbjohnson/litestream/pull/70 and added some tests. Let me know if you're still seeing an issue.

oderwat:
I just updated to the current master and tried it. Works as expected. Thank you for the quick response!

angelbarrera92:
Is there any release date planned to release this feature?
Thanks!

benbjohnson:
I'm targeting next week to cut a release for v0.3.3.

PR #84, opened 2021-02-25
Add snapshot interval

### LINKED ISSUE #3: Snapshot Interval
Currently, litestream performs a snapshot when the retention period rolls over but this means that a long retention period can extend restore times. It would be useful to separate the retention interval from the snapshot interval so a user can have multiple snapshots within their retention period (e.g. snapshot every day, keep for one week).

benbjohnson:
## Overview

This commit adds the ability to periodically perform snapshots on an interval that is separate from retention. For example, this lets you retain backups for 24 hours but you can snapshot your database every six hours to improve recovery time.

This also fixes a bug where the retention check interval can be longer than the retention time which causes retention to not be enforced often enough.

Fixes #3


benbjohnson:
There is code duplication between `FileReplica` & `s3.Replica` but I'm deferring a refactoring until the HTTP replica is in place.

PR #242, opened 2021-10-10
Reduce snapshot check frequency

### LINKED ISSUE #234: ListBucket is 99% of our AWS cost
Can we cache some query and save some $ ?

| Opération            | sept. 2021 | 
|----------------------|------------|
| Total cost ($)               | 6.37 | 
| ListBucket ($)               |  6.12 | 
| CreateVolume-Gp2 ($) |  0.21 | 
| StandardStorage  ($) |  0.02 |
| PutObject  ($) |  0.01 | 

Let us know.

benbjohnson:
Previously, a bug was introduced that added a `LIST` operation on every replica sync which significantly increased the cost of running Litestream against S3. This changes the behavior to only issue the `LIST` operation when the generation has changed.

Fixes #234

### DIFF
--- a/replica.go
+++ b/replica.go
@@ -159,22 +159,21 @@ func (r *Replica) Sync(ctx context.Context) (err error) {
 
 	Tracef("%s(%s): replica sync: db.pos=%s", r.db.Path(), r.Name(), dpos)
 
-	// Create snapshot if no snapshots exist for generation.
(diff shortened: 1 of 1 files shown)


benbjohnson:
Ran this for a couple days to do a sanity check and it seems to work well. 👍

PR #267, opened 2022-01-15
Add end-to-end replication/restore testing

benbjohnson:
This commit adds automated tests against the compiled `litestream` CLI for the following test cases:

- Simple, multi-index replication
- Resuming Litestream with the same generation
- Resuming Litestream with a new generation
- High load replication
- Long-running unit test (5m)

Additional longer term testing (2 weeks) is the next step and will be introduced in future pull requests.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -29,6 +29,9 @@ jobs:
         env:
           LITESTREAM_SFTP_KEY: ${{secrets.LITESTREAM_SFTP_KEY}}
 
+      - name: Build binary
+        run: go install ./cmd/litestream
+
       - name: Run unit tests
         run: make testdata && go test -v ./...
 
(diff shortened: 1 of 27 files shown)


.github/workflows/test.yml
```diff
@@ -53,10 +56,13 @@ jobs:
...
-          LITESTREAM_SFTP_USER:     ${{ secrets.LITESTREAM_SFTP_USER }}
-          LITESTREAM_SFTP_KEY_PATH: /opt/id_ed25519
-          LITESTREAM_SFTP_PATH:     ${{ secrets.LITESTREAM_SFTP_PATH }}
+      #- name: Run sftp tests
+      #  run: go test -v -run=TestReplicaClient ./integration -replica-type sftp
+      #  env:
+      #    LITESTREAM_SFTP_HOST:     ${{ secrets.LITESTREAM_SFTP_HOST }}
+      #    LITESTREAM_SFTP_USER:     ${{ secrets.LITESTREAM_SFTP_USER }}
+      #    LITESTREAM_SFTP_KEY_PATH: /opt/id_ed25519
+      #    LITESTREAM_SFTP_PATH:     ${{ secrets.LITESTREAM_SFTP_PATH }}
```
benbjohnson, line 65: I accidentally deleted the server that held the SFTP endpoint. I'm rebuilding this as part of the next effort of longer term testing.

db.go
```diff
@@ -1731,3 +1731,12 @@ func headerByteOrder(hdr []byte) (binary.ByteOrder, error) {
...
 }
+
+// logPrefixPath returns the path to be used for logging.
+// The path is reduced to its base if it appears to be a temporary test path.
+func logPrefixPath(path string) string {
+	if strings.Contains(path, "TestCmd") {
+		return filepath.Base(path)
+	}
+	return path
+}
```
benbjohnson, line 1742: This helps make the testing log output sane.

internal/locking_buffer.go
```diff
@@ -0,0 +1,145 @@
...
+	"bytes"
+	"io"
+	"sync"
+)
+
+// LockingBuffer wraps a bytes.Buffer with a mutex.
+type LockingBuffer struct {
+	mu sync.Mutex
+	b  bytes.Buffer
+}
```
benbjohnson, line 13: The `LockingBuffer` was added to handle reading STDOUT/STDERR while the external `litestream` process was writing to it at the same time.

litestream.go
```diff
@@ -43,7 +45,7 @@ var (
 
 var (
 	// LogWriter is the destination writer for all logging.
-	LogWriter = os.Stderr
+	LogWriter = os.Stdout
```
benbjohnson, line 48: Updating all logging output to STDOUT.

litestream.go
```diff
@@ -460,6 +462,12 @@ func isHexChar(ch rune) bool {
 	return (ch >= '0' && ch <= '9') || (ch >= 'a' && ch <= 'f')
 }
 
+// md5hash returns a hex-encoded MD5 hash of b.
+func md5hash(b []byte) string {
+	sum := md5.Sum(b)
+	return hex.EncodeToString(sum[:])
+}
+
```
benbjohnson, line 470: I'm leaving this helper in as it is useful when debugging page mismatches.

replica.go
```diff
@@ -43,7 +43,7 @@ type Replica struct {
 	cancel func()
 
 	// Client used to connect to the remote replica.
-	Client ReplicaClient
+	client ReplicaClient
```
benbjohnson, line 46: The `ReplicaClient` no longer has a reference back to its parent `Replica` so it can be set on instantiation and left unexported. Previously, the file replica client required a reference to the DB so it could determine file permissions.

Issue #276, opened 2022-01-23
caused by: InvalidArgument: Part number must be an integer between 1 and 1000, inclusive


anacrolix:
I suspect my DB is too large (5x1000=5GB?), my DB is 36 GB, running replicate I get:
```
litestream initialization complete
some-file(s3): monitor error: MultipartUpload: upload multipart failed
        upload id: x
caused by: InvalidArgument: Part number must be an integer between 1 and 1000, inclusive
        status code: 400, request id: x host id: x
```

benbjohnson:
Hmm, that's a new one. Yeah, it looks like your DB is too big. :)

Looks like the `PartSize` and `MaxUploadParts` on the [`s3manager.Uploader`](https://docs.aws.amazon.com/sdk-for-go/api/service/s3/s3manager/#Uploader) needs to be set within the S3 client. However, the SDK docs say the default [`MaxUploadParts`](https://github.com/aws/aws-sdk-go/blob/v1.42.39/service/s3/s3manager/upload.go#L20-L22) is 10,000 so I'm not sure why it's showing `1000` in the error.

It'd be nice to set this automatically although Litestream runs the snapshot through an LZ4 compressor so we don't know the final size. We could probably estimate it. Or it could just be a config option since most folks probably aren't running 36GB databases.

danp:
@anacrolix Just double checking: you are connecting to S3 proper and not some other S3-compat service? Searching for `Part number must be an integer between 1 and 1000`, I mostly see mentions of Scaleway.

anacrolix:
@danp thank you for picking this. Yes I am using Scaleway! @benbjohnson how can I modify it so that the parts are bigger? I guess my max parts are 1000 for scaleway? Is there somewhere that is set so it can adjust the part size to suit?

benbjohnson:
@anacrolix Litestream doesn't currently support setting the part size. It shouldn't be an especially difficult change though:

1. Add field to [`ReplicaConfig`](https://github.com/benbjohnson/litestream/blob/39114502f3959cf6ddf068148c4ff92e450dd831/cmd/litestream/main.go#L331-L338)
2. Pass it to the `ReplicaClient` in [`newS3ReplicaFromConfig()`](https://github.com/benbjohnson/litestream/blob/39114502f3959cf6ddf068148c4ff92e450dd831/cmd/litestream/main.go#L485-L495)
[...]

PR #284, opened 2022-01-25
Alter S3 uploader config for large uploads to Scaleway

anacrolix:
Implements https://github.com/benbjohnson/litestream/issues/276#issuecomment-1020785492.

### DIFF
--- a/s3/replica_client.go
+++ b/s3/replica_client.go
@@ -100,6 +100,8 @@ func (c *ReplicaClient) Init(ctx context.Context) (err error) {
 	}
 	c.s3 = s3.New(sess)
 	c.uploader = s3manager.NewUploader(sess)
+	c.uploader.MaxUploadParts = 1000
+	c.uploader.PartSize = 100 << 20
 	return nil
 }
 


benbjohnson:
@anacrolix I think this needs to be configurable as I'm not sure the performance effects of adjusting the part size and count.

benbjohnson (changes requested):
@anacrolix I think this needs to be configurable as I'm not sure the performance effects of adjusting the part size and count.

benbjohnson:
@anacrolix I'm going to close this PR out for now. Let me know if you want it reopened. 👍

anacrolix:
Perhaps it's worth trying with this as it has broader compatibility, and working back the other way if issues are found? For example a configuration option could instead increase the allowed part count and decrease the size.

benbjohnson:
Yeah, I'm not opposed to changing the defaults but I think it'll need testing to understand the implications. Also, the new settings in this PR put the part size at 100MB and IIRC the uploader defaults to a concurrency of 8. I'm not sure what the heap requirements are per part but I'm a little concerned about making it OOM on small VM instances.

PR #289, opened 2022-01-29
Enable test runner approval workflow

benbjohnson:
This pull request adds a "test" environment to GitHub Actions so that the test runner deployment can be behind a deployment approval. This is so the test runner does not need to be deployed for CI changes or non-code related changes.


benbjohnson:
Trying to get this flow to work is a PITA. I'm closing this and the test runner can just be run on every PR instead.

PR #303, opened 2022-02-06
Add WAL overrun validation

benbjohnson:
Under high write load, it is possible for write transactions from another process to overrun the WAL between the time when Litestream performs a RESTART checkpoint and when it obtains the write lock immediately after. This change adds validation that an overrun has not occurred and, if it has, it will start a new generation.

### DIFF
--- a/db.go
+++ b/db.go
@@ -34,7 +34,7 @@ const (
 
 // MonitorDelayInterval is the time Litestream will wait after receiving a file
 // change notification before processing the WAL file for changes.
-const MonitorDelayInterval = 100 * time.Millisecond
+const MonitorDelayInterval = 10 * time.Millisecond
 
 // MaxIndex is the maximum possible WAL index.
 // If this index is reached then a new generation will be started.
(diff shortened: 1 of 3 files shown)


db.go
```diff
@@ -422,14 +422,20 @@ func (db *DB) Close() (err error) {
...
 				err = e
 			}
 		}
-		if e := r.Stop(true); e != nil && err == nil {
+
+		// Close out replica.
+		if e := r.Close(); e != nil && err == nil {
 			err = e
 		}
 	}
```
benbjohnson, line 441: Previously, this tried to sync before stopping the replica but that could cause the replica to sync twice at the same time. Now it stops, syncs, & closes.

db.go
```diff
@@ -795,10 +801,25 @@ func (db *DB) createGeneration(ctx context.Context) (string, error) {
...
+		if err := func() error {
+			db.mu.Lock()
+			defer db.mu.Unlock()
+			return db.sync(ctx)
+		}(); err != nil {
+			db.Logger.Printf("sync error, retrying: %s", err)
+		} else {
+			break
+		}
+	}
```
benbjohnson, line 817: We no longer check the database for changes every second so we are attempting retries instead.

PR #313, opened 2022-02-11
Add docker multiarch build and push to release

### LINKED ISSUE #310: Build docker image multiarch (armv7, arm64)
Love the project, it has made it possible to run a  lot of services on my k3s cluster! Thank you. 

Unfortunately the [docker image](https://hub.docker.com/r/litestream/litestream/tags) is only built for amd64, so it will only run on certain nodes.

If you haven't done it before, building container images cross platform is very easy. I would submit a PR if I could find the pipeline where you're doing it!  In a GitHub Actions workflow these are the relevant steps:

``` yaml
      - name: setup qemu
        uses: docker/setup-qemu-action@v1
      - name: setup buildx
        id: buildx
        uses: docker/setup-buildx-action@v1
      - name: Log in to the Container registry
        run: echo ${{ secrets.GITHUB_TOKEN }} | docker login ghcr.io -u ${{ github.actor }} --password-stdin
      - name: build the image and push
        run: |
          cd cluster/Dockerfiles/sidecar-sync && 
          docker buildx build \
          --push \
          --tag ghcr.io/${{ github.actor }}/sidecar-sync:multiarch \
          --platform linux/amd64,linux/arm64,linux/arm/v7 . 
```

You can see a complete example in [one of my own repos](https://github.com/ohthehugemanatee/homenet-documentation/blob/master/.github/workflows/build-sidecar-sync.yaml).

ohthehugemanatee:
I added the buildx to the end of the release pipeline. 

- I'm not 100% this is where you want it to happen. Why does this pipeline run on every PR? Maybe it's better to split this out.
- this won't work until the appropriate secrets and permissions are configured, and I don't have access to do that on your repo. Have a look at the [details in the doco](https://docs.github.com/en/packages/working-with-a-github-packages-registry/working-with-the-container-registry).
- There are other platforms available if you want to add to the list. I included a build step to list the currently available platforms. Here's the list at the moment: `linux/amd64,linux/arm64,linux/riscv64,linux/ppc64le,linux/s390x,linux/386,linux/mips64le,linux/mips64,linux/arm/v7,linux/arm/v6`

closes #310

### DIFF
--- a/.github/workflows/release.docker.yml
+++ b/.github/workflows/release.docker.yml
@@ -1,0 +1,43 @@
+on:
+  release:
+    types:
+      - published
+  pull_request:
+    types:
+      - opened
+      - synchronize
+      - reopened
+    branches-ignore:
+      - "dependabot/**"
+
+name: Release (Docker)
+jobs:
+  docker:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: docker/setup-qemu-action@v1
+      - uses: docker/setup-buildx-action@v1
+
+      - uses: docker/login-action@v1
+        with:
+          username: benbjohnson
+          password: ${{ secrets.DOCKERHUB_TOKEN }}
+      
+      - id: meta
+        uses: docker/metadata-action@v3
+        with:
+          images: litestream/litestream
+          tags: |
+            type=ref,event=branch
+            type=ref,event=pr
+            type=semver,pattern={{version}}
+            type=semver,pattern={{major}}.{{minor}}
+      
+      - uses: docker/build-push-action@v2
+        with:
+          context: .
+          push: true
+          platforms: linux/amd64,linux/arm64,linux/arm/v7
+          tags: ${{ steps.meta.outputs.tags }}
+          labels: ${{ steps.meta.outputs.labels }}
--- a/.github/workflows/release.linux.yml
+++ b/.github/workflows/release.linux.yml
@@ -1,7 +1,7 @@
 on:
   release:
     types:
-      - created
+      - published
   pull_request:
     types:
       - opened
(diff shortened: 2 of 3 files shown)


benbjohnson:
@ohthehugemanatee Thanks for the PR. Why use GHCR versus the Docker registry? I haven't used GHCR before.

ohthehugemanatee:
> Why use GHCR versus the Docker registry? I haven't used GHCR before.

No particular reason at the time, though now Docker enforces rules about frequency of access etc. you could just as easily push to docker hub.

ohthehugemanatee:
Looks like the SFTP test failed on the remote host being unavailable... Anything we can do about this?

benbjohnson:
@ohthehugemanatee Sorry, I pushed to PR #318 to make testing CI easier and I missed your messages on here before merge. I split out the `linux/amd64` and `linux/arm64` into separate builds since the `amd64` is way faster. I also added tags for short SHA, long SHA, and PR#. 

For example, the PR produced 3 tags:

- `litestream/litestream:pr-318`
- `litestream/litestream:sha-21cc863`
- `litestream/litestream:sha-21cc863eae60b16ff2b51cafbd2501525a121998`

I dropped `arm7` just to get it working. I can add it back in with a separate PR. 

> _ Docker enforces rules about frequency of access etc. you could just as easily push to docker hub._

I switched to Docker Hub since that's where the Litestream images currently reside. If folks have issues with access frequency then I can push to `ghcr.io` as well.

> _Looks like the SFTP test failed on the remote host being unavailable... Anything we can do about this?_

I'm not sure what was going on with that server. I just restarted it and it's working now. 🤷‍♂️

benbjohnson:
Added `linux/arm/v7` here: https://github.com/benbjohnson/litestream/pull/319

PR #348, opened 2022-04-01
Remove Defer

gsquire:
This patch removes the defer statement inside of the loop and instead closes the directory if an error occurs.

I'm not sure if you want to capture the error in this case too as opposed to the error that is causing the early return. I'm happy to change it if need be.

I found it by running [this tool](https://github.com/gsquire/dll).

### DIFF
--- a/file_replica_client.go
+++ b/file_replica_client.go
@@ -450,10 +450,10 @@ func (itr *FileWALSegmentIterator) Next() bool {
 			itr.err = err
 			return false
 		}
-		defer f.Close()
 
 		fis, err := f.Readdir(-1)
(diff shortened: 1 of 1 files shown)


benbjohnson:
@gsquire Sorry for taking so long to get back to you on this. I prefer to use `defer` and just call `Close()` twice since it's a no-op the second time. There's two reasons—first, it covers cases where a `panic()` would cause the function to exit early, and second, it's more future proof since I would have to remember to `Close()` if I added another early return in the future.

gsquire:
It's no problem! Your reasoning makes total sense.

PR #359, opened 2022-04-17
Calculate test coverage as part of build_and_test action

mtlynch:


benbjohnson:
Thanks, Michael. Good idea to generate a test coverage artifact. 👍

Issue #299, opened 2022-02-05
Discontinue Windows Support


benbjohnson:
Currently, Windows support for Litestream is minimal at best. Some testing was done early on but on-going testing has been subpar and I unfortunately don't have the time or the expertise to continue with it. Additionally, usage stats show minimal Windows usage with Litestream.

I'm open to re-adding support in the future if there is interest and there is someone dedicated to its support.

PR #369, opened 2022-05-06
Use fsnotify

mattn:
I removed the Linux/BSD file-watcher implementations and tests in this change. I also added code to use fsnotify. This change improves compatibility of litestrearm. Windows is also supported.

Update #299


benbjohnson:
@mattn I check the code but there was a build error (missing `internal.Fileinfo()`). I re-added that but it looks like there's test failures now.

mattn:
Fixed.

benbjohnson:
Thanks, @mattn!

PR #373, opened 2022-05-10
Support SAKURA internet Object Storage

hnakamur:
First of all, thanks for a wonderful software!

This pull request adds support for [object storage](https://manual.sakura.ad.jp/cloud/objectstorage/about.html) provided by [SAKURA internet Inc.](https://www.sakura.ad.jp/)

Disclaimer: I work at [SAKURA internet Inc.](https://www.sakura.ad.jp/), on [a CDN service named WebAccelerator](https://www.sakura.ad.jp/services/cdn/).


benbjohnson:
Hi @hnakamur, thanks for reaching out. I'm actually trying to avoid adding more custom URL handling in Litestream so I'm going to close this issue. There's a lot of S3-like providers and it just gets to be too much. I'd like to change it so that the URL format allows query parameters for the different fields instead.

PR #379, opened 2022-05-13
Add environment variables for scheme and forcePathStyle

hnakamur:
This pull request is needed for running a minio server at non-localhost server for example in docker compose.


benbjohnson:
I'm not sure I love the `panic()` but this works for now. Thanks!

hnakamur:
Thanks for merging!
I agree `panic` is not ideal here.
Maybe `parseHost` can changed to return an error in the future.

PR #380, opened 2022-05-13
Fix read replication stream restart position

benbjohnson:
Previously, a reconnecting replica would reset its position's "offset" to the beginning of a given WAL file in an effort to be more conservative about what was replayed. However, the client replays each segment one-at-at-time which causes the database to get in an inconsistent state during this replay since it is being replayed from the beginning instead of the current position. The SQLite sometimes detects this inconsistent state and prevents the replica WAL from checkpointing which caused all replication to stop.

This pull request removes this "offset reset" and simply replays from the last position.

### DIFF
--- a/db.go
+++ b/db.go
@@ -1444,6 +1444,8 @@ func (db *DB) readPositionFile() (Pos, error) {
 		return Pos{}, err
 	}
 
+	println("dbg/posfile", strings.TrimSpace(string(buf)))
+
 	// Treat invalid format as a non-existent file so we return an empty position.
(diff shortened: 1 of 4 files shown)


ddevienne:
Hi Ben. Was looking at the changes, and noticed you used system-calls for locking, thus bypassing the SQLite VFS.
I guess that implies Litestreams assumes the SQLite VFS must also use the system locking primitives?
And thus Litestreams is incompatible with any kind of custom VFS? Just thinking aloud.
(Are those system-calls for locking also available for Windows in Go?)

PR #389, opened 2022-05-30
Improve readability

ryanrussell:
Signed-off-by: Ryan Russell <ryanrussell@users.noreply.github.com>

### DIFF
--- a/wal_downloader.go
+++ b/wal_downloader.go
@@ -15,7 +15,7 @@ import (
 // WALDownloader represents a parallel downloader of WAL files from a replica client.
 //
 // It works on a per-index level so WAL files are always downloaded in their
-// entiretry and are not segmented. WAL files are downloaded from minIndex to
+// entirety and are not segmented. WAL files are downloaded from minIndex to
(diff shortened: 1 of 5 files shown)


wal_downloader.go
```diff
@@ -15,12 +15,12 @@ import (
 // WALDownloader represents a parallel downloader of WAL files from a replica client.
 //
 // It works on a per-index level so WAL files are always downloaded in their
-// entirety and are not segmented. WAL files are downloaded from minIndex to
+// entiretry and are not segmented. WAL files are downloaded from minIndex to
```
mtlynch: The original (entirety) is correct.

ryanrussell: @mtlynch Sorry about that. Must have been going too fast.

Pushed the fix to put it back to the original(correct) `entirety`...

Best,
Ryan

benbjohnson:
@ryanrussell Sorry for the slow response. Thanks for the doc fixes. I rebased with `main` and merge it in. 👍

PR #394, opened 2022-06-30
feat(docker): harden docker image

hazcod:
This change removes the `alpine` docker image with an empty scratch image.
This removes a bunch of attack surface which is unused anyway.
Additionally it adds-in an unprivileged user such that root is no longer used to reduce attack surface.


benbjohnson:
@hazcod Thanks for the PR. I'll have to do some testing with this. Isn't running as `root` inside Docker typical as the container is already sandboxed?

hazcod:
@benbjohnson It's still a best practice not to, running as root in a container greatly increases the risk of kernel exploitation or container escape. https://dockerlabs.collabnix.com/security/Running-Containers-as-ROOT.html

benbjohnson:
Sorry for the really slow response. I don't have the bandwidth right now to dig into this further so I'm closing for the time being.

PR #414, opened 2022-08-20
Fix restore -if-replica-exists with multiple replicas

hifi:
On 0.3.9 trying to restore from multiple empty replicas with `-if-replica-exists` correctly fails gracefully if they are empty, on master this behavior has regressed making automatic recovery impossible.

I hope this isn't intentional. Thanks!


benbjohnson:
@hifi I added a test case on https://github.com/benbjohnson/litestream/pull/421 but when running it through CI it looks like a bunch of other `RestoreCommand` tests are failing:


```sh
$ go install ./cmd/litestream && go test -run=TestRestoreCommand ./cmd/litestream
flag provided but not defined: -no-such-flag
--- FAIL: TestRestoreCommand (0.10s)
    --- FAIL: TestRestoreCommand/ErrNoOutputPathWithReplicaURL (0.00s)
        restore_test.go:258 unexpected error: cannot determine latest replica: output path required when using a replica URL
    --- FAIL: TestRestoreCommand/ErrReplicaNameWithReplicaURL (0.00s)
        restore_test.go:266: unexpected error: cannot determine latest replica: cannot specify both the replica URL and the -replica flag
    --- FAIL: TestRestoreCommand/ErrInvalidReplicaURL (0.00s)
        restore_test.go:274: unexpected error: cannot determine latest replica: unknown replica type in config: "xyz"
    --- FAIL: TestRestoreCommand/ErrDatabaseNotFound (0.00s)
        restore_test.go:285: unexpected error: cannot determine latest replica: database not found in config: /no/such/db
    --- FAIL: TestRestoreCommand/ErrNoReplicas (0.00s)
        restore_test.go:297: unexpected error: cannot determine latest replica: database has no replicas: /Users/benbjohnson/src/benbjohnson/litestream/cmd/litestream/testdata/restore/no-replicas/db
    --- FAIL: TestRestoreCommand/ErrReplicaNotFound (0.00s)
        restore_test.go:309: unexpected error: cannot determine latest replica: replica "no_such_replica" not found
    --- FAIL: TestRestoreCommand/ErrGenerationWithNoReplicaName (0.00s)
        restore_test.go:321: unexpected error: cannot determine latest replica: must specify -replica flag when restoring from a specific generation
FAIL
FAIL	github.com/benbjohnson/litestream/cmd/litestream	0.377s
FAIL
```

hifi:
These test failures happen because the error handling changed a little as all errors that happen during `loadReplica` at that point are prefixed. Removing the prefix fixes the tests but it also eats the high level action (determine latest replica) from the output.

I'll incorporate your changes and change the tests to use the prefixed text.

hifi:
Did that but if you don't want the prefix to be included then just changing the return to not wrap the error would make the original tests pass. Otherwise the functionality needs to be split so that we can distinguish between ErrNoGeneration and any other error when loading replica.

m7shj:
[mahmoud]()

hifi:
Not relevant with reverted main.

PR #419, opened 2022-08-29
Use cancellable context for snapshotN

hifi:
If the database is heavily written to and there's a one second sync replica running it's possible to end up in a deadlock during shutdown. I've pinpointed this to `snapshotN` not using cancellable context and things grind to a halt.

If there's a good reason not to pass the context to `snapshotN` the implementation might need to be changed to accommodate. I haven't looked deep into how the snapshot iterator works (with S3 backend) to look into the exact spot it locks up.


hifi:
I think this isn't relevant with reverted main.

PR #455, opened 2023-01-11
Fix: AWS Credential Chain

### LINKED ISSUE #427: AWS S3 region detection prior to authentication causes access denied
In environments where AWS S3 credentials are only valid after `session.NewSession` (e.g. when using [assume role with web identity](https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-role.html#cli-configure-role-oidc) which uses `AWS_WEB_IDENTITY_TOKEN_FILE` and `AWS_ROLE_ARN` as the authentication mechanism, supported by `aws-sdk-go`) litestream can encounter access denied errors [when attempting to fetch the bucket region](https://github.com/benbjohnson/litestream/blob/868d5649885ac35111c93346f40fc0f0dd061128/s3/replica_client.go#L85) since the credentials aren't valid yet.

ekristen:
This fixes the aws credential chain such that AWS IRSA works.

Resolves #427

### DIFF
--- a/s3/replica_client.go
+++ b/s3/replica_client.go
@@ -18,7 +18,6 @@ import (
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/aws/awserr"
 	"github.com/aws/aws-sdk-go/aws/credentials"
-	"github.com/aws/aws-sdk-go/aws/defaults"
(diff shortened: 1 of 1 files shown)


benbjohnson:
Thanks for the fix, @ekristen!

ekristen:
Thanks for the awesome tool! Glad to help.

Issue #200, opened 2021-05-23
Azure MSI Support


simongottschlag:
Hi!

Great work on adding the Azure Blob Storage! 😊👍

In Azure, there is something called a Managed Service Identity (MSI) and services in Azure as well as in Kubernetes (using another tool called aad-pod-identity) can use this to get an access token without sharing a secret. This means that as long as your MSI has access to the storage account, you will be able to use it.

It would be a nice addition to the authentication to ABS as a complement to the primary key.

You can see this for more info about how Thanos are about to enable this: https://github.com/thanos-io/thanos/issues/3957

benbjohnson:
Thanks for the heads up, Simon! I'll take a look. If it's not too hard then I'll add it into the next release (v0.3.5). If it's more trouble than I expect then I'll push it back.

PR #459, opened 2023-02-03
Upgrade Azure Storage SDK to "track 2"

ItalyPaleAle:
This application currently depends on an [older version](https://github.com/Azure/azure-storage-blob-go) of the Azure Storage SDK, internally called "track 1", which is going to be deprecated in the not-too-distant future.

This PR updates litestream to use the [newer Azure Storage SDK](https://pkg.go.dev/github.com/Azure/azure-sdk-for-go/sdk/storage/azblob) ("track 2").

Once this is merged, it should be relatively easy to also add support for authentication with Azure AD, which includes using service principals and even MSI (#200)

**Important:** the new SDK requires Go 1.18 so I've updated the version of Go used in this package to Go 1.19 (Go 1.20 was released yesterday, but it's not widely available yet!)

CI tests are passing locally.


hifi:
Hi,

I know it's been a while but would you still be interested in rebasing this PR? The intention seems good and the old SDK is indeed deprecated by now.

Thanks!

corylanou:
Thank you for your contribution to upgrade the Azure SDK to track 2! This was important work that helped identify the necessary changes.

  We've created a fresh implementation in #709 that incorporates your approach while:
  - Using the latest Azure SDK versions
  - Aligning patterns with our recent AWS SDK v2 upgrade
  - Minimizing the changeset for easier review

  Closing in favor of #709. Your work was valuable in understanding the migration requirements!

PR #462, opened 2023-02-21
Upgrade go build environment and lint fixes

tydavis:
- fix: update go versions in mod and docker
  - Update Go module to v1.19 format
  - Docker builder pinned to Go v1.20.1
  - Alpine image pinned to 3.17.2 (rather than `latest`)
- fix: remove deprecated ioutil and rand calls
  - Change ioutil to io and os calls per Go 1.16 deprecation
  - Move global random to localized random source

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM golang:1.17 as builder
+FROM golang:1.20.1 as builder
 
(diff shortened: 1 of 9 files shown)


benbjohnson:
@tydavis Thanks for all these lint fixes. Sorry I missed the PR!

PR #465, opened 2023-03-27
sftp: support ssh certificates

btoews:
I was playing around with replicating to another fly.io app.

### DIFF
--- a/sftp/replica_client.go
+++ b/sftp/replica_client.go
@@ -79,16 +79,15 @@ func (c *ReplicaClient) Init(ctx context.Context) (_ *sftp.Client, err error) {
 	}
 
 	if c.KeyPath != "" {
-		buf, err := os.ReadFile(c.KeyPath)
+		signer, err := c.getSigner()
 		if err != nil {
-			return nil, fmt.Errorf("cannot read sftp key path: %w", err)
-		}
-
-		signer, err := ssh.ParsePrivateKey(buf)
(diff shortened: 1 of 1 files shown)


sftp/replica_client.go
```diff
@@ -128,6 +127,44 @@ func (c *ReplicaClient) Init(ctx context.Context) (_ *sftp.Client, err error) {
...
+	}
+
+	return signer, nil
+}
+
+func (c *ReplicaClient) getCertSigner(priv ssh.Signer) (ssh.Signer, error) {
+	certBuf, err := os.ReadFile(c.KeyPath + "-cert.pub")
+	if err != nil {
+		return nil, err
+	}
```
benbjohnson, line 132: Should there be a separate certificate path (e.g. `c.CertPath`)? That seems better than assuming the user has the cert relative to the key path.

btoews: Most SSHy things have kind of a convention around the names of public keys and certs relative to the private key. If you `ssh -i ~/foobar`, it'll try both the plan key and `~/foobar-cert.pub`. That said,  there's no harm in being explicit and using a separate config if you'd rather.

benbjohnson: I think I'd prefer it be explicit if the convention isn't required. Also, I'm rolling back replication changes that ended up moving to LiteFS so I'll merge this in after I finish that.

Issue #458, opened 2023-01-28
LTX Support


benbjohnson:
The LiteFS project uses the [LTX file format](https://github.com/superfly/ltx) for storing change sets and there a lot of improvements that can be made to Litestream if it used the same format:

- Compaction support
- WAL offset & salt maintained in LTX header
- Compression support can be moved to LTX
- Encryption support (planned)

hifi:
As the streaming replication support that was canned made changes to the file structure, will LTX support also keep the breaking change or allow restoring from legacy (0.3.x) WAL files and structure at least for one minor version (0.4.x) to make migration seamless and safe?

benbjohnson:
@hifi I'm on the fence about backward compatibility. The new version will take a snapshot right away when it starts up so it'll be seamless and safe from that perspective. In theory, it could handle pulling in historic backups but that'd also be a fair bit of code to maintain that support as well.

PR #468, opened 2023-04-16
Client side encryption support for remote storage

### LINKED ISSUE #88: Encryption
Currently, Litestream sends cleartext bytes to a replica such as S3. This works well enough for many use cases as S3 buckets are generally secured through IAM. However, other use cases require additional security measures to prevent attackers from obtaining raw data if a bucket is compromised.

Litestream could implement either encryption to prevent data theft and/or signatures to prevent data alteration.

hifi:
Extremely simple Age wrapper that supports separate recipients and identities. Encrypts both snapshots and WAL files transparently. Technically possible to not have the private key at all if you never restore or do manual restores only.

You can migrate between encrypted/plain by adjusting the identities and recipients respectively but restoring will not work correctly if the remote files aren't encrypted when an identity is configured and vice versa so some care needs to be taken.

Intended use is to encrypt all data before sending out to a potentially untrusted storage provider.

This may conflict with #458 and fixes #88.


benbjohnson:
@hifi Any objections to upgrading to `age` v1.1.1?

cmd/litestream/main.go
```diff
@@ -323,6 +324,10 @@ type ReplicaConfig struct {
 	User     string `yaml:"user"`
 	Password string `yaml:"password"`
 	KeyPath  string `yaml:"key-path"`
+
+	// Encryption identities and recipients
+	AgeIdentities []string `yaml:"age-identities"`
+	AgeRecipients []string `yaml:"age-recipients"`
```
benbjohnson: It'd be nice to reduce the stutter and make the config like:

```yml
age:
  identities: ...
  recipients: ...
```

It can just be an anonymous struct in the config:

```go
type ReplicaConfig struct {
	...
	
	Age struct {
		Identities []string `yaml:"identities"`
		Recipients []string `yaml:"recipients"`
	} `yaml:"age"`
}
```

cmd/litestream/main.go
```diff
@@ -349,6 +354,22 @@ func NewReplicaFromConfig(c *ReplicaConfig, db *litestream.DB) (_ *litestream.Re
 	if v := c.ValidationInterval; v != nil {
 		r.ValidationInterval = *v
 	}
+	for _, str := range c.AgeIdentities {
+		identities, err := age.ParseIdentities(strings.NewReader(str))
+		if err != nil {
+			return nil, err
+		}
+
+		r.AgeIdentities = append(r.AgeIdentities, identities...)
```
benbjohnson, line 365: Why the `append()` instead of just assigning the `identities`? Same goes for the `recipients` below.

go.mod
```diff
@@ -4,14 +4,15 @@ go 1.16
 
 require (
 	cloud.google.com/go/storage v1.15.0
+	filippo.io/age v1.0.0
```
benbjohnson: Upgrade to latest? `v1.1.1`

replica.go
```diff
@@ -67,6 +68,10 @@ type Replica struct {
 	// If true, replica monitors database for changes automatically.
 	// Set to false if replica is being used synchronously (such as in tests).
 	MonitorEnabled bool
+
+	// Encryption identities and recipients
+	AgeIdentities []age.Identity
+	AgeRecipients []age.Recipient
```
benbjohnson, line 74: I don't mind the stutter as much on the struct here. Mostly just the config.

replica.go
```diff
@@ -221,8 +226,20 @@ func (r *Replica) syncWAL(ctx context.Context) (err error) {
...
+	var ew io.WriteCloser = pw
+
+	// Add encryption if we have recipients.
+	if len(r.AgeRecipients) > 0 {
+		var err error
+		ew, err = age.Encrypt(pw, r.AgeRecipients...)
+		defer ew.Close()
+		if err != nil {
+			return err
+		}
```
benbjohnson: The `defer ew.Close()` needs to be moved after the `if` block. If an error returns then `ew` will be `nil` and panic.

replica.go
```diff
@@ -334,6 +353,15 @@ func (r *Replica) calcPos(ctx context.Context, generation string) (pos Pos, err
 	}
 	defer rd.Close()
 
+	if len(r.AgeIdentities) > 0 {
+		drd, err := age.Decrypt(rd, r.AgeIdentities...)
+		if err != nil {
+			return pos, err
+		}
+
+		rd = ioutil.NopCloser(drd)
```
benbjohnson: `ioutil` is deprecated. This should be `io.NopCloser()`.

replica.go
```diff
@@ -470,7 +498,23 @@ func (r *Replica) Snapshot(ctx context.Context) (info SnapshotInfo, err error) {
 	// Copy the database file to the LZ4 writer in a separate goroutine.
 	var g errgroup.Group
 	g.Go(func() error {
-		zr := lz4.NewWriter(pw)
+		var wc io.WriteCloser = pw
+
+		// Add encryption if we have recipients.
+		if len(r.AgeRecipients) > 0 {
+			// We need to ensure the pipe is closed.
+			defer pw.Close()
```
benbjohnson: This seems like it should go at the start of this anonymous function rather than in this `if` block.

replica.go
```diff
@@ -470,7 +498,23 @@ func (r *Replica) Snapshot(ctx context.Context) (info SnapshotInfo, err error) {
...
+
+			var err error
+			wc, err = age.Encrypt(pw, r.AgeRecipients...)
+			defer wc.Close()
+			if err != nil {
+				pw.CloseWithError(err)
+				return err
+			}
```
benbjohnson: The `defer wc.Close()` should be moved after the `if` block.

benbjohnson (changes requested):

cmd/litestream/main.go
hifi: `age.ParseIdentities` and `age.ParseRecipients` work with NL terminated lists and in the config format I chose to use real YAML lists instead of a multi line string to store them, hence need to loop and append them when parsed one by one.

If preferred I can change both to `string` from `[]string`, then it's just an assignment.

go.mod
hifi: Yeah, when this was originally implemented v1.0.0 was the latest one :+1:

replica.go
hifi: Indeed. It wasn't a problem with the compressor that the pipe wasn't closed but with age the last chunk was cut off so I if guarded it. Shouldn't be a problem it's deferred at the beginning.

cmd/litestream/main.go
benbjohnson: Ah, sorry, I misread it. Yeah, that's fine. 👍

benbjohnson (approved):

benbjohnson:
Thanks, @hifi! Sorry for the delay.

ixxie:
I would love to use this feature; is it documented anywhere? I can't seem to find it...

hifi:
@ixxie It's not yet in a final release so the website hasn't been updated.

TL;DR for configuration using v0.3.10 beta if my memory serves right:
```yaml
dbs:
  - path: /path/to/database.sqlite3
    replicas:
      - type: file
        path: /path/to/backups
        age:
          identities:
            - <identity>
          recipients:
            - <recipient>
```

For how to generate age identities (private keys) and recipients (public keys), refer to the [age documentation](https://github.com/FiloSottile/age). Public keys are used for restore only and private keys are used for encryption only so it's possible to create a setup where a Litestream instance only encrypts with a public key but cannot restore from the destination on itself without giving it the private key.

hifi:
As far as I know, the releases aren't on a schedule so they'll happen at their own pace.

ixxie:
I see, thanks @hifi!

PR #473, opened 2023-04-26
Force truncation checkpoint if WAL becomes runaway

hifi:
A failsafe to force WAL truncation if somehow between normal checkpoints the WAL grows to extreme sizes. Default is around 200 megabytes with 4k page size.

This helps if there's sudden write pressure that doesn't allow passive or restart checkpoints to happen for a while. When Litestream is able to checkpoint again it will issue a TRUNCATE one which will roll in the WAL.

### DIFF
--- a/db.go
+++ b/db.go
@@ -31,6 +31,7 @@ const (
 	DefaultCheckpointInterval = 1 * time.Minute
 	DefaultMinCheckpointPageN = 1000
 	DefaultMaxCheckpointPageN = 10000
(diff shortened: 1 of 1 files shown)


hifi:
@benbjohnson Also interested if there's something more that could/should be done? Preventing the WAL from growing indefinitely has been important in high write pressure scenarios.

benbjohnson:
@hifi If you can rebase this PR then I'll merge it as well.

hifi:
Done :+1:

PR #474, opened 2023-04-28
Copy WAL frames through temp file to shadow

hifi:
This implementation replicates what the new behavior in the legacy branch did using a temporary file instead of memory to buffer WAL frames between commits.

Fixes excess memory use if the WAL includes massive commits due to big inserts or migrations.

A small refactor would remove the requirement of a temporary file that was already part of the legacy branch where we keep the cached valid position of the shadow WAL in the DB struct and seek the real WAL to the last valid position instead of expecting the file to be completely intact which isn't guaranteed. After that we can directly write the pages from real WAL to shadow WAL and not worry about partial pages being treated as part of the WAL.

### DIFF
--- a/db.go
+++ b/db.go
@@ -1022,6 +1022,16 @@ func (db *DB) copyToShadowWAL(filename string) (newSize int64, err error) {
 		return 0, fmt.Errorf("last checksum: %w", err)
 	}
 
+	// Write to a temporary shadow file.
+	tempFilename := filename + ".tmp"
+	defer os.Remove(tempFilename)
+
+	f, err := internal.CreateFile(tempFilename, db.fileInfo)
+	if err != nil {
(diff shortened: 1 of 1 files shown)


hifi:
As for why this matters, I created an extremely bad test case that forced the WAL to grow a lot within a single write transaction.

image

Between 10:05 and 10:18 I ran increasingly harder write transactions against a Litestreamed (downstream patched v0.3.9) database that did not have this backport from the legacy branch. Eventually the container went into OOMKill loop as every time Litestream started it would exhaust all of the RAM it was given.

After the gap I ran the same tests with this patch applied. I think because of gc pressure it needed even that much RAM as it did just to get through of everything but regardless it lived. Between 10:40 and 10:45 I dropped a 6 gigabyte table that would normally cause the same issue as Litestream would end up buffering that 6 gigs into RAM.

Issue #237, opened 2021-10-03
A RESTART checkpoint during a snapshot transfer locks the database for writing until the snapshot has finished


t-voet:
Litestream v0.3.5 and a 25 GB database on ZFS with 64k page_size. A SFTP snapshot takes hours, during which the WAL hits 10k pages (640 MB). An obvious solution could be skipping RESTART checkpoints while a snapshot transfer is in progress.

t-voet:
You're the best. If you send me your public key, I can provide a high-speed (2.5 Gbps) SFTP account for testing.

hifi:
As I believe #477 fixed this I'll close. Thanks.

PR #477, opened 2023-05-07
Prevent checkpoints during snapshots

hifi:
Use a mutex that is held during snapshots and only try locking it during checkpoints.

If the database is in high write pressure and taking a snapshot takes long it is possible for the automatic checkpointing to go from PASSIVE to RESTART.

When a RESTART checkpoint is issued SQLite will block new read transactions and wait for the existing ones to finish. However, during a snapshot Litestream keeps a persistent read transaction open until it finishes which will in turn create a deadlock situation for the checkpointer as the RESTART checkpoint will start blocking writers as long as the snapshot is being written out.

This failure condition doesn't break everything persistently but it will create an unfortunate persistent write lock for the application until the snapshot finishes.

### DIFF
--- a/db.go
+++ b/db.go
@@ -49,6 +49,7 @@ type DB struct {
 	rtx      *sql.Tx       // long running read transaction
 	pageSize int           // page size, in bytes
 	notify   chan struct{} // closes on WAL change
+	chkMu    sync.Mutex    // checkpoint lock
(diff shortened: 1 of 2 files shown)


hifi:
This actually fixes #237 as we hit the same issue.

PR #480, opened 2023-05-11
Sync replica snapshots to previous

### LINKED ISSUE #479: Replica snapshot interval should sync to previous snapshot
Noticed a weird pattern that we hadn't gotten any snapshots for a couple of days. This was due to setting high retention time (multiple days) but 24 hour snapshot interval.

What actually happened from what I understand is that when we made releases within 24 hours of the previous one Litestream got restarted each time and as the snapshot interval _begins_ on startup and doesn't try to find the previous snapshot from the replica first it pushed the next snapshot time so far in the future that we had a single WAL stream that spanned multiple wall clock snapshot intervals.

This doesn't cause data loss but it does make the storage pattern weird and restore times longer when that keeps happening for a long time.

I suggest changing the startup pattern to check the current generation latest snapshot time with `maxSnapshot` and if there's one sync the ticker to that.

hifi:
If Litestream is restarted often with high snapshot interval it's possible for it to skip multiple snapshots in a row causing long restore times and inconsistent storage patterns.

This syncs the first snapshot after restart to where it should have been, gracefully fails and continues to normal interval if the remote is unavailable. If the previous snapshot is older than the interval it will snapshot immediately and then continue.

Fixes #479

### DIFF
--- a/replica.go
+++ b/replica.go
@@ -733,6 +733,31 @@ func (r *Replica) snapshotter(ctx context.Context) {
 		return
 	}
 
+	logger := r.Logger()
+	if pos, err := r.db.Pos(); err != nil {
+		logger.Error("snapshotter cannot determine generation", "error", err)
+	} else if !pos.IsZero() {
+		if snapshot, err := r.maxSnapshot(ctx, pos.Generation); err != nil {
+			logger.Error("snapshotter cannot determine latest snapshot", "error", err)
+		} else if snapshot != nil {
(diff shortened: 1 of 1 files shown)


hifi:
@benbjohnson I'd like to know what you think about this. We've been using this in production since I opened this PR and have found it working well.

benbjohnson (approved):
This looks good. If you can update it to use `slog` then I'll merge it in. :+1:

hifi:
Done :+1:

PR #481, opened 2023-05-19
Plumb a custom logger through the core rather than only in Restore

evanphx:
Currently `replica.Restore()` accepts a Logger in it's options, but other things within `Replica` and `DB` use `log.Printf`. This just tidys it up a bit so that both Replica and DB contain a logger (that defaults to `log.Default()`) that can be changed to redirect their logging output.

### DIFF
--- a/db.go
+++ b/db.go
@@ -94,6 +94,9 @@ type DB struct {
 	// List of replicas for the database.
 	// Must be set before calling Open().
 	Replicas []*Replica
+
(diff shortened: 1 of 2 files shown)


benbjohnson:
lgtm! Thanks for wiring that up, @evanphx. 🙏

evanphx:
Thanks @benbjohnson!

PR #483, opened 2023-05-30
Apply memory patches

athoscouto:


athoscouto:
Sorry about it! Was supposed to open it in the fork itself.

PR #488, opened 2023-06-21
Add replicate flags to pump Litestream manually

### LINKED ISSUE #486: Add -once and -force-snapshot arguments to replicate command
Related to #436 I spoke with someone either in some issue or on Slack sometime ago that there's a need for Litestream to be able to be executed periodically instead of continuously.

Our use case is to ensure the local database has been fully replicated to remote storage before we remove it to prevent data loss when we are intentionally moving the process out from the host it was running on and starting it on another. The current workaround is to use exec but it's not always reliable enough and for scheduled maintenance it's sometimes a lot faster to upload a snapshot and download it with restore instead of using the WAL chunks.

My proposal is that the `replicate` command gains `-once` to make it only replicate anything it has not replicated before on startup if the WAL hasn't been rolled out by the application process and `-force-snapshot` to force a snapshot to be taken on startup regardless if `-once` was used or not. This allows us to use them in combination to do a final backup of a database and ensure it's consistent by running `restore` as well.

The `-once` argument needs force all the maintenance tasks as well or they need to be behind another flag like `--force-retention` 

This change would also allow cron style executing of Litestream given two preconditions are met:
 - application uses `PRAGMA wal_autocheckpoint = 0`
 - application uses `SQLITE_FCNTL_PERSIST_WAL`
 - Litestream can issue a passive checkpoint during at least some of the runs

Of course in cron style execution the write pressure must be extremely low not to block Litestream checkpoints and to not grow the WAL ridiculously but for some workloads this might make sense - or they just want to take full snapshots once in a while and have a binary that can do encrypted uploads and downloads to a bucket and always get the latest one on restore with 

I will implement something like this soon in downstream but it'd be great if I could architect it to be upstream compatible given it would be accepted in some form.

hifi:
For some special setups it is sometimes useful to run Litestream manually instead of letting it replicate in the background.

This commit implements the following flags for replicate:

  * -once for doing synchronous replication and then exit
  * -force-snapshot to force a snapshot during -once
  * -enforce-retention to enforce retention rules during -once

Because running once does not respect the snapshot interval the caller is expected to use -force-snapshot and -enforce-retention regularly to ensure the replication targets stay clean.

For this to work correctly with a live database it needs to be opened with auto checkpointing disabled and SQLITE_FCNTL_PERSIST_WAL.

Other uses include only using -force-snapshot to create regular backups of the database instead of live replication.

Fixes #486

---

This PR is still WIP but it'd be nice to hear suggestions around the implementation and leveraging the exec stuff. Screwing around checkpointing by changing the values feels quite dirty. `MinCheckpointPageN` is checked on `Open()` so in most normal runs the added check for it to be non-zero doesn't make a difference.

### DIFF
--- a/cmd/litestream/main.go
+++ b/cmd/litestream/main.go
@@ -92,19 +92,19 @@ func (m *Main) Run(ctx context.Context, args []string) (err error) {
 
 		// Wait for signal to stop program.
 		select {
-		case err = <-c.execCh:
+		case err = <-c.runCh:
 			slog.Info("subprocess exited, litestream shutting down")
 		case sig := <-signalCh:
 			slog.Info("signal received, litestream shutting down")
 
-			if c.cmd != nil {
+			if c.runSignal != nil {
 				slog.Info("sending signal to exec process")
-				if err := c.cmd.Process.Signal(sig); err != nil {
+				if err := c.runSignal(sig); err != nil {
 					return fmt.Errorf("cannot signal exec process: %w", err)
 				}
 
 				slog.Info("waiting for exec process to close")
-				if err := <-c.execCh; err != nil && !strings.HasPrefix(err.Error(), "signal:") {
+				if err := <-c.runCh; err != nil && !strings.HasPrefix(err.Error(), "signal:") {
 					return fmt.Errorf("cannot wait for exec process: %w", err)
 				}
 			}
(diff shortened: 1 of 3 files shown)


jtoar:
Hey @hifi, I just started using Litestream and was interested in using this feature. I'm still a bit of a novice Go programmer but am eager to help if I can. Would comments still be helpful? Or anything else I could do?

PR #517, opened 2023-11-07
fix(restore): not defined verbose flag

choyri:
I was originally using v0.3.11, but recently I upgraded to v0.3.13 and found that a previously working program is now throwing an exception. My execution command is:

```shell
litestream restore -v -if-db-not-exists -if-replica-exists /data/test.db
```

This command exits with an error in v0.3.13. You can reproduce the issue as follows:

```shell
root@bd9b09db6031:/# litestream version
v0.3.13
root@bd9b09db6031:/# litestream restore -v
flag provided but not defined: -v

...

Arguments:

...

	-v
	    Verbose output.

...

2023/11/07 12:19:26 ERROR failed to run error="flag provided but not defined: -v"
```

This breaking change was introduced in [this](https://github.com/benbjohnson/litestream/commit/b1abd6bd993b2b1ee2e9c3a7271a3f523b9be97f#diff-141fdde90cba111ee365dcbb51f4be7f2871ba982d5c22a13e90a799898d0ef2L34). In my commit, I restored the `-v` flag and added relevant warnings.

### DIFF
--- a/cmd/litestream/main.go
+++ b/cmd/litestream/main.go
@@ -36,6 +36,8 @@ var (
 // errStop is a terminal error for indicating program should quit.
 var errStop = errors.New("stop")
 
+var loggerLevel slog.LevelVar
+
 func main() {
 	m := NewMain()
 	if err := m.Run(context.Background(), os.Args[1:]); err == flag.ErrHelp || err == errStop {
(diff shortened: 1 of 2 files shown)


hifi:
This is actually slightly incorrect as verbose logging is enabled by default now and the flag is just dropped. Debug level logging is now the same as trace logging.

I'm going to update the documentation and command line reference to remove that flag.

hifi:
Fixed with #540 and a previous update to litestream.io reference.

choyri:
For those old users who have used "-v" for a long time, directly removing this flag is a breaking change and will cause application downtime. So sad :(

PR #519, opened 2023-11-12
Always release long-running fd on db close

### LINKED ISSUE #510: Unclean close leaves Litestream in weird state
I've been using Litestream as a library and tried a couple of things to abuse it since the usage pattern is a bit different than when using it as a separate process.

What I'm seeing is that the final sync at the beginning of Close() will fail due to this because I had sqlite3 cli just writing while closing (and rolling in the WAL): `cannot verify wal state: stat foo/bar.db-wal: no such file or directory`. This might be fine to ignore but the caller doesn't know it and calling Close() again will fail because the underlying database connection has already been closed and will always return an error at that point.

This is very much an edge case of an edge case and it shouldn't ever really happen but the way database Close() is structured it's designed to ignore all errors and just continue but it might be better to return early and allow calling it more than once to retry if it's a passing issue or handle the errors more gracefully. Now because it's designed to ignore errors the caller needs to assume the database is somewhat closed and ignore the returned error and at most log it.

hifi:
If Litestream is used as a library and it can be closed and re-opened against the same database this leaking fd can break SQLite database file locking. During normal cli operation this race cannot happen as the whole process will exit after closing databases.

Closing Litestream and opening again right after will leave the unclosed file handle out-of-scope and free to garbage collect. Go tries to be helpful and closes the file handle during GC to prevent it leaking permanently but that will cause a race with POSIX advisory locks.

This order of operations will release the _real_ lock the embedded SQLite of Litestream is holding while leaving it clueless. That will allow the main application to take a new lock on the file and eventually both may be checkpointing the same WAL into the main database file causing permanent corruption.

https://www.sqlite.org/howtocorrupt.html#_posix_advisory_locks_canceled_by_a_separate_thread_doing_close_

Fixes #510

### DIFF
--- a/db.go
+++ b/db.go
@@ -355,6 +355,12 @@ func (db *DB) Close() (err error) {
 		}
 	}
 
+	if db.f != nil {
+		if e := db.f.Close(); e != nil && err == nil {
+			err = e
+		}
+	}
+
 	return err
 }
 


benbjohnson (approved):

gjtorikian:
@benbjohnson could a new patch release of the .deb file be made that includes this?

XChikuX:
@benbjohnson New docker images would be nice

PR #522, opened 2023-11-15
Use safe checkpointing before snapshots

### LINKED ISSUE #521: Manual checkpoint before snapshot loses WAL position
If a database is not written to by the main application, Litestream can lose its own WAL position by checkpointing before taking a snapshot without writing into the sequence table after. This requires an extremely specific sequence of events to happen.

1. Start Litestream against a new database
2. Litestream snapshots the initial generation
3. Restart Litestream
4. Litestream snapshots <- this is when the only page in the WAL is checkpointed away
5. Write to database
6. Litestream complains that the WAL was overwritten by another process, starts a new generation

The issue happens here:
https://github.com/benbjohnson/litestream/blob/85ddf322258b96dbde40772661a39e8a7d5a32d9/replica.go#L484-L487

Normally when checkpointing, Litestream will issue a write to the sequence table to force at least one page to be there after restarting the long read transaction which will ensure that regardless of what checkpoint is done that page will stay in until the next "full" checkpoint is made by Litestream and the sequence table is written again.

In this case the read transaction keeps a lock on the page in the WAL at step 2 but at step 4 the read transaction had been restarted after the page was written during previous incarnation of Litestream which makes it free to be checkpointed from the WAL to the main database. Because snapshot will do that rogue checkpoint it will move it silently out without Litestream realizing it until the next write which checks the last page doesn't match the shadow WAL.

The intention of the code is to call `r.db.Checkpoint(ctx, CheckpointModePassive)` rather than do it manually as it will do the required write to sequence table and also make sure the database file is as up-to-date as possible before taking the snapshot.

hifi:
The intention of the code is to do a checkpoint before starting to write out the snapshot. Executing a manual checkpoint will at worst make Litestream lose the track of the WAL as it will in specific circumtances flush out the single page in the WAL file out.

Using the public checkpoint function of the database to issue a managed passive checkpoint will ensure the sequence table is written after and the position of the WAL isn't lost.

Fixes #521

### DIFF
--- a/replica.go
+++ b/replica.go
@@ -482,16 +482,16 @@ func (r *Replica) Snapshot(ctx context.Context) (info SnapshotInfo, err error) {
 	r.muf.Lock()
 	defer r.muf.Unlock()
 
-	// Prevent checkpoints during snapshot.
-	r.db.BeginSnapshot()
-	defer r.db.EndSnapshot()
-
 	// Issue a passive checkpoint to flush any pages to disk before snapshotting.
-	if _, err := r.db.db.ExecContext(ctx, `PRAGMA wal_checkpoint(PASSIVE);`); err != nil {
+	if err := r.db.Checkpoint(ctx, CheckpointModePassive); err != nil {
 		return info, fmt.Errorf("pre-snapshot checkpoint: %w", err)
 	}
 
-	// Acquire a read lock on the database during snapshot to prevent checkpoints.
+	// Prevent internal checkpoints during snapshot.
+	r.db.BeginSnapshot()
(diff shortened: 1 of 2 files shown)


hifi:
This doesn't pass tests because they expect the WAL not to be checkpointed during snapshots. I'm looking into it if I can change the tests to match the new behavior but they are very specific about it and the first WAL index (0) is basically never used when we always checkpoint on snapshot.

hifi:
Figured out the test fixes. Should be good if "losing" zero index is acceptable.

benbjohnson (approved):

PR #523, opened 2023-11-15
Allow setting replica interval offsets

hifi:
If using S3 replicas the extra LIST call on startup may be expensive in scale if a lot of Litestreams are started in quick succession.

Allowing configuring the snapshot offset externally allows the caller to spread around the snapshots without relying on access to the remote replica making restarts no-op.

Same thing applies to retention checks that they can easily stack too close.

### DIFF
--- a/replica.go
+++ b/replica.go
@@ -55,6 +55,9 @@ type Replica struct {
 	// Frequency to create new snapshots.
 	SnapshotInterval time.Duration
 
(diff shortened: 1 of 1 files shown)


hifi:
Expanded to cover all intervals. We use both snapshot and retention check offsets to prevent storming a bucket with requests around the same time.

PR #475, opened 2023-04-30
Use structured logging with slog

hifi:
Configuration file can set the logging handler between two: text, which is the default, and json. Default level is info which matches previous behavior.

A bit more info level logging is enabled by default to default restore to being verbose compared to quiet and snapshotting.

Trace logging has been converted to be on DEBUG level and always enabled if the configured level is high enough.

Bumps Go requirement to 1.20

Some example output with the default text handler and info level:
```
$ gow run ./cmd/litestream/ replicate -config config-file.yaml
time=2023-04-30T20:06:35.254+03:00 level=INFO msg=litestream version="(development build)"
time=2023-04-30T20:06:35.254+03:00 level=INFO msg="initialized db" path=/home/hifi/git/beeper/litestream/restore.db
time=2023-04-30T20:06:35.254+03:00 level=INFO msg="replicating to" name=file type=file sync-interval=1s path=/tmp/litestream-local
time=2023-04-30T20:06:36.258+03:00 level=INFO msg="init: cannot determine last wal position, clearing generation" db=/home/hifi/git/beeper/litestream/restore.db error="primary wal header: EOF"
time=2023-04-30T20:06:36.272+03:00 level=INFO msg="sync: new generation" db=/home/hifi/git/beeper/litestream/restore.db generation=9b0ed7758f1927b6 reason="no generation exists"
time=2023-04-30T20:06:36.274+03:00 level=INFO msg="write snapshot" db=/home/hifi/git/beeper/litestream/restore.db replica=file position=9b0ed7758f1927b6/00000000:4152
time=2023-04-30T20:06:36.890+03:00 level=INFO msg="snapshot written" db=/home/hifi/git/beeper/litestream/restore.db replica=file position=9b0ed7758f1927b6/00000000:4152 elapsed=615.685125ms sz=51092980
time=2023-04-30T20:06:36.890+03:00 level=INFO msg="write wal segment" db=/home/hifi/git/beeper/litestream/restore.db replica=file position=9b0ed7758f1927b6/00000000:0
time=2023-04-30T20:06:36.891+03:00 level=INFO msg="wal segment written" db=/home/hifi/git/beeper/litestream/restore.db replica=file position=9b0ed7758f1927b6/00000000:0 elapsed=1.606214ms sz=4152
time=2023-04-30T20:06:37.264+03:00 level=INFO msg="write wal segment" db=/home/hifi/git/beeper/litestream/restore.db replica=file position=9b0ed7758f1927b6/00000000:4152
[...]

PR #525, opened 2023-11-21
Fix missing imports and out-of-date call for Windows-specific code.

myrkvi:
This Pull Request adds missing imports in `cmd/litestream/main_windows.go` to the `log` and `log/slog` packages, ref. #475, and removes the context from a method call to `ReplicateCommand.Run`, as the Context argument was removed in 2045363cd12a2f8d099eebc63fde046347aa135b.

### DIFF
--- a/cmd/litestream/main_windows.go
+++ b/cmd/litestream/main_windows.go
@@ -5,6 +5,8 @@ package main
 import (
 	"context"
 	"io"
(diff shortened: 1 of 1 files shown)


hifi:
Sorry for the late response. I merged #543 which fixes the build and uses slog only.

PR #530, opened 2023-12-09
feat: add devcontainer configuration

damien-abos:
Add DevContainer to allow development from a Docker container.
See https://code.visualstudio.com/docs/devcontainers/containers


hifi:
Unfortunately I don't think this adds much value because Litestream is trivial to build and run and this file needs to be maintained by someone who actually uses it - even if it works currently.

PR #531, opened 2023-12-09
feat: Add support for storj.io

damien-abos:
Allow to use storj.io with End-to-End encryption and a distributed file storage
See https://www.storj.io/how-it-works


db.go
```diff
@@ -1464,7 +1464,7 @@ func (db *DB) CalcRestoreTarget(ctx context.Context, opt RestoreOptions) (*Repli
 		}
 
 		// Use the latest replica if we have multiple candidates.
-		if !updatedAt.After(target.updatedAt) {
+		if generation == target.generation && !updatedAt.After(target.updatedAt) {
```
hifi, line 1467: Why was this change needed?

hifi (commented):
The author has stated before that if a provider has S3 compatibility no new integrations will be added to Litestream and storj.io seems to.

db.go
damien-abos: I had an issue when restoring a database with a single generation and a target.updateAt set at 0, then the condition was true and at the end there was no replica available.

hifi:
This PR might suffer from #535 which should be checked before merging as the SDK doesn't seem to say anything about object order.

hifi:
Okay #538 fixes Storj through their S3 gateway and probably this PR as well.

I suppose there's still an argument to use their native SDK for client side encryption (even though Litestream currently supports age encryption as well) and direct access to the network rather than through a gateway.

benbjohnson:
@damien-abos I'm going close out this PR as it adds in additional client dependencies which will inflate the binary size. It sounds like there's an S3 gateway as well which would work with Litestream.

db.go
hifi: Even though the PR itself was rejected I'm interested in understanding this issue if it's reproducible with the S3 gateway of Sorj?

hifi: Looking into the code I believe the problem was that your replica implementation returned a generation without updatedAt set to the latest snapshot/WAL file which would've been wrong as otherwise it would always be >0.

PR #553, opened 2023-12-28
Allow configuring all replica defaults

### LINKED ISSUE #504: Feature request: Allow global variables to avoid duplication
I have a dozen sqlite databases, all with the same properties: retention 600h, snapshot-interval 24h, alternate s3 endpoint, etc.  As far as I can tell, at the moment the only way to apply the same properties to all 12 databases is to copy identical values to all 12 databases.

It would be nice if I could specify at the top level that the default retention is 600h, default snapshot-interval is 24h, default s3 endpoint is XXX, so that I don't have to c&p the same information 12 times.  (Similar behavior to access-key-id and secret-access-key.)

hifi:
Fixes #504

### DIFF
--- a/cmd/litestream/main.go
+++ b/cmd/litestream/main.go
@@ -156,6 +156,9 @@ The commands are:
 
 // Config represents a configuration file for the litestream daemon.
 type Config struct {
+	// Global replica settings
+	ReplicaSettings
+
 	// Bind address for serving metrics.
 	Addr string `yaml:"addr"`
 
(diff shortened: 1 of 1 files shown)


corylanou:
Thank you for this PR! We're taking over this work as the branch has become significantly outdated (2 years old) and the codebase has evolved substantially.

The core idea of implementing global replica defaults is excellent and we're implementing a fresh version that builds on your approach but works with the current codebase structure.

New tracking issue: #745

We'll reference this original PR in our implementation. Thanks for the contribution!

PR #554, opened 2023-12-28
Use file extensions as extra metadata

hifi:
Compression algorithm and encryption are derived from the file name.

This is a breaking change for people already using encryption.


corylanou:
Thank you @hifi for this contribution and for your interest in improving Litestream! We really appreciate the time and effort you put into this PR.

As you noted, this would be a breaking change for existing users who are already using encryption. Additionally, with the recent move to the LTX format, compression is now built in automatically, and we're planning to integrate encryption directly into LTX in the future as well.

Given these architectural changes and our current policy of accepting bug fixes only (not new features), we'll need to close this PR. The functionality you were looking to add will be addressed through the LTX format improvements.

Thanks again for your contribution to the project!

PR #557, opened 2024-01-05
 Never directly dereference AWS SDK returned pointers

### LINKED ISSUE #556: Crash with invalid memory address or nil pointer dereference when restoring
Hi @benbjohnson,

first of all: Thank you for this project. This helps a lot for my databases!

I'm using it in different projects and databases and it works quite good.

But now I have an issue when restoring a database after it was working for months. The database file has 10 tables and the largest one has 11072 rows with 35 columns. The total size of the database file is ~ 1 MB. The restore crashes with the following error message:
```
./litestream restore -config litestream.yml /tmp/litestream/experiment.db

/tmp/litestream/experiment.db(s3): restoring snapshot 4b0d294741b7c9ad/00000000 to /tmp/litestream/experiment.db.tmp
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0xed93f9]

goroutine 1 [running]:
github.com/benbjohnson/litestream/s3.(*ReplicaClient).SnapshotReader(0xc00005df00, {0x178bd50, 0xc000120000}, {0xc000515bac, 0x10}, 0x0?)
        /home/runner/work/litestream/litestream/s3/replica_client.go:295 +0x3b9
github.com/benbjohnson/litestream.(*Replica).restoreSnapshot(0xc00037e0f0, {0x178bd50, 0xc000120000}, {0xc000515bac, 0x10}, 0x7fffffff?, {0xc000759200, 0x21})
        /home/runner/work/litestream/litestream/replica.go:1308 +0x18d
github.com/benbjohnson/litestream.(*Replica).Restore(0xc00037e0f0, {0x178bd50, 0xc000120000}, {{0x7ffd1a68bf29, 0x1d}, {0x0, 0x0}, {0xc000515bac, 0x10}, 0x7fffffff, ...})
        /home/runner/work/litestream/litestream/replica.go:1074 +0xae5
main.(*RestoreCommand).Run(0x20a8d18, {0x178bd50, 0xc000120000}, {0xc00012e020, 0x3, 0x3})
        /home/runner/work/litestream/litestream/cmd/litestream/restore.go:90 +0x971
main.(*Main).Run(0xc0000061a0?, {0x178bd50, 0xc000120000}, {0xc00012e010, 0x4, 0x4})
        /home/runner/work/litestream/litestream/cmd/litestream/main.go:123 +0x165
main.main()
        /home/runner/work/litestream/litestream/cmd/litestream/main.go:43 +0x7c
``` 

If I comment the lines 294 and 295 and compile the code:
https://github.com/benbjohnson/litestream/blob/e71e6856d0a892d4bda4e7c6b0ad1463f4a65a25/s3/replica_client.go#L285-L298
the restore is working again. If I instead remove 52 rows of my table or remove 1 column it is working too.

When I download the snapshot file manually and decompress it, it is working too.

I don't really know go language and I'm not sure what these two lines are needed for. When I understand it correctly they are used for some metrics, but I'm not sure if it is safe to remove these two lines, although it would be better to fix the root cause.

Does anybody know why this happens?

Thank you!

hifi:
Apparently AWS or some other provider doesn't set the Content-Length header for all get object requests and we assume it can never be nil. Since it's only used for metrics we can allow them to be incorrect in those cases.

Proper fix would be to wrap the returned `io.ReadCloser` with one that will count the bytes read but that feels a bit much for some edge cases.

Fixes #556

Also replaced #555 with this by wrapping all possibly unsafe uses of AWS SDK returned pointers.

### DIFF
--- a/s3/replica_client.go
+++ b/s3/replica_client.go
@@ -161,7 +161,7 @@ func (c *ReplicaClient) Generations(ctx context.Context) ([]string, error) {
 		internal.OperationTotalCounterVec.WithLabelValues(ReplicaClientType, "LIST").Inc()
 
 		for _, prefix := range page.CommonPrefixes {
-			name := path.Base(*prefix.Prefix)
+			name := path.Base(aws.StringValue(prefix.Prefix))
 			if !litestream.IsGenerationName(name) {
 				continue
 			}
@@ -292,7 +292,7 @@ func (c *ReplicaClient) SnapshotReader(ctx context.Context, generation string, i
 		return nil, err
 	}
 	internal.OperationTotalCounterVec.WithLabelValues(ReplicaClientType, "GET").Inc()
-	internal.OperationBytesCounterVec.WithLabelValues(ReplicaClientType, "GET").Add(float64(*out.ContentLength))
+	internal.OperationBytesCounterVec.WithLabelValues(ReplicaClientType, "GET").Add(float64(aws.Int64Value(out.ContentLength)))
 
 	return out.Body, nil
 }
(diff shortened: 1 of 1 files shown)


benbjohnson (approved):

MalteMagnussen:
@hifi / @benbjohnson - Will this be included in a new Release? :)  

Thanks for all the work on Litestream

PR #568, opened 2024-02-21
Allow to set intervals with cli options

sylr:


corylanou:
Thank you for your contribution! While we appreciate the effort to improve Litestream's configuration options, we generally prefer to avoid adding CLI flags for configuration settings as they become a maintenance burden over time.

Litestream already provides a robust solution for configuration through environment variable expansion in config files. By default, any references to `$VAR` or `${VAR}` in the configuration file are automatically replaced with their environment variable values. This gives users the flexibility to:

1. Define intervals in the config file directly
2. Use environment variables for dynamic configuration (e.g., `sync_interval: ${LITESTREAM_SYNC_INTERVAL}`)
3. Disable expansion with the `-no-expand-env` flag if needed

This approach provides the same flexibility as CLI flags while keeping the command-line interface simpler and the codebase more maintainable. Users can achieve the same dynamic configuration by setting environment variables before running Litestream:

```bash
export LITESTREAM_SYNC_INTERVAL="10s"
export LITESTREAM_SNAPSHOT_INTERVAL="1h"
litestream replicate -config /path/to/config.yml
```

And in the config file:
```yaml
dbs:
  - path: /path/to/db
    replicas:
      - url: s3://bucket/db
        sync-interval: ${LITESTREAM_SYNC_INTERVAL}
        snapshot-interval: ${LITESTREAM_SNAPSHOT_INTERVAL}
```

Given that we already have this flexible configuration system in place, we'll be closing this PR. However, we truly appreciate your interest in contributing to Litestream! If you're interested in other ways to contribute, please check our open issues or feel free to discuss ideas in our GitHub Discussions.

PR #573, opened 2024-03-01
Only update litestream_seq if size is below WAL header size

### LINKED ISSUE #422: Replica gets flooded with WAL segments without writes
first of all: Thank you for making Litestream. It really helped to get our product off the ground with a very simple setup :slightly_smiling_face:

Unfortunately I'm starting to see a super weird behavior in Litestream. Disclaimer: Our application connects to ~18 DBs all replicated.
It started when the db client reported "DB Timeout" errors, when users tried to login (or perform any writing op). In a panic, I restarted the app container. The new container tried to restore the databases, but got stuck.
Checking S3 showed me that some WAL directories had thousands of wal files with only 91B, which were created by Litestream every second. I stopped the container, cleaned all "empty" wal files, started the container again and everything went back to normal.

What I also saw, were DBs that did not receive writes while in this condition, did not had those empty WAL files. Only after I tried to write something to those dbs, litestream started to generate those empty WAL files. This also rooted out my speculation about some loop that would constantly write to the DB. 

The container was running for 28 days, so I thought this might be one reason for this behavior, but now it happened again on a staging container, that was created 2 days ago. Right now I have no idea how to reproduce the issue.

My current litestream version is **0.3.8** and I will upgrade to 0.3.9 now.

Is this a known bug? Can I provide you with more information, that would help you investigating this?

Here is my litestream config:

```yaml
access-key-id: XXXXXXXXXXXXXXX
secret-access-key: XXXXXXXXXXXXXXXX

addr: :9090
dbs:

- path: /XXXXXXX.db
  snapshot-interval: 2h
  rentention: 2d
  replicas:
    - url: s3://XXXXX

# there are 18 more dbs with the same snapshot options in this config
```

benbjohnson:
Previously, the `litestream_seq` could increment on a `PASSIVE` checkpoint even if the WAL existed and did not need to be created. This caused a loop where the increment would trigger replication. This PR fixes this by only executing the increment if the WAL is below a size threshold.

Fixes #422


db.go
```diff
@@ -629,7 +629,7 @@ func (db *DB) acquireReadLock() error {
 	}
 
 	// Execute read query to obtain read lock.
-	if _, err := tx.ExecContext(db.ctx, `SELECT COUNT(1) FROM _litestream_seq;`); err != nil {
+	if _, err := tx.ExecContext(context.Background(), `SELECT COUNT(1) FROM _litestream_seq;`); err != nil {
```
hifi, line 632: Why was this changed to background context? If the db context gets canceled we'd want the lock to fail fast as well, right?

hifi (commented):
Other than one question I don't see there being any problems with this, other than needing testing with real data. I could throw this in our dev environment to see it doesn't do anything bad.

hifi (changes requested):
With enough data, we've discovered that this commit causes a slow stream of `wal overwritten by another process` errors which forces new generations.

Here's a Loki grep of that string over a period of 6 hours:

image

The quiet gap is where I reverted this commit from our downstream fork and then the errors came back when I put the commit back. The final quiet 10 minutes is when I removed the commit again.

I don't yet know why this happens but do you have ideas @benbjohnson?

PR #577, opened 2024-03-24
Upgrade Azure Storage SDK, allow use of default credential

hauntsaninja:
This is mostly just a rebased version of https://github.com/benbjohnson/litestream/pull/459

I also add support for the default Azure credential flow

Co-authored-by: ItalyPaleAle


thecoldwine:
@benbjohnson that is a good patch.

Essentially, it allows Litestream to function with the workload identity, so operations won't need to rotate the key and could utilise the OIDC from Microsoft Entra in K8s or VM deployments.

corylanou:
Thank you for rebasing #459! We appreciate the effort to keep this upgrade moving forward.

  We've created a new implementation in #709 with:
  - Latest Azure SDK packages (v1.6.2)
  - Default credential chain support
  - Pattern consistency with our AWS SDK v2 implementation

  Closing in favor of #709. Thanks for helping push this forward!

PR #580, opened 2024-04-16
Switch to S3 HEAD to find bucket location

### LINKED ISSUE #574: Switch to HEAD instead of get-bucket-location in S3 init routine
AWS documents that HEAD should be used in favour of the get-bucket-location API call that litestream currently uses to determine which S3 endpoint / region to connect to: https://docs.aws.amazon.com/cli/latest/reference/s3api/get-bucket-location.html 

> We recommend that you use [HeadBucket](https://docs.aws.amazon.com/AmazonS3/latest/API/API_HeadBucket.html) to return the Region that a bucket resides in. For backward compatibility, Amazon S3 continues to support GetBucketLocation.

Unfortunately the backwards compatibility support appears to be a bit spotty - in partiuclar, there are *many* reports to the get-bucket-location API call failing to reliably work cross-region - the very situation where it is being used... e.g. 
https://github.com/aws/aws-sdk-go/issues/720
https://github.com/cloud-custodian/cloud-custodian/issues/7695

I have a role account running under ECS that is hitting this. I know the role account permissions are fine, because litestream works if I explicitly configure the region in the config file; but when trying to use litestream without a config file on the command-line it all falls apart:
```
/ # litestream restore -o test.db s3://BUCKET/PATH
2024/03/21 06:42:23 ERROR failed to run error="cannot fetch generations: cannot lookup bucket region: AccessDenied: Access Denied\n\tstatus code: 403, request id: Z9QYDYMRKTC8SRPB, host id: ..."
```
Which seems to match with the problematic get-bucket-location call being bypassed when a region is specified in the config per https://github.com/benbjohnson/litestream/blob/94f69a0eb32a9a23571f0cd88dcfc9c32ddd426f/s3/replica_client.go#L79-L88, but unfortunately I cannot see/find any way to set a similar config option from the command-line only... 

Attempting to replicate the behaviour with the raw aws CLI shows the issue:
```
/ # aws s3api get-bucket-location --bucket BUCKET --output text | cat
ap-southeast-2
/ # AWS_REGION=us-east-1 aws s3api get-bucket-location --bucket BUCKET --output text | cat

An error occurred (AccessDenied) when calling the GetBucketLocation operation: Access Denied
```

It's not that litestream is doing anything "wrong" here (other than using an old/not recommended API)... just the API is unreliable and stupid. 

The recommended head-bucket call does seem to reliably work however:
```
/ # aws s3api head-bucket --bucket BUCKET --debug 2>&1 | grep x-amz-bucket-region
2024-03-21 07:21:13,247 - MainThread - botocore.parsers - DEBUG - Response headers: {'x-amz-id-2': 'viRHfLso2hnlc/saJECbtrIgrQfOWI4nDxff7QCsYUWEOucXO42Xkm8DdtsBzCz2XwFyLAut98KUvUPpRDmKZA==', 'x-amz-request-id': 'DJ92Q5D6A8717KPB', 'Date': 'Thu, 21 Mar 2024 07:21:14 GMT', 'x-amz-bucket-region': 'ap-southeast-2', 'x-amz-access-point-alias': 'false', 'Content-Type': 'application/xml', 'Server': 'AmazonS3'}

/ # aws s3api head-bucket --bucket BUCKET --endpoint https://s3.us-east-1.amazonaws.com/ --debug 2>&1| grep x-amz-bucket-region
2024-03-21 07:23:20,040 - MainThread - botocore.parsers - DEBUG - Response headers: {'x-amz-bucket-region': 'ap-southeast-2', 'x-amz-request-id': 'ZKZBSKCGY78EXRBC', 'x-amz-id-2': 'mPQgYi9PPQj/fW3S+gJ0qBdttaj2db9DBjdjk3/RhIJL95T9/yA2w5KUGvFNOQldZWoXtMeQNFM=', 'Content-Type': 'application/xml', 'Date': 'Thu, 21 Mar 2024 07:23:19 GMT', 'Server': 'AmazonS3'}
```

Any objections to changing to head instead? 

I'm happy to send a PR if this change would be welcome.

ananthb:
S3's GetBucketLocation is deprecated so switching to a HEAD request to find the bucket location.

Fixes #574.

### DIFF
--- a/s3/replica_client.go
+++ b/s3/replica_client.go
@@ -136,12 +136,12 @@ func (c *ReplicaClient) findBucketRegion(ctx context.Context, bucket string) (st
 
 	// Fetch bucket location, if possible. Must be bucket owner.
 	// This call can return a nil location which means it's in us-east-1.
-	if out, err := s3.New(sess).GetBucketLocation(&s3.GetBucketLocationInput{
+	if out, err := s3.New(sess).HeadBucketWithContext(ctx, &s3.HeadBucketInput{
 		Bucket: aws.String(bucket),
 	}); err != nil {
 		return "", err
-	} else if out.LocationConstraint != nil {
-		return *out.LocationConstraint, nil
+	} else if out.BucketRegion != nil {
+		return *out.BucketRegion, nil
 	}
 	return DefaultRegion, nil
 }


ananthb:
Not sure if the default bucket stuff is still needed. AWS S3 only says that directory buckets do not have a location associated with them.

benbjohnson:
Thanks!

myisaak:
This PR breaks the replicate functionality and we get again the error from #574. Reverting to previous commit removed this error.

The error:
```bash
/ # litestream restore -o test.db s3://BUCKET/PATH
2024/03/21 06:42:23 ERROR failed to run error="cannot fetch generations: cannot lookup bucket region: AccessDenied: Access Denied\n\tstatus code: 403, request id: Z9QYDYMRKTC8SRPB, host id: ..."
```

sylr:
@ananthb @benbjohnson I can confirm that this breaks the replicate command when the bucket is not in the default us-east-1 region.

ananthb:
Have you tried opening a new issue instead?

PR #596, opened 2024-07-31
Add NATS replica client

delaneyj:
[NATS](https://docs.nats.io/) is a distributed application framework.  It has the ability to store objects similar to S3/Minio in a replicated persistence store.  This replica client supports this feature for Litestream.

***Note***
The `replica_client_test.go` was geared towards only being tested against `file`.  Have moved from `fs/os` errors to Litestream specific ones.  This meant updating the file replica client to use these as well.


file/replica_client.go
```diff
@@ -146,7 +146,7 @@ func (c *ReplicaClient) DeleteGeneration(ctx context.Context, generation string)
 func (c *ReplicaClient) Snapshots(ctx context.Context, generation string) (litestream.SnapshotIterator, error) {
 	dir, err := c.SnapshotsDir(generation)
 	if err != nil {
-		return nil, fmt.Errorf("cannot determine snapshots path: %w", err)
+		return nil, litestream.ErrSnapshotPathNoGeneration
```
benbjohnson: Why is this removing the error wrapping? The returned `err` from `SnapshotsDir()` could be anything and not just `"no generation"`.

nats/replica_client.go
```diff
@@ -0,0 +1,402 @@
...
+import (
+	"context"
+	"fmt"
+	"io"
+	"strings"
+	"time"
+
+	"github.com/benbjohnson/litestream"
+	"github.com/benbjohnson/litestream/internal"
+	natsClient "github.com/nats-io/nats.go"
```
benbjohnson: Why not use `nats` as the alias since that's what's used in the client package? https://github.com/nats-io/nats.go/blob/v1.37.0/context.go#L14

nats/replica_client.go
```diff
@@ -0,0 +1,402 @@
+package nats
```
benbjohnson, line 1: This doesn't seem to be imported or usable by the `cmd/litestream` binary. Am I missing something?

nats/replica_client.go
```diff
@@ -0,0 +1,402 @@
...
+
+const (
+	ReplicaClientType     = "nats"
+	DefaultBucketName     = "litestream"
+	DefaultBucketReplicas = 3
+	DefaultBucketMaxBytes = 0               // unlimited
+	DefaultBucketTTL      = 0 * time.Second // unlimited
+
+	generationsPrefix    = "generations/"
+	generationsPrefixLen = len(generationsPrefix)
```
benbjohnson: The `generationsPrefixLen` seems extraneous since the [StringHeader](https://pkg.go.dev/reflect#StringHeader) includes the length. I can't imagine there's a performance boost.

nats/replica_client.go
```diff
@@ -0,0 +1,402 @@
...
+	return &ReplicaClient{
+		nc:             nc,
+		BucketName:     DefaultBucketName,
+		BucketReplicas: DefaultBucketReplicas,
+		BucketMaxBytes: DefaultBucketMaxBytes,
+		BucketTTL:      DefaultBucketTTL,
+	}
+}
+
+// Returns the type of client.
```
benbjohnson: Go method and type comments should start with the name of the method or type:

```
// Type returns the type of client.
```

instead of

```
// Returns the type of client. 
```

benbjohnson (changes requested):
I added a few points of feedback. Can you wire this up to `cmd/litestream` and add some instructions on how to test it?

file/replica_client.go
delaneyj: I got over zealous with the moving to a general set of errors, fixed.

nats/replica_client.go
delaneyj: because I was in the `nats` package I thought that would cause issues, apparently not the case, fixed

nats/replica_client.go
delaneyj: was using as a constant it a few places,  moved to place of usage.

nats/replica_client.go
delaneyj: I had hooked it up (i thought) in `cmd/litestream/replicate.go` around line 125.  Apparently this code wasn't commited

betamos:
Big NATS fan here (oh, and Litestream too of course). Just stumbled across this PR accidentally and looked through it briefly. Some thoughts:

- Why auto-create the bucket? IIUC litestream wants you to provision resources yourself (like the S3 client), which is trivial with the NATS CLI. There are lots of config options that could pollute the litestream config file format (currently I see replica count).
- Related: Retention is set on the NATS bucket. Without official support for out-of-bounds retention, I think it's much safer to let litestream manage retention in the same way it does elsewhere, using periodic checks.
- Only JWT authentication? I think user/pass should absolutely be supported as a base-case, it's very common. (It can technically be encoded in the URL but since user/pass fields exist in litestream.yml, it violates the principle of least surprise).

Apologies for stepping in unannounced. Let me know if I can help out any other way.

stumct-sa:
I was looking through the litestream docs to see if it supported NATS then came across this PR. Would be great to see this get over the line.

I agree with the previous comment about the bucket configuration. I would expect to configure my own bucket and just tell litestream which one to use. There is a number of bucket configurations that will be set depending on my use case at the time. I suggest just taking in a bucket name and expecting it to exist.

JWT authentication is my preferred approach but again I wouldn't have litestream be so opinionated. It should support any NATS auth mechanism

delaneyj:
I have been slammed with other projects but I agree with the comments above and will remove the bucket config from this sorry and look for a review

nats/replica_client.go
```diff
@@ -0,0 +1,403 @@
...
+		return nil, err
+	}
+
+	c.objectStore = objecStore
+
+	return objecStore, nil
+}
+
+func (c *ReplicaClient) Cleanup(ctx context.Context) (err error) {
+	if c.nc == nil {
```
bruth, line 373:
```suggestion
	if c.nc != nil {
```

nats/replica_client.go
```diff
@@ -0,0 +1,403 @@
...
+		c.nc.Close()
```
bruth, line 374:
```suggestion
		c.nc.Drain()
		// Block until draining is done. There is also a `ClosedHandler`
		// callback option for the NATS connection where a channel
		// could be used.. 
		for c.nc.IsDraining() {}
```

nats/replica_client.go
```diff
@@ -0,0 +1,403 @@
...
+
+type ReplicaClient struct {
+	nc          *nats.Conn
+	objectStore jetstream.ObjectStore
+
+	// NATS connection info
+	BucketName     string
+	BucketReplicas int
+	BucketMaxBytes int64
+	BucketTTL      time.Duration
```
bruth, line 35: I presume the goal is an out-of-box "no setup required", but for all other backends (based on a quick scan of the docs) require setting up the bucket ahead of time. These particular properties are operational concerns which should, IMO, be handled out-of-band and would be safer. For instance, the replicas would need to change for some reason. When Litestream would startup, an existing configuration could conflict with the remote settings of that bucket and either overwrite it (depending on the setting) or error since it has changed.

I suggest removing these and the requirement would be that a bucket is pre-created and you only need credentials and the bucket name when configuring this backend.

This would also remove the need to have `upsertObjectStore`.

nats/replica_client.go
```diff
@@ -0,0 +1,403 @@
...
+	}
+}
+
+// Type returns the type of client.
+func (c *ReplicaClient) Type() string {
+	return ReplicaClientType
+}
+
+// Returns a list of available generations. Order is undefined.
+func (c *ReplicaClient) Generations(ctx context.Context) (_ []string, err error) {
```
bruth, line 54: I did not peak at the other backend implementations, so if this is the idiomatic expectation in the codebase then 👍. However, at least for this backend implementation is a bit more difficult to read which values are being returned at times across some of the methods since there is scoped explicit assignment of `err` in particular vs. implicit.

nats/replica_client.go
```diff
@@ -0,0 +1,403 @@
...
+
+	buf := bytebufferpool.Get()
+	defer bytebufferpool.Put(buf)
+
+	if _, err := io.Copy(buf, r); err != nil {
+		return info, err
+	}
+
+	b := buf.Bytes()
+	natsInfo, err := objectStore.PutBytes(ctx, filename, b)
```
bruth, line 149: Could use `Put` which takes an `io.Reader` directly.

cmd/litestream/main.go
```diff
@@ -363,6 +366,12 @@ type ReplicaConfig struct {
 	Password string `yaml:"password"`
 	KeyPath  string `yaml:"key-path"`
 
+	// NATS settings
+	JWT            string `yaml:"jwt"`
+	Seed           string `yaml:"seed"`
```
bruth, line 371: All auth methods should be supported. If an additional dependency is welcome, this package handles all the options as well as looking up the `nats context` profile. https://pkg.go.dev/github.com/nats-io/jsm.go/natscontext

bruth (commented):
@delaneyj Took some time to review given our conversation this past week.

nats/replica_client.go
bruth: Sorry, I didn't define what "it" is in the first sentence.. named return values.

bruth:
Annnd.. I just realized I was a bit redundant with the other commentary 🤦. Apologies folks..

colinmollenhour:
This is really interesting, can someone describe the use cases or advantages for it compared to a regular S3 bucket? I.e. how is this being utilized to do something different than before?

corylanou:
Great question! Litestream supports [multiple replica destinations](https://litestream.io/reference/config/) (S3, GCS, Azure, SFTP, File, and NATS), each with different advantages depending on your infrastructure and deployment model.

Here are the key use cases and advantages for using NATS as a replication backend:

**Infrastructure Consolidation**: If you're already running NATS for messaging/eventing in your stack, you can use the same infrastructure for database backups without needing separate cloud storage services or credentials. See the [NATS replica guide](https://litestream.io/guides/nats/) for configuration details.

**Edge and Distributed Deployments**: NATS JetStream excels at edge computing scenarios where you might want to keep backups close to where your application runs. The built-in clustering and replication capabilities make it easier to maintain backups across geographically distributed locations without managing multiple S3 buckets or regions.

**Self-Hosted Infrastructure**: For organizations that prefer to maintain complete control over their infrastructure, NATS provides a robust option without dependency on cloud providers. This aligns well with Litestream's philosophy of [simplifying single-server deployments](https://litestream.io/) while still maintaining reliable replication.

**Simplified Operations**: In NATS-centric environments (especially [Kubernetes clusters](https://litestream.io/guides/kubernetes/) with NATS already deployed), you can avoid managing cloud provider accounts, API keys, and IAM policies. Authentication is handled through NATS's existing mechanisms (JWT, NKey, or username/password).

**Cost Control**: For organizations already running NATS infrastructure, there's no additional cost for object storage beyond the infrastructure you already have. This can be more economical than paying per-GB to cloud storage providers, especially for development/staging environments.

**On-Premises and Air-Gapped Environments**: NATS provides an excellent option for environments that can't or don't want to use cloud storage services, while still getting distributed, replicated backup storage.

In summary, if you're already invested in the NATS ecosystem or need self-hosted infrastructure, NATS is a natural fit. If you're AWS-native, the [S3 replica](https://litestream.io/guides/s3/) remains an excellent choice. The beauty of Litestream is having these options to match your specific infrastructure needs.

PR #597, opened 2024-07-31
fix nil == nil warnings

delaneyj:
Invalid checks that will always be true

### DIFF
--- a/db.go
+++ b/db.go
@@ -330,7 +330,7 @@ func (db *DB) Close(ctx context.Context) (err error) {
 
 	// Perform a final db sync, if initialized.
 	if db.db != nil {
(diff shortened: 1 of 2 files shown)


corylanou:
Hi @delaneyj, thank you for this PR!

  Good news - all three issues you identified have already been addressed in the current codebase:

  1. The db.Close() method in db.go no longer has the redundant err == nil check
  2. The applyWAL function has been removed entirely
  3. The Restore() method in replica.go now correctly uses just !os.IsNotExist(err) without the redundant nil check

  Since these fixes are already in place, I'll close this PR. Thanks again for taking the time to identify and fix these issues!

delaneyj:
Great, I'll finish up my nats PR now!

PR #607, opened 2024-09-14
Add godotenv

containerman17:
Purpose: To retrieve S3 credentials from a .env file.


benbjohnson:
Sorry for the slow response. I'm going to close this PR as I think the environment should handle the environment variables rather than a built-in library.

Issue #440, opened 2022-10-27
Programmatic Usage


alapidas:
I couldn't find an existing issue asking this - is there a way to consume this programmatically?  On some app platforms, one can point to a code base and have it automatically built and deployed.  In this case, you can't use litestream, but if you're writing a Go-based app then you could import it and do a restore on app startup, and run periodic backups in a goroutine.

As a workaround, one could build a docker image that had litestream in it, but it can be more attractive to use it in-process in the example above.

Apologies in advance if this is already addressed elsewhere.

alapidas:
And I just discovered https://github.com/benbjohnson/litestream/issues/215

alapidas:
I read through https://github.com/benbjohnson/litestream-library-example/blob/main/main.go and it's sort of complex - is there any interest in creating a friendlier API for this?

lherman-cs:
@alapidas this might be what you're looking for, https://github.com/benbjohnson/litestream/pull/608.

bbkane:
I still want this 😀

corylanou:
I've created a PR with library usage examples: #921

The examples show:
- **Basic (file backend)**: Simplest possible usage with local filesystem replication
- **S3 backend**: Complete example with restore-on-startup pattern

Key API pattern:
```go
db := litestream.NewDB("/path/to/db.sqlite")
client := file.NewReplicaClient("/path/to/replica")
replica := litestream.NewReplicaWithClient(db, client)
db.Replica = replica
db.Open()
defer db.Close(context.Background())
sqlDB := db.SQLDB() // use for normal operations
```

The README includes an API stability warning since the library API is not considered stable (the CLI interface is more stable for production use).

Full documentation will be added to litestream.io separately.

corylanou:
Hi @alapidas - I've created a PR (#921) with library usage examples that show how to use Litestream programmatically in a Go application.

The examples demonstrate:
- Creating a `litestream.DB` wrapper around your database
- Attaching a replica client (file, S3, etc.)
- Running replication in the background
- Restore-on-startup pattern (check for local DB, restore from backup if missing)

Is this what you were looking for? Specifically, does this address your use case of:
- Restoring on app startup
- Running periodic backups in a goroutine

Or were you looking for something different, like a more formal/stable library API with versioning guarantees?

The current approach uses the existing internal types (`DB`, `Replica`, `ReplicaClient`) directly. The README includes a warning that the library API is not considered stable compared to the CLI.

[...]

Issue #215, opened 2021-06-06
Documentation on using litestream as a library in a Go project


Koeng101:
I'm looking at embedding litestream into another Go project I have running, but found that there aren't any docs on how to use litestream within a different Go project. It would be much appreciated if there was a simple example of how to use litestream as a library. Thanks!

benbjohnson:
It is possible to use Litestream as a library but since it’s not v1.0.0 yet, the API doesn’t have any stability guarantees so we don’t publicize it very much. You can find an example repo here though: https://github.com/benbjohnson/litestream-library-example

janflyborg:
I know this ticket has been closed, but any updates on this? Specifically, I'm thinking about having one SQLite database for each user in my system, and it would be cumbersome (and not verify efficient) to start one Litestream process for each database, so is the API stable enough for such usage now?

benbjohnson:
@janflyborg I provided that `litestream-library-example` repo, but after further use I've found that there's an issue with it on macOS. The lock system in macOS doesn't allow for per file descriptor locks so there can be issues with adding and removing SQLite databases under management with Litestream on that OS. Not sure if it applies to you but it's a bit sticky if it does.

For your use case, there is ticket #42 scheduled for the `v0.4.0` release where it will be able to replicate entire directories so you can have multiple databases managed without setup a separate process for each. However, Litestream uses a polling mechanism currently that can be really slow if you have a lot of databases. That's also being fixed via #190.

janflyborg:
Thanks for the answer. That was exactly what I was looking for. 

Let me tell you that I really appreciate your work. Our company has been a long time user of Bolt, but switched to SQLite a few years back. This initiative (Litestream) is really covering new ground, and your effort it is highly appreciated.

benbjohnson:
Thanks, @janflyborg! I appreciate it. That means a lot to hear.

lherman-cs:
@Koeng101 @janflyborg https://github.com/benbjohnson/litestream/pull/608 might fulfill your needs, please feel free to scrutinize the change.

Issue #42, opened 2021-02-11
Replicate whole directories


noteflakes:
This is a great project! Would it be possible to configure replication for whole directories with multiple database files, with arbitrary number of nested directories?

benbjohnson:
Thanks! It's not possible right now but that's definitely something I would like to add a some point.

tuliofaria:
This is a perfect feature for multitenancy splitting into by tenants db files.

lherman-cs:
What do you guys think about this change? https://github.com/benbjohnson/litestream/pull/608

@tuliofaria @benbjohnson @noteflakes

yawaramin:
@lherman-cs see https://github.com/benbjohnson/litestream?tab=readme-ov-file#contribution-policy

lherman-cs:
> @lherman-cs see [benbjohnson/litestream#contribution-policy](https://github.com/benbjohnson/litestream?tab=readme-ov-file#contribution-policy)

@yawaramin I was not aware of this. Thanks for the link. I'll cancel the PR.

PR #608, opened 2024-10-02
refactor main to allow cli as high-level lib API

lherman-cs:
Motivation: use Litestream as a library

**Related Issues:**
  1. https://github.com/benbjohnson/litestream/issues/440
  2. https://github.com/benbjohnson/litestream/issues/215

**Current Workaround: https://github.com/benbjohnson/litestream-library-example**

The current workaround is complex, and no guarantee on the API backward compatibility
(https://github.com/benbjohnson/litestream/issues/215#issuecomment-855426990).

**Solution: refactor out main package into `cmd` package**

1. Backward Compatibility

> CLI interface is more stable than calling Litestream directly via function calls. Most users already use Litestream through CLI (including service wrappers), so backward compatibility at this interface level is desired.

2. Simple

> Using Litestream as a library is now as simple as the following:

`cmd.Run([]string{"litestream", "replicate", "/path/to/db", "s3://BUCKETNAME/PATHNAME"})` `cmd.Run([]string{"litestream", "restore", "-o", "/path/to/db", "s3://BUCKETNAME/PATHNAME"})`


Koeng101:
`main.go` is mostly just copied, and the commit history should be maintained. Move it first, commit, then commit updates.

I wish there was a better way to do this, but this does seem like an alright way to run it. However, I don't see any documentation updates or new examples. I'd like to see a documentation update somewhere, as well as a [Go example](https://go.dev/blog/examples) of how to use it.  Especially around proper closing and error handling (I imagine you'll have to load as a goroutine, but then how about context returns? Or error returns upon failure to replicate? How do you handle?)

lherman-cs:
@Koeng101 I mainly wanted to use this to start a conversation before investing too early in examples. From https://github.com/benbjohnson/litestream/issues/42#issuecomment-2387601816, this is very likely not going to get merged. So, I will cancel the PR.

There are 2 ways to handle closing and error handling:

1. Via signals, https://pkg.go.dev/os/signal#Notify

Litestream listens to SIGINT. `signal.Notify` allows multiple signal handlers. So, the main program can also listen to SIGINT and wait for Litestream to end gracefully.

2. Via context cancellation, similar to 1 but more generic. Although, I just saw that Litestream doesn't seem to handle context cancellation for replicate subcommands. So, it needs to be patched.

So, the workflow requires a "supervisor" goroutine that manages the error and closing for each replicate goroutine. This is far from ideal, but it works with low effort.

Something like this (WARNING: haven't been tested):
```go
ctx, stop := signal.NotifyContext(context.Background(), os.Interrupt)
defer stop()
g, ctx := errgroup.WithContext(ctx)

gCtx := ctx
g.Go(func() error {
  return cmd.Run(gCtx, []string{"litestream", "replicate", "/path/to/db", "s3://BUCKETNAME/PATHNAME"})
})

// proceed with main program
g.Go(...) <-- main goroutine

err := g.Wait()
// handle main program or litestream errors here
```

PR #620, opened 2024-12-09
Add busy-timeout-interval setting

darkgnotic:
Adds a `busy-timeout` yaml setting that adjusts the `PRAGMA busy_timeout` in litestream to handle applications that hold the lock for longer than the default value of 1 second. 

(i.e. the converse of setting the https://litestream.io/tips/#busy-timeout in the application)

### DIFF
--- a/cmd/litestream/main.go
+++ b/cmd/litestream/main.go
@@ -283,12 +283,13 @@ func ReadConfigFile(filename string, expandEnv bool) (_ Config, err error) {
 
 // DBConfig represents the configuration for a single database.
 type DBConfig struct {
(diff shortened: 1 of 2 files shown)


benbjohnson (changes requested):
Can you change the name from "busy timeout interval" to just "busy timeout"? It's not really an interval since it's not repeating. Otherwise the PR looks good.

darkgnotic:
> Can you change the name from "busy timeout interval" to just "busy timeout"? It's not really an interval since it's not repeating. Otherwise the PR looks good.

Ah, good point! Done!

benbjohnson (approved):

PR #655, opened 2025-07-08
Add MCP Server Integration and Tools for Litestream Remote Control

corylanou:
This PR introduces a new MCP (Machine Control Protocol) server integration into Litestream, enabling remote control and monitoring capabilities. Key highlights:

- **MCP Server:**  
  - Adds a new MCP server (with `-mcp` and `-mcp-port` flags) to the `replicate` command, allowing Litestream to be controlled remotely via HTTP.
  - Implements server startup, shutdown, and HTTP handling logic in `cmd/litestream/mcp/mcp.go`.

- **MCP Tools:**  
  - Adds a suite of MCP tools under `cmd/litestream/mcp/tools/` for operations such as listing databases, generations, snapshots, WAL files, and version info.
  - Each tool wraps the corresponding Litestream CLI command and exposes it via the MCP server.

- **Configuration & Flags:**  
  - Updates the configuration struct and default config to support MCP options.
  - Adds new flags to the `replicate` command for enabling and configuring the MCP server.

- **Dependency Updates:**  
  - Bumps Go version to 1.24.x.
  - Adds dependencies for MCP server and HTTP logging.

- **Makefile & Dockerfile:**  
  - Updates build targets to use Go 1.24.
  - Adds new Makefile targets for MCP wrapping and inspection.

This change lays the foundation for advanced remote automation and observability of Litestream instances, making it easier to manage and monitor deployments programmatically.


cmd/litestream/main.go
```diff
@@ -172,6 +174,14 @@ type Config struct {
...
 	// Logging
 	Logging LoggingConfig `yaml:"logging"`
+
+	// MCP server options
+	MCPEnabled bool `yaml:"mcp-enabled"`
+	MCPPort    int  `yaml:"mcp-port"`
+
+	// Path to the config file
+	// This is only used internally to pass the config path to the MCP tool
+	ConfigPath string `yaml:"-"`
```
corylanou, line 181: I'm not super happy about this.  They only reason I need it is that I found if you were on a non-traditional install of litestream (like local mac development) it would start to hallucinate about where the config ~should~ be and then the LLM just wasted all your time.  Since the program definitively knows what the config path is when it loads, it makes sense to pass that knowledge to the MCP tools so that if the user doesn't specify it when asking the LLM questions, it doesn't get confused.

I'm open to options/thoughts here.

cmd/litestream/main.go
```diff
@@ -172,6 +174,14 @@ type Config struct {
 
 	// Logging
 	Logging LoggingConfig `yaml:"logging"`
+
+	// MCP server options
+	MCPEnabled bool `yaml:"mcp-enabled"`
+	MCPPort    int  `yaml:"mcp-port"`
```
benbjohnson: Can you make this similar to the `Addr` field and just have an `MCPAddr` that is enabled if set? I prefer that over "port" so it can fully set the bind address.

cmd/litestream/main.go
```diff
@@ -670,7 +682,7 @@ func newSFTPReplicaClientFromConfig(c *ReplicaConfig, r *litestream.Replica) (_
 	return client, nil
 }
 
-// applyLitestreamEnv copies "LITESTREAM" prefixed environment variables to
+// applyLitestreamEnv copies "LITESTEAM" prefixed environment variables to
```
benbjohnson: Was this intentional?

cmd/litestream/mcp/mcp.go
```diff
@@ -0,0 +1,58 @@
...
+
+import (
+	"context"
+	"fmt"
+	"log/slog"
+	"net/http"
+	"time"
+
+	"github.com/MadAppGang/httplog"
+	"github.com/benbjohnson/litestream/cmd/litestream/mcp/tools"
```
benbjohnson: Do these functions need to live in a separate package? It seems like they could live in `cmd/litetream`.

cmd/litestream/mcp/tools/replicate.go
```diff
@@ -0,0 +1,44 @@
...
+			args = append(args, strings.Split(replicaURLs, ",")...)
+		}
+		cmd := exec.CommandContext(ctx, "litestream", args...)
+		output, err := cmd.CombinedOutput()
+		if err != nil {
+			return mcp.NewToolResultError(strings.TrimSpace(string(output)) + ": " + err.Error()), nil
+		}
+		return mcp.NewToolResultText(string(output)), nil
+	}
+}
```
benbjohnson: I'm not sure we need the replicate tool. If Litestream should already be running generally and I'd be a little hesitant to have it correctly configure everything correctly.

cmd/litestream/replicate.go
```diff
@@ -43,6 +49,8 @@ func (c *ReplicateCommand) ParseFlags(ctx context.Context, args []string) (err e
 	fs := flag.NewFlagSet("litestream-replicate", flag.ContinueOnError)
 	execFlag := fs.String("exec", "", "execute subcommand")
 	configPath, noExpandEnv := registerConfigFlag(fs)
+	mcpFlag := fs.Bool("mcp", false, "Enable MCP server (agent mode)")
+	mcpPortFlag := fs.Int("mcp-port", 8080, "Port for MCP server (default 8080)")
```
benbjohnson: Can you remove these from the CLI? We put all the real configuration stuff in the `Config`.

cmd/litestream/replicate.go
```diff
@@ -175,35 +200,38 @@ func (c *ReplicateCommand) Close() (err error) {
...
 
 // Usage prints the help screen to STDOUT.
 func (c *ReplicateCommand) Usage() {
-	fmt.Printf(`
-The replicate command starts a server to monitor & replicate databases.
-You can specify your database & replicas in a configuration file or you can
-replicate a single database file by specifying its path and its replicas in the
-command line arguments.
+	fmt.Print(`
+The replicate command runs a server to replicate databases.
```
benbjohnson: Can you revert the description to the original?

cmd/litestream/replicate.go
```diff
@@ -175,35 +200,38 @@ func (c *ReplicateCommand) Close() (err error) {
...
+Examples:
 
-	-no-expand-env
-	    Disables environment variable expansion in configuration file.
+	# Run replication and enable MCP server on port 8080.
+	$ litestream replicate -mcp -mcp-port 8080
 
-`[1:], DefaultConfigPath())
+`)
 }
```
benbjohnson: These CLI changes can be reverted too.

benbjohnson (changes requested):

cmd/litestream/mcp/mcp.go
corylanou: No, I just wasn't sure how much you wanted in the main folder.  Do you want all the tools and the mcp all in the `cmd/litestream`?  Just want to make sure before I move all them into main.

I've been beaten into submission enough times not to put stuff in main I think it was just a natural reaction.  But I get this is a super small project and probably doesn't matter that much.  I just won't tell the Go deities what we did here... :-).

cmd/litestream/mcp/tools/replicate.go
corylanou: yeah, that makes sense.  This tool should really be used to help you configure or diagnose your issues, not actually execute a replicate command.  I removed it.

cmd/litestream/replicate.go
corylanou: done

cmd/litestream/replicate.go
corylanou: ugg, I hate cursor sometimes.  I didn't even see it make that change...

cmd/litestream/replicate.go
corylanou: another aggressive cursor edit.  I should have caught this in review, sorry.

cmd/litestream/mcp/mcp.go
benbjohnson: Yeah, go ahead and throw it in `main`. You could even put them all in one `mcp` file. The tools themselves aren't very big.

cmd/litestream/main.go
```diff
@@ -176,8 +176,8 @@ type Config struct {
 	Logging LoggingConfig `yaml:"logging"`
 
 	// MCP server options
-	MCPEnabled bool `yaml:"mcp-enabled"`
-	MCPPort    int  `yaml:"mcp-port"`
+	MCPEnabled bool   `yaml:"mcp-enabled"`
+	MCPAddr    string `yaml:"mcp-addr"`
```
benbjohnson: You can drop `MCPEnabled` and just check if `MCPAddr` is non-blank to see if it's enabled. You'll need to take off the default addr though.

benbjohnson (approved):
A minor change requested but otherwise 👍

cmd/litestream/mcp/mcp.go
corylanou: collapsed it all to a single mcp.go file.

Issue #245, opened 2021-11-06
If -exec is used after the replicate positional arguments it throws a confusing error


simonw:
This works fine:
```
% litestream replicate \
  -exec "datasette -h 0.0.0.0 -p 8042 notebook.db" \
  notebook.db \
  "s3://my-litestream-bucket/notebook.db"
```
But if you put the `-exec` option AFTER the positional arguments (which is usually how I use CLI tools - positional arguments first, options second) you instead get a confusing error message:
```
% litestream replicate \
  notebook.db \
  "s3://my-litestream-bucket/notebook.db" \
  -exec "datasette -h 0.0.0.0 -p 8042 notebook.db"
litestream v0.3.6
replica url scheme required: -exec
```
I'd like it if `-exec` worked after the other arguments, but an alternative fix would be to improve that error message, maybe to:

```
-exec should be positioned before DB_PATH and REPLICA_URL
```

simonw:
I just spotted that `litestream replicate --help` does provide additional clues here:
```
% litestream replicate --help
The replicate command starts a server to monitor & replicate databases. 
You can specify your database & replicas in a configuration file or you can
replicate a single database file by specifying its path and its replicas in the
command line arguments.

Usage:

	litestream replicate [arguments]

	litestream replicate [arguments] DB_PATH REPLICA_URL [REPLICA_URL...]
```

benbjohnson:
Thanks for reporting the issue, @simonw. I'm surprised to see that Go's built-in `flag` package doesn't handle flags after the positional args (https://play.golang.org/p/uWQiOrWIK18). I always assumed it did.

I don't necessarily want to add in a third party dependency like [cobra](https://github.com/spf13/cobra) just to add support for this but I can fix up the error message so it's more clear. I could potentially rewrite the args list to move flags to the beginning but that seems like it could have edge cases.

alexpetros:
Want to +1, had the same issue and found the solution here.

corylanou:
This issue has been transferred to the litestream.io repository where documentation issues are now tracked: https://github.com/benbjohnson/litestream.io/issues/101

corylanou:
Transferred to litestream.io repository for documentation tracking.

PR #736, opened 2025-09-05
Improve error message when flags are positioned after positional arguments

corylanou:
## Summary
- Fixes issue #245 where `-exec` flag positioned after positional arguments produced a confusing error message
- Instead of "replica url scheme required: -exec", users now see: "flag \"-exec\" must be positioned before DB_PATH and REPLICA_URL arguments"

## Test plan
- [x] Built and tested the binary with flags in wrong position - shows clear error message
- [x] Tested with flags in correct position - works as expected  
- [x] All existing tests pass
- [x] Code passes go vet, go fmt, goimports, and staticcheck

### DIFF
--- a/cmd/litestream/replicate.go
+++ b/cmd/litestream/replicate.go
@@ -79,6 +79,10 @@ func (c *ReplicateCommand) ParseFlags(_ context.Context, args []string) (err err
 
 		dbConfig := &DBConfig{Path: fs.Arg(0)}
 		for _, u := range fs.Args()[1:] {
+			// Check if this looks like a flag that was placed after positional arguments
+			if len(u) > 0 && u[0] == '-' {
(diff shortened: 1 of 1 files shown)


cmd/litestream/replicate.go
```diff
@@ -79,6 +79,10 @@ func (c *ReplicateCommand) ParseFlags(_ context.Context, args []string) (err err
 
 		dbConfig := &DBConfig{Path: fs.Arg(0)}
 		for _, u := range fs.Args()[1:] {
+			// Check if this looks like a flag that was placed after positional arguments
+			if len(u) > 0 && u[0] == '-' {
+				return fmt.Errorf("flag %q must be positioned before DB_PATH and REPLICA_URL arguments", u)
+			}
```
benbjohnson: `strings.HasPrefix(u, "-")` is probably a little more idiomatic

PR #741, opened 2025-09-09
fix: Resolve flaky TestStore_Integration spurious failures

corylanou:
## Summary

Fixes a race condition in `TestStore_Integration` that was causing spurious test failures at exactly 10.01s.

## Root Cause

The test has a goroutine race condition where the insert goroutine calls `t.Errorf()` after the main test has completed:

1. **Main test goroutine**: Exits immediately when 10s timer closes the `done` channel
2. **Insert goroutine**: May still be mid-operation using `t.Context()`
3. **Race condition**: When main test exits, `t.Context()` gets cancelled
4. **Spurious failure**: Insert goroutine gets "database locked" error and calls `t.Errorf()` after test completion

In Go 1.14+, calling testing methods from a goroutine after the test has finished causes the test to fail.

## Evidence

- **Failed CI run**: https://github.com/benbjohnson/litestream/actions/runs/17583602203/job/49945685558
- **Failure pattern**: Always fails at exactly 10.01s (just after the 10s test duration)
- **Error logs**: `store_test.go:150: insert error: database is locked (5) (SQLITE_BUSY)`

## Solution

- **Error channel**: Capture insert errors from the goroutine
- **Shutdown awareness**: Check if `done` channel is closed before treating errors as failures
- **Defer cleanup**: Use `defer` to check error channel on any test exit path
- **Expected behavior**: Errors during shutdown (after `done` closes) are ignored as expected

## Test Results

The fix eliminates spurious failures while still catching legitimate insert errors that occur during normal test execution.

```bash
# Before: Intermittent failures at 10.01s
--- FAIL: TestStore_Integration (10.01s)

# After: Reliable passes
--- PASS: TestStore_Integration (10.18s)
```

## Changes

- Add `insertErr` channel to capture errors from insert goroutine
- Add shutdown-aware error handling: only report errors if not shutting down
- Use `defer` to ensure error checking happens on any test exit
- Maintain original clean loop structure with `return` statements

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -166,5 +166,5 @@ release:
 #       - APPLE_DEVELOPER_TEAM_ID={{ .Env.APPLE_TEAM_ID }}
 #       - AC_PASSWORD={{ .Env.AC_PASSWORD }}
 
-sboms:
-  - artifacts: archive
+# sboms:
+#   - artifacts: archive
--- a/store_test.go
+++ b/store_test.go
@@ -134,6 +134,19 @@ func TestStore_Integration(t *testing.T) {
 	done := make(chan struct{})
 	time.AfterFunc(10*time.Second, func() { close(done) })
 
+	// Channel for insert errors
+	insertErr := make(chan error, 1)
+
+	// Check for any insert errors that occurred before shutdown
+	defer func() {
+		select {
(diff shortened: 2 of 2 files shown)


corylanou:
Closing this PR to create a clean version without the goreleaser changes. See new PR coming up.

PR #760, opened 2025-09-24
Fix compaction against eventual consistency

corylanou:
## Summary

Fixes database corruption during compaction when using eventually consistent storage backends (e.g., Cloudflare R2, custom HTTP replica clients).

Resolves #725

## Problem

Remote replicas with eventual consistency can return a newly written LTX object before all bytes are fully available. When compaction attempts to read these incomplete files, it produces a corrupted merged snapshot, causing restore failures with errors like:
- `decode database: decode page 48: ... nonsequential page numbers`
- `decode page 2: EOF`

This was particularly affecting users with custom ReplicaClients writing to Cloudflare Workers/R2, causing:
- Intermittent replication failures
- Database restore errors on startup
- Corrupted snapshots after compaction

## Root Cause

`DB.Compact()` was always reopening input LTX files through the replica client, even when the complete file existed locally. On eventually consistent backends, this could race and return a truncated stream before the remote write fully propagated.

### 1. Primary Fix (db.go:1280-1294)
Compaction now preferentially reads from local disk when available:
- First attempts `os.Open(db.LTXPath(...))` for the local file
- Only falls back to `replica.Client.OpenLTXFile()` if the local file doesn't exist
- This ensures compaction always reads the fully written LTX file

### 2. Concurrency Fix (replica.go:217-221) 
Fixed a potential race condition in `Replica.SetPos()`:
- Changed from `RLock()` to `Lock()` for proper write exclusion
- Prevents concurrent readers/writers from racing on position updates

## Testing

Added comprehensive regression test (`TestStore_CompactDB_RemotePartialRead`) that:
- Simulates an eventually consistent replica client with 200ms delay
- Returns truncated LTX files during the delay period  
- Verifies compaction succeeds and restore works correctly
- Test fails on main branch (restore hits EOF) and passes with fix

All existing tests continue to pass: `go test ./...`

## Impact

- Operators using eventually consistent storage (S3, R2, custom HTTP) should recompact/resnapshot if they suspect corrupted LTX files
- The fix maintains full backward compatibility
- No configuration changes required

## Related Issues

Addresses symptoms reported in discussion #725:
- LTX files stopping transmission for several minutes
- Database restore failures on startup
- Compaction errors during initialization

### DIFF
--- a/db.go
+++ b/db.go
@@ -1277,6 +1277,17 @@ func (db *DB) Compact(ctx context.Context, dstLevel int) (*ltx.FileInfo, error) 
 			maxTXID = info.MaxTXID
 		}
 
+		// Prefer the on-disk LTX file when available so compaction is not subject
+		// to eventual consistency quirks of remote storage providers. Fallback to
+		// downloading the file from the replica client if the local copy has been
+		// removed.
+		if f, err := os.Open(db.LTXPath(srcLevel, info.MinTXID, info.MaxTXID)); err == nil {
+			rdrs = append(rdrs, f)
+			continue
+		} else if !os.IsNotExist(err) {
+			return nil, fmt.Errorf("open local ltx file: %w", err)
+		}
+
 		f, err := db.Replica.Client.OpenLTXFile(ctx, info.Level, info.MinTXID, info.MaxTXID, 0, 0)
 		if err != nil {
 			return nil, fmt.Errorf("open ltx file: %w", err)
--- a/replica.go
+++ b/replica.go
@@ -216,8 +216,8 @@ func (r *Replica) Pos() ltx.Pos {
 
 // SetPos sets the current replicated position.
 func (r *Replica) SetPos(pos ltx.Pos) {
-	r.mu.RLock()
-	defer r.mu.RUnlock()
+	r.mu.Lock()
+	defer r.mu.Unlock()
 	r.pos = pos
 }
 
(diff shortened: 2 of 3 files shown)


corylanou:
## Test Verification

I've confirmed the regression test properly validates this fix:

### Test fails on main branch:
```bash
$ git checkout main
$ git checkout fix-725 -- store_compaction_remote_test.go
$ go test -v ./... -run TestStore_CompactDB_RemotePartialRead

=== RUN   TestStore_CompactDB_RemotePartialRead
...
    store_compaction_remote_test.go:97: restore: decode database: decode page 2: EOF
--- FAIL: TestStore_CompactDB_RemotePartialRead (0.45s)
FAIL
```

The test reproduces the exact error reported in discussion #725: `decode page 2: EOF` during restore after compaction.

### Test passes with the fix:
```bash
$ git checkout fix-725
$ go test -v ./... -run TestStore_CompactDB_RemotePartialRead

=== RUN   TestStore_CompactDB_RemotePartialRead
...
--- PASS: TestStore_CompactDB_RemotePartialRead (0.34s)
PASS
```

The test simulates an eventually consistent storage backend (200ms delay) that returns partial LTX files during the delay window. Without the fix, compaction reads these incomplete files and creates corrupted snapshots. With the fix, compaction reads from the complete local file first, avoiding the issue entirely.

This confirms the fix correctly addresses the eventual consistency issues causing database corruption during compaction.

PR #761, opened 2025-09-24
Detect full checkpoints

benbjohnson:
## Description
This pull request adds detection for multiple salts in the WAL file.

## Motivation and Context
During normal operation, when the WAL rolls over there should only be two sets of salts—the old one and the new one. However, if Litestream stops while SQLite continues to receive writes and `FULL` or `RESTART` checkpoints occur, we did not previously detect this.

This pull request checks for any unexpected salts and issues a full snapshot for the next transaction.

Note that this does not fix all cases and it is recommended to turn off WAL autocheckpoint inside your application:

```
PRAGMA wal_autocheckpoint = 0;
 ```

## How Has This Been Tested?
This is tested using the `reproduce-critical-bug`.sh` script.

## Types of changes

## Checklist

### CODE REVIEW

File: wal_reader_test.go [hunk 1]
```diff
@@ -245,3 +245,34 @@ func TestWALReader(t *testing.T) {
...
+			t.Fatalf("len(m)=%d, want %d", got, want)
+		}
+		if _, ok := m[[2]uint32{0x1b9a294b, 0x37f91916}]; !ok {
+			t.Fatalf("salt 0 not found")
+		}
+		if _, ok := m[[2]uint32{0x1b9a294a, 0x031f195e}]; !ok {
+			t.Fatalf("salt 1 not found")
+		}
+		if _, ok := m[[2]uint32{0x1b9a2949, 0x13b3dd67}]; !ok {
+			t.Fatalf("salt 2not found")
```
corylanou:
```suggestion
			t.Fatalf("salt 2 not found")
```

db.go
```diff
@@ -689,74 +690,127 @@ func (db *DB) verify(ctx context.Context) (info syncInfo, err error) {
...
+	if err != nil {
+		return false, fmt.Errorf("frame salts until: %w", err)
+	}
+
+	// Remove known salts from the map.
+	delete(m, knownSalts[0])
+	delete(m, knownSalts[1])
+
+	// If we have more than one unknown salt, then we have a FULL or RESTART checkpoint.
+	return len(m) > 1, nil
```
corylanou: based on the comment I'd expect `>=` but not sure.  Just double checking 

```suggestion
	return len(m) >=1, nil
```

corylanou: AI Also questioned this with this comment:

> After deleting 2 known salts, checking for > 1 means you need 4+ unique salts to detect
   an issue. Should be >= 1 to detect any unexpected salt.

db.go
```diff
@@ -689,74 +690,127 @@ func (db *DB) verify(ctx context.Context) (info syncInfo, err error) {
...
+	if err != nil {
+		return false, fmt.Errorf("open wal file: %w", err)
+	}
+	defer walFile.Close()
+
+	rd, err := NewWALReader(walFile, db.Logger)
+	if err != nil {
+		return false, fmt.Errorf("new wal reader: %w", err)
+	}
+	m, err := rd.FrameSaltsUntil(ctx, knownSalts[len(knownSalts)-1])
```
corylanou: Should we check `len(knownSalts)` here to ensure we don't get an index out of range and panic at runtime?

wal_reader.go
```diff
@@ -235,6 +235,32 @@ func (r *WALReader) PageMap(ctx context.Context) (m map[uint32]int64, maxOffset
 	return m, end, commit, nil
 }
 
+// FrameSaltsUntil returns a set of all unique frame salts in the WAL file.
+func (r *WALReader) FrameSaltsUntil(ctx context.Context, until [2]uint32) (map[[2]uint32]struct{}, error) {
+	m := make(map[[2]uint32]struct{})
+	for offset := int64(WALHeaderSize); ; offset += int64(WALFrameHeaderSize + r.pageSize) {
```
corylanou, line 241: I don't know how intensive this loop is, and how much you want to check context, but perhaps checking `ctx.Done()` inside the loop?  Not sure if it matters or not here.

wal_reader.go
```diff
@@ -235,6 +235,32 @@ func (r *WALReader) PageMap(ctx context.Context) (m map[uint32]int64, maxOffset
...
+		hdr := make([]byte, WALFrameHeaderSize)
+		if n, err := r.r.ReadAt(hdr, offset); n != len(hdr) {
+			break
```
corylanou, line 244: Is this an error condition as well?  Should we log a debug or something if we need to understand this behavior in the future or when we get logs in for debugging?

db.go
```diff
@@ -689,74 +690,127 @@ func (db *DB) verify(ctx context.Context) (info syncInfo, err error) {
...
-	fsalt2 := binary.BigEndian.Uint32(buf[12:])
+	pgno := binary.BigEndian.Uint32(frame[0:])
+	fsalt1 := binary.BigEndian.Uint32(frame[8:])
+	fsalt2 := binary.BigEndian.Uint32(frame[12:])
+	data := frame[WALFrameHeaderSize:]
 
 	if fsalt1 != dec.Header().WALSalt1 || fsalt2 != dec.Header().WALSalt2 {
-		info.reason = "frame salt mismatch, wal overwritten by another process"
-		return info, nil
+		// info.reason = "frame salt mismatch, wal overwritten by another process"
```
corylanou: Delete?

db.go
benbjohnson: Ah, good catch! Thanks! :)

db.go
benbjohnson: That's probably a good idea to be safe.

wal_reader.go
benbjohnson: It's generally going to be quick but a WAL could get quite large so a context check isn't a bad idea.

wal_reader.go
benbjohnson: WAL frames are only valid if the whole frame exists and its contents match the rolling checksum. So here we're iterating over the WAL until we hit our last known salt or the end of file which could have a torn write and give us a partial header (but that's not an error condition).

Issue #772, opened 2025-10-01
Read-replica VFS issues.


ncruces:
## Environment

**Litestream version:**

```text
0.5.0
```

**Operating system & version:**
Debian GNU/Linux forky

**Installation method:**
binary download, built from source

**Storage backend:**
file

## Bug Description

This may be way too early, but I was trying to "port" the (just released) [read-replica VFS](https://github.com/benbjohnson/litestream/blob/v0.5.0/vfs.go) to my [SQLite driver](https://github.com/ncruces/go-sqlite3/tree/litestream/litestream), and I've bumped into a few issues, which may cause you future pain, so I decided to open this to discuss them here.

If it _is_ too early, feel free to ignore me and close the issue, I genuinely just wanted to help.
Hopefully this reads "helpful" and not "demanding".

### VFS Open

`Open` is trying to handle all file types using the `ReplicaClient`. This may seem that it mostly works but, depending on settings, SQLite will eventually try to use the VFS to open temporary files.

IIUC, for Litestream, it should be sufficient to open [`SQLITE_OPEN_MAIN_DB`](https://www.sqlite.org/c3ref/c_open_autoproxy.html) files and refuse everything else.

However, unless the user is instructed to configure SQLite to use memory for all temporary files, SQLite will eventually try to at least `SQLITE_OPEN_TEMP_JOURNAL` to handle large sorts (such as an `ORDER BY` of millions of records). Small tests won't surface this, though.

Also, `Open` can probably or `SQLITE_OPEN_READONLY` to the flags it returns.

### Cancel polling replication

This is probably an oversight, `wg`, `ctx`, and `cancel` are all there to handle this, but unused.

### Handle sub-page reads

Since you're not returning [`SQLITE_IOCAP_SUBPAGE_READ`](https://www.sqlite.org/c3ref/c_iocap_atomic.html), SQLite 3.48.0 and above will mostly avoid sub-page reads (otherwise, overflow BLOBs would use them). However, SQLite **will** use sub-page to read the database header in the first page, and your code here is buggy.

SQLite will read 100 bytes at offset 0 to load the entire header and, when in rollback mode, **will read 16 bytes at offset 24 to check for changes**.

[...]

PR #786, opened 2025-10-09
Update retention enforcement to use minTXID

ncruces:
## Description

Update retention enforcement to use `minTXID`, as described in the existing comment.

## Motivation and Context

Using `minTXID` matches the comment and contributes to fix the issue described in https://github.com/benbjohnson/litestream/issues/772#issuecomment-3380771816.

Using `maxTXID` causes issues with doing `litestream restore` concurrently `litestream replicate` (restoring to a "replica" while the "master" is continuously replicating): immediately after `CalcRestorePlan` creates a plan for "now", the L0 files needed can get deleted by a L0->L1 compaction.

This would also impact a read-replica VFS.

## How Has This Been Tested?

I tested this by creating a database with to which I continuously make changes, that is replicated with `litestream replicate fruits.db file:backup/fruits.db` and verifying that under L0->L1 compactions enough L0 files are kept to allow concurrent `litestream restore` operations to conclude at any recent TXID.

Only under a LN->L9 snapshot might necessary files be deleted, but that's avoided by increasing snapshot retention, which should at least be twice the snapshot interval for "concurrent restore" and "read-replica VFS" to work.

## Types of changes

## Checklist

### DIFF
--- a/db.go
+++ b/db.go
@@ -1381,7 +1381,7 @@ func (db *DB) Compact(ctx context.Context, dstLevel int) (*ltx.FileInfo, error) 
 
 	// If this is L1, clean up L0 files that are below the minTXID.
 	if dstLevel == 1 {
-		if err := db.EnforceRetentionByTXID(ctx, 0, maxTXID); err != nil {
+		if err := db.EnforceRetentionByTXID(ctx, 0, minTXID); err != nil {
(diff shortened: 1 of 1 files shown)


benbjohnson:
Thanks!

PR #801, opened 2025-10-20
Avoid unnecessary import of SQLite driver.

ncruces:
## Description

Avoids unnecessarily importing the modernc SQLite driver.
The package is still imported by the main and test packages that need it.

## Motivation and Context

Library users (such as the VFS) don't need to import the driver, which registers it on `init`.

See https://github.com/benbjohnson/litestream/issues/772#issuecomment-3419981135

For example, the package used by the current prototype `vfs.go` is meant to work with `cgo` drivers, not `modernc`:
https://github.com/benbjohnson/litestream/blob/2ee408b66bb78ca3ff52fc867ca66f3c8b154e6b/vfs.go#L15

In fact, I'm not sure a VFS can be implemented for `modernc`. My reading of their [`vfs`](https://pkg.go.dev/modernc.org/sqlite/vfs) package says no, but I'm sure with some effort, dropping down to the [lower level](https://pkg.go.dev/modernc.org/sqlite/lib#Xsqlite3_vfs_register) allows it.

## How Has This Been Tested?

I've tested the `litestream` command, and it works, which is expected, as it imports `modernc` and thus registers the driver (i.e., no change):
https://github.com/benbjohnson/litestream/blob/2ee408b66bb78ca3ff52fc867ca66f3c8b154e6b/cmd/litestream/main.go#L24

All tests pass, as both the `litestream_test` package, and `testingutil` import the driver.

Library clients will have themselves to import `modernc`: AFAICT if they use `litestream.DB`, `litestream.Replica` or `litestream.Store`.

## Types of changes

- [x] Breaking change (would cause existing functionality to not work as expected)

**This _may_ break library users.** It's a simple fix, though, just add this somewhere in their code:
```go
import _ "modernc.org/sqlite"
```

## Checklist

**I haven't updated docs.** Where would docs for library users be? This should probably me mentioned in release notes.

### DIFF
--- a/db.go
+++ b/db.go
@@ -20,7 +20,6 @@ import (
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/client_golang/prometheus/promauto"
 	"github.com/superfly/ltx"
-	"modernc.org/sqlite"
 
 	"github.com/benbjohnson/litestream/internal"
 )
@@ -334,8 +333,13 @@ func (db *DB) setPersistWAL(ctx context.Context) error {
 	}
 	defer conn.Close()
 
-	return conn.Raw(func(driverConn interface{}) error {
-		fc, ok := driverConn.(sqlite.FileControl)
+	return conn.Raw(func(driverConn any) error {
+		// Using an anonymous interface here avoids unnecessarily importing our
+		// chosen driver. This is important for library users (e.g. the VFS)
+		// that don't need to depend on the driver.
+		fc, ok := driverConn.(interface {
+			FileControlPersistWAL(string, int) (int, error)
+		})
 		if !ok {
 			return fmt.Errorf("driver does not implement FileControl")
 		}
--- a/internal/testingutil/testingutil.go
+++ b/internal/testingutil/testingutil.go
@@ -17,6 +17,7 @@ import (
 
 	sftpserver "github.com/pkg/sftp"
 	"golang.org/x/crypto/ssh"
+	_ "modernc.org/sqlite"
 
 	"github.com/benbjohnson/litestream"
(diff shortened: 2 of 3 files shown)


ncruces:
I can achieve close to the same with a replace directive:
https://github.com/ncruces/go-sqlite3/blob/8f9a6ca/litestream/go.mod#L63

But it's not as good. The upshot is that with a small change in my driver, Litestream seems to work fine based entirely on my driver (not that switching is a good idea, it's not):
https://github.com/ncruces/go-sqlite3/blob/8f9a6ca/driver/driver.go#L445-L459

ncruces:
This, and the move to 1.25 in #855 make it hard(er) to use Litestream as a library.

I'm going to drop this for now and point dependencies at my fork.

I think a fork is unfortunate, so if there's interest in making Litestream more usable as a library, please ping me. I'd like to help.

benbjohnson:
@ncruces Sorry about that. We reverted the changes to move it to 1.24. https://github.com/benbjohnson/litestream/pull/892

ncruces:
Don't worry. For now, you're moving fast, I am too.

It makes sense that you don't treat Litestream as a library; it's a service that you run as a command, and now also an .so that doesn't offer a Go API.

Adding more use cases now will only slow you down. But once/if you feel it makes sense, I'm willing to help.

My fork for the most part, only deletes code. And it's good enough to configure and run a VFS programmatically with my driver.

Ideally I'd also like to allow in-process streaming replication (also configured programmatically), but on most platforms, you're better running the sidecar service. The sample code I found for it works, but doesn't handle compactions/snapshoting and I didn't come up with a nice API, so I dropped it.

PR #886, opened 2025-12-10
fix(store): handle uninitialized db gracefully during compaction

### LINKED ISSUE #877: GCS support broken?
## Bug Description

It seems to me that using GCS as a backend for litestream replication is broken. If I run through the "Getting Started" example in S3 all is well, if I run the same vs GCS I see errors.

Separately, the doc page for GCS lists url's as "gcs://bucket/path" when it should be "gs://bucket/path"

## Environment

**Litestream version:**
```text
v0.5.2
```

Ubuntu 24.04

**Installation method:**
Docker

**Storage backend:**
GCS

## Steps to Reproduce

```sh
↳ sqlite3 mydbgcs.sqlite
SQLite version 3.51.1 2025-11-28 17:28:25
Enter ".help" for usage hints.
sqlite> PRAGMA busy_timeout = 5000;
PRAGMA journal_mode=WAL;
PRAGMA foreign_keys = ON;
PRAGMA synchronous = NORMAL;
5000
wal
sqlite> CREATE TABLE fruits (name TEXT, color TEXT);
sqlite> INSERT INTO fruits (name, color) VALUES ('apple', 'red');
INSERT INTO fruits (name, color) VALUES ('banana', 'yellow');
sqlite> select * from fruits;
apple|red
banana|yellow
sqlite>
```

then replicate:
```sh
$ litestream replicate mydbgcs.sqlite gs://mybucket/backups/litestream/mydb
time=2025-12-08T12:37:15.868Z level=INFO msg=litestream version="(development build)" level=""
time=2025-12-08T12:37:15.868Z level=INFO msg="initialized db" path=/tmp/backup-test2/mydbgcs.sqlite
time=2025-12-08T12:37:15.868Z level=INFO msg="replicating to" type=gs sync-interval=1s bucket=mybucket path=backups/litestream/mydb
time=2025-12-08T12:37:15.868Z level=INFO msg="starting compaction monitor" level=9 interval=24h0m0s
time=2025-12-08T12:37:15.868Z level=INFO msg="starting compaction monitor" level=2 interval=5m0s
time=2025-12-08T12:37:15.868Z level=INFO msg="starting compaction monitor" level=1 interval=30s
time=2025-12-08T12:37:15.868Z level=INFO msg="starting compaction monitor" level=3 interval=1h0m0s
time=2025-12-08T12:37:16.017Z level=ERROR msg="compaction failed" level=9 error="page size not initialized yet"
time=2025-12-08T12:37:30.446Z level=INFO msg="compaction complete" level=1 txid.min=0000000000000001 txid.max=0000000000000001 size=651
time=2025-12-08T12:39:00.305Z level=INFO msg="deleting ltx file" db=mydbgcs.sqlite level=0 minTXID=0000000000000001 maxTXID=0000000000000001
time=2025-12-08T12:39:00.464Z level=INFO msg="compaction complete" level=1 txid.min=0000000000000002 txid.max=0000000000000002 size=223
time=2025-12-08T12:40:00.130Z level=ERROR msg="compaction failed" level=2 error="write ltx file: extract timestamp from LTX header: EOF"
```

And replicate:
```sh
$ litestream restore -o mydb3.sqlite gs://mybucket/backups/litestream/mydb
time=2025-12-08T12:39:18.296Z level=ERROR msg="failed to run" error="decode database: decode header: EOF"
```

The same steps vs S3 works fine.

**Expected behavior:**
It should replicate and restore without errors.

**Actual behavior:**
Errors in replication as per above, and can't restore.

fwiw I do see files created in GCS:

Image

## Configuration
Using the CLI as per above, and setting GOOGLE_APPLICATION_CREDENTIALS to a suitable service account.

## Logs

I included the log output above with the steps.

corylanou:
## Summary

- Add `ErrDBNotReady` sentinel error to handle the race condition where level 9 compaction fires immediately at startup before the database page size has been initialized
- This eliminates the confusing ERROR log `"page size not initialized yet"` that appeared during startup
- The compaction monitor now handles this condition like other skip conditions (`ErrNoCompaction`, `ErrCompactionTooEarly`) with a debug log instead of an error

## Root Cause

Race condition between DB initialization and compaction monitor startup:
1. `Store.Open()` starts compaction monitors with `time.NewTimer(time.Nanosecond)` (fires immediately)
2. Level 9 compaction attempts to run before `db.Sync()` → `db.init()` sets the page size
3. Result: ERROR log that confuses users even though it's benign

## Changes

- `store.go`: Added `ErrDBNotReady` sentinel error, early `PageSize()` check in `CompactDB()`, and handler in `monitorCompactionLevel()`
- `store_test.go`: Added regression test `TestStore_CompactDB/DBNotReady`

## Test plan

- [x] Added unit test that reproduces the issue and verifies the fix
- [x] All existing tests pass (`go test ./...`)
- [x] Pre-commit hooks pass

Fixes #877

### DIFF
--- a/store.go
+++ b/store.go
@@ -23,6 +23,10 @@ var (
 
 	// ErrTxNotAvailable is returned when a transaction does not exist.
 	ErrTxNotAvailable = errors.New("transaction not available")
+
+	// ErrDBNotReady is returned when compaction is attempted before the
+	// database has been initialized (e.g., page size not yet known).
+	ErrDBNotReady = errors.New("db not ready")
 )
 
 // Store defaults
@@ -199,6 +203,9 @@ LOOP:
 				} else if errors.Is(err, ErrCompactionTooEarly) {
 					slog.Debug("recently compacted, skipping", "level", lvl.Level, "path", db.Path())
 					continue
+				} else if errors.Is(err, ErrDBNotReady) {
+					slog.Debug("db not ready, skipping", "level", lvl.Level, "path", db.Path())
+					continue
 				} else if err != nil {
 					// Don't log or sleep on context cancellation errors during shutdown
 					if !errors.Is(err, context.Canceled) && !errors.Is(err, context.DeadlineExceeded) {
@@ -250,6 +257,11 @@ LOOP:
 // CompactDB performs a compaction or snapshot for a given database on a single destination level.
 // This function will only proceed if a compaction has not occurred before the last compaction time.
 func (s *Store) CompactDB(ctx context.Context, db *DB, lvl *CompactionLevel) (*ltx.FileInfo, error) {
+	// Skip if database is not yet initialized (page size unknown).
+	if db.PageSize() == 0 {
+		return nil, ErrDBNotReady
+	}
+
 	dstLevel := lvl.Level
 
 	// Ensure we are not re-compacting before the most recent compaction time.
--- a/store_test.go
+++ b/store_test.go
@@ -96,6 +96,31 @@ func TestStore_CompactDB(t *testing.T) {
 			t.Fatalf("unexpected error: %s", err)
 		}
 	})
+
+	// Regression test for GitHub issue #877: level 9 compaction fails with
+	// "page size not initialized yet" error when attempted before DB initialization.
+	t.Run("DBNotReady", func(t *testing.T) {
+		db0 := testingutil.MustOpenDB(t)
+		defer testingutil.MustCloseDB(t, db0)
+
+		levels := litestream.CompactionLevels{
+			{Level: 0},
+			{Level: 1, Interval: 100 * time.Millisecond},
+		}
+		s := litestream.NewStore([]*litestream.DB{db0}, levels)
+		s.CompactionMonitorEnabled = false
+		if err := s.Open(t.Context()); err != nil {
+			t.Fatal(err)
+		}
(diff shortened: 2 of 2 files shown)


benbjohnson:
@corylanou Will this re-run the compaction soon after the database is ready or will it wait for the full interval for the compaction level?

corylanou:
> @corylanou Will this re-run the compaction soon after the database is ready or will it wait for the full interval for the compaction level?

@benbjohnson  Current behavior: It will wait for the full interval. The timer is reset at the start of the compaction loop (before iterating DBs), so if ErrDBNotReady is returned, the next attempt won't happen
   until the scheduled interval:
  - L1: 30s
  - L2: 5m
  - L3: 1h
  - L9 (snapshot): 24h

  This matches the previous behavior - the original error also resulted in waiting the full interval (just with an ERROR log first).

  Options:

  1. Keep as-is - The snapshot will eventually run at the next 24h interval. Since the DB monitor runs every 1s by default, L1 compaction (30s) will likely succeed shortly after startup anyway, and
   the data is safe once L1 runs.
  2. Add short retry for ErrDBNotReady - If any DB returns ErrDBNotReady, reset timer to ~1-2 seconds instead of the full interval. This ensures compaction runs promptly after DB initialization.

  Let me know which you'd prefer, or if you have a different approach in mind.

PR #888, opened 2025-12-11
fix(store): retry compaction when db not ready instead of waiting full interval

corylanou:
## Summary

Follow-up to #886. When a database returns `ErrDBNotReady`, the compaction monitor now retries after 1 second instead of waiting for the full compaction interval (which could be 24 hours for snapshots).

## Changes

- Track when any DB returns `ErrDBNotReady` during compaction iteration
- Use short retry interval (1s) if retry needed, up to 30s timeout
- Log warning if timeout exceeded (DB never initialized)
- Reset retry counter and use normal interval after success

## Behavior

**Before:** If DB not ready at startup, wait full interval (L9: 24h, L1: 30s, etc.)

**After:** Retry every 1 second until DB is ready (max 30s timeout), then use normal interval

## Test plan

- [x] All existing tests pass (`go test ./...`)
- [x] Pre-commit hooks pass

Related to #877

### DIFF
--- a/store.go
+++ b/store.go
@@ -264,7 +264,6 @@ func (s *Store) monitorCompactionLevel(ctx context.Context, lvl *CompactionLevel
 	slog.Info("starting compaction monitor", "level", lvl.Level, "interval", lvl.Interval)
 
 	retryDeadline := time.Time{}
-	var errBackoff time.Duration
 	timer := time.NewTimer(time.Nanosecond)
 	defer timer.Stop()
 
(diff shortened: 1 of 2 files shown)


store.go
```diff
@@ -263,53 +263,107 @@ func (s *Store) SnapshotLevel() *CompactionLevel {
...
+
+	next := prev * 2
+	if next < min {
+		next = min
+	}
+	if next > max {
+		next = max
+	}
+	return next
+}
```
benbjohnson: I don't think we need exponential backoff here. Checking if the DB is ready is not expensive. Can you change this to check every second and keep it simple?

benbjohnson (changes requested):

store.go
corylanou: I had that originally, but wasn't sure.  I'll revert it back.  Thanks!

PR #901, opened 2025-12-15
fix(db): handle WAL offset at header boundary in verify

### LINKED ISSUE #900: cannot verify wal state: prev WAL offset is less than the header size: -4088
Litestream fails with `cannot verify wal state: prev WAL offset is less than the header size: -4088` when an LTX file has `WALOffset=WALHeaderSize (32)` and `WALSize=0`.

The verify function in `db.go` calculates `prevWALOffset = info.offset - frameSize` before checking boundary conditions. When `info.offset == 32` and `frameSize == 4120` (4096 page + 24 frame header), this results in `-4088`, triggering the error at line 896.

## Environment

**Litestream version:**
```text
0.5.3
```

Linux

## Logs

```text
cannot verify wal state: prev WAL offset is less than the header size: -4088
```

corylanou:
## Summary

Fixes #900 - `prev WAL offset is less than the header size: -4088`

When an LTX file has `WALOffset=32` (WALHeaderSize) and `WALSize=0`, the verify function calculated `prevWALOffset = 32 - 4120 = -4088`, which triggered the error.

### Root Cause

The code calculated `prevWALOffset = info.offset - frameSize` before checking if `info.offset` was already at the WAL header boundary. When `info.offset == WALHeaderSize`, there's no previous frame to reference.

### Fix

Add an early check for `info.offset == WALHeaderSize` before computing `prevWALOffset`, handling the edge case where an LTX file represents a state at the beginning of the WAL with no frames written.

## Test plan

- [x] Added regression test `TestDB_Verify_WALOffsetAtHeader` that reproduces the bug
- [x] All existing tests pass
- [x] Pre-commit hooks pass

### DIFF
--- a/db.go
+++ b/db.go
@@ -885,6 +885,21 @@ func (db *DB) verify(ctx context.Context) (info syncInfo, err error) {
 	salt2 := binary.BigEndian.Uint32(hdr0[20:])
 	saltMatch := salt1 == dec.Header().WALSalt1 && salt2 == dec.Header().WALSalt2
 
+	// Handle edge case where we're at WAL header (WALOffset=32, WALSize=0).
+	// This can happen when an LTX file represents a state at the beginning of the WAL
+	// with no frames written yet. We must check this before computing prevWALOffset
+	// to avoid underflow (32 - 4120 = -4088).
+	// See: https://github.com/benbjohnson/litestream/issues/900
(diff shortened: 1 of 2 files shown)


corylanou:
bugbot run

benbjohnson (approved):

PR #905, opened 2025-12-16
feat(s3): add X-Tigris-Consistent header for Tigris endpoints

### LINKED ISSUE #903: Set X-Tigris-Consistent header to 'true' for Tigris endpoints in S3 replica client
## Summary

When using the S3 replica client with Tigris object storage, the `X-Tigris-Consistent` HTTP header should be set to `"true"` to ensure strong read-after-write consistency.

## Background

Tigris is an S3-compatible object storage service that offers strong consistency when the `X-Tigris-Consistent: true` header is included in requests. Without this header, Tigris may return eventually consistent results, which could cause issues during replication and restore operations.

## Current Behavior

Litestream already detects Tigris endpoints via `IsTigrisEndpoint()` and applies Tigris-specific defaults:
- `SignPayload` is set to `true`
- `RequireContentMD5` is set to `false`

However, the consistency header is not currently set.

## Proposed Change

Add middleware in `s3/replica_client.go` to set the `X-Tigris-Consistent: true` header on all requests when a Tigris endpoint is detected. This could be done in the `middlewareOption()` function alongside the existing User-Agent middleware.

## Implementation Notes

- The `isTigris` boolean is already computed in `NewReplicaClientFromURL()`
- A new field (e.g., `IsTigris bool`) could be added to `ReplicaClient` to track this
- The header should be added in the Build middleware phase, similar to the User-Agent header

## References

- Tigris consistency documentation: https://www.tigrisdata.com/docs/concepts/consistency/
- Existing Tigris handling: `s3/replica_client.go:157-188`

corylanou:
## Summary

- Add `IsTigris` field to `ReplicaClient` struct to track Tigris endpoint detection
- Set `X-Tigris-Consistent: true` header on all S3 requests when using Tigris endpoints
- Enable strong read-after-write consistency for Tigris object storage

## Background

Tigris is an S3-compatible object storage service that offers strong consistency when the `X-Tigris-Consistent: true` header is included in requests. Without this header, Tigris may return eventually consistent results, which could cause issues during replication and restore operations.

## Implementation

The implementation follows the existing pattern used for the User-Agent middleware:
1. Added `IsTigris bool` field to `ReplicaClient` struct
2. Set `IsTigris = true` in both `NewReplicaClientFromURL()` and `NewS3ReplicaClientFromConfig()` when Tigris endpoint is detected
3. Added middleware in `middlewareOption()` to set the `X-Tigris-Consistent: true` header on all requests when `IsTigris` is true

## Test plan

- [x] Added unit test `TestReplicaClient_TigrisConsistentHeader` that verifies:
  - Header is set to "true" when `IsTigris` is enabled
  - Header is not set when `IsTigris` is disabled
- [x] All existing tests pass
- [x] Pre-commit hooks pass (go-vet, staticcheck, goimports)

Closes #903

### DIFF
--- a/cmd/litestream/main.go
+++ b/cmd/litestream/main.go
@@ -1307,6 +1307,7 @@ func NewS3ReplicaClientFromConfig(c *ReplicaConfig, _ *litestream.Replica) (_ *s
 
 	client.SignPayload = signSetting.value
 	client.RequireContentMD5 = requireSetting.value
+	client.IsTigris = isTigris
 
 	// Apply upload configuration if specified.
 	if c.PartSize != nil {
--- a/s3/replica_client.go
+++ b/s3/replica_client.go
@@ -80,6 +80,7 @@ type ReplicaClient struct {
 	SkipVerify        bool
 	SignPayload       bool
 	RequireContentMD5 bool
+	IsTigris          bool
 
 	// Upload configuration
 	PartSize    int64 // Part size for multipart uploads (default: 5MB)
@@ -181,6 +182,7 @@ func NewReplicaClientFromURL(scheme, host, urlPath string, query url.Values, use
 	// Apply provider-specific defaults for S3-compatible providers.
 	if isTigris {
 		// Tigris: requires signed payloads, no MD5
+		client.IsTigris = true
 		if !signPayloadSet {
 			signPayload, signPayloadSet = true, true
 		}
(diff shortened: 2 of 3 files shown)


corylanou:
bugbot run

chrismccord:
note, we also need to check for `t3.storage.dev` here: https://github.com/benbjohnson/litestream/blob/feb65fb2a165ee17ec856b0d27e03907c17588ae/replica_url.go#L201

PR #908, opened 2025-12-16
fix(claude): move personal hooks to local settings

corylanou:
## Summary

- Remove hooks configuration from tracked `.claude/settings.json` since the hook scripts in `.claude/hooks/` are gitignored
- This fixes errors for contributors who have the settings file but not the hook files
- Also fixes incorrect project name/description that was copied from another project

Fixes the error Ben was seeing:
```
Stop hook error: Failed with non-blocking status code: /bin/sh:
/Users/benbjohnson/src/benbjohnson/litestream/.claude/hooks/session-summary.sh: No such file or directory
```

## Test plan

- [x] Verify settings.json no longer references hook files
- [ ] Confirm Ben no longer sees the hook error after pulling this change

### DIFF
--- a/.claude/settings.json
+++ b/.claude/settings.json
@@ -1,70 +1,6 @@
 {
-  "project_name": "Planning Center Sync",
-  "description": "Planning Center Sync project for Parable",
-
-  "hooks": {
-    "PostToolUse": [
-      {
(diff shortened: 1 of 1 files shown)


benbjohnson:
works for me

PR #921, opened 2025-12-18
docs(examples): add library usage examples

corylanou:
## Summary

Add examples showing how to use Litestream programmatically as a Go library:
- `_examples/library/basic/` - minimal file backend example  
- `_examples/library/s3/` - S3 backend with restore-on-startup pattern
- `_examples/library/README.md` - API overview and usage instructions

Addresses #440

## Key Features Demonstrated

- **Store-based lifecycle**: Uses `Store` to wrap `DB` for proper compaction support
- **CompactionLevels**: Shows how to configure L0 + L1 compaction
- **App-owned SQLite connection**: Opens separate `sql.DB` with WAL mode and busy_timeout (avoids nil pointer issues before Litestream initializes)
- **Restore-on-startup pattern**: Check for local DB, restore from backup if missing
- **Graceful shutdown**: Proper `store.Close()` handling

## API Pattern

```go
// 1. Create Litestream DB wrapper
db := litestream.NewDB(path)

// 2. Create and attach replica
client := file.NewReplicaClient(replicaPath)
replica := litestream.NewReplicaWithClient(db, client)
db.Replica = replica
client.Replica = replica // file backend: preserves permissions

// 3. Create Store with compaction levels
levels := litestream.CompactionLevels{
    {Level: 0},
    {Level: 1, Interval: 10 * time.Second},
}
store := litestream.NewStore([]*litestream.DB{db}, levels)

// 4. Open Store (starts replication + compaction monitors)
store.Open(ctx)
defer store.Close(context.Background())

// 5. Open app's own SQLite connection
sqlDB, _ := sql.Open("sqlite", path)
sqlDB.ExecContext(ctx, `PRAGMA journal_mode = wal;`)
sqlDB.ExecContext(ctx, `PRAGMA busy_timeout = 5000;`)
```

## Notes

- README includes API stability warning (library API not considered stable)
- Uses `modernc.org/sqlite` driver (pure Go, no CGO required)
- Handles both `ErrTxNotAvailable` and `ErrNoSnapshots` for restore fallback

## Test plan

- [x] Both examples compile
- [x] All pre-commit hooks pass
- [x] Markdown linting passes

### DIFF
--- a/.github/workflows/commit.yml
+++ b/.github/workflows/commit.yml
@@ -107,8 +107,8 @@ jobs:
 
       - run: go install ./cmd/litestream
 
-      - run: go build ./_examples/library/basic
-      - run: go build ./_examples/library/s3
+      - run: go build -o /dev/null ./_examples/library/basic
+      - run: go build -o /dev/null ./_examples/library/s3
       - run: go test -v ./_examples/library
 
       - run: go test -v .


benbjohnson:
@corylanou You'll need to instantiate a `Store` instance since compaction has been moved to that type. That will then store a `DB` inside of it.

benbjohnson:
Should this go on a page on `litestream.io` as well?

corylanou:
> Should this go on a page on `litestream.io` as well?

Yes, I will write up a litestream doc as well.  Also, I'm going to add a CI test to make sure 1) This always builds so it doesn't become stale.  2) A small integration test that uses the same exact model the main.go does so that we know this really does what it says it does (just building alone doesn't get us that).  I know that the API will change, but this should be minimal effort to keep these examples valid even when it does so that people can continue to use this as a library

ncruces:
This is a bad pattern:
```go
_, err := db.ExecContext(ctx, `PRAGMA busy_timeout = 5000;`)
```

An `sql.DB` is a pool of connections; this sets the `PRAGMA` for **one** random connection from the pool. The better pattern for _most_ `PRAGMA`s (using modernc) is:
https://github.com/benbjohnson/litestream/blob/3480067f4b7577b2a5295c5d1a4d8acb1796b16e/db.go#L401-L402

But you really need to rethink these.

E.g., this one may be fine, because WAL mode is persistent across connections:
https://github.com/benbjohnson/litestream/blob/3480067f4b7577b2a5295c5d1a4d8acb1796b16e/db.go#L429-L436

This one is fine too, because not only is page size persistent, but you're just reading the value:
https://github.com/benbjohnson/litestream/blob/3480067f4b7577b2a5295c5d1a4d8acb1796b16e/db.go#L461-L466

This one, however, isn't, it will apply to only **one** connection from the pool:
https://github.com/benbjohnson/litestream/blob/3480067f4b7577b2a5295c5d1a4d8acb1796b16e/db.go#L438-L441

If you want this to apply to all connections, move it to the DSN. If you want it to apply to a specific connection where you'll do something, you need to grab an `sql.Conn`.

My suggestion is to do:
```go
	dsn := fmt.Sprintf("file:%s?"+
		"_pragma=busy_timeout(%d)&"+
		"_pragma=journal_mode(wal)&"+
		"_pragma=wal_autocheckpoint(0)",
		db.path, db.BusyTimeout.Milliseconds())
```

ncruces:
Also, with modernc/mattn all POSIX platforms use POSIX locks (this is not macOS specific):
https://github.com/benbjohnson/litestream/blob/3480067f4b7577b2a5295c5d1a4d8acb1796b16e/_examples/library/README.md?plain=1#L12

There's also no way to control which driver Litestream uses, so… the advice here is that you **must** use modernc **and** can never do `DB.Close()` or `Replica.Stop(true)` (unless you've also closed all database connections, or stopping your process).

PR #922, opened 2025-12-19
feat(s3): add distributed leasing using conditional writes

corylanou:
## Summary

Distributed leasing for S3 replicas - Prevents split-brain scenarios when multiple Litestream instances target the same S3 destination.

### Why This Feature?

When running Litestream in production environments, you often want high availability. Without coordination, multiple Litestream instances targeting the same S3 replica can cause **split-brain scenarios**:

- Two instances simultaneously writing LTX files with conflicting transaction IDs
- Interleaved writes corrupting the replication stream
- Race conditions during compaction and retention enforcement

### Who Needs This?

- **Kubernetes deployments** running Litestream as a sidecar with multiple pod replicas
- **High-availability setups** with primary/standby Litestream instances
- **Auto-scaling environments** where multiple instances may start simultaneously
- **Any production deployment** where you cannot guarantee only one Litestream instance runs at a time

### Key Features

- **Atomic lease acquisition** using S3 conditional writes (`If-None-Match: *`)
- **Epoch-based fencing tokens** that increment on each leader change
- **Server-side timestamps** to prevent clock skew issues
- **Graceful lease release** on shutdown for fast failover
- **Automatic failover** when lease expires (default 30s timeout)

### Phase 1 Scope

Per feedback, this PR focuses on the standalone lease client only, without integrating it into the rest of Litestream. Integration will be done in a follow-up PR after this is merged.

---

## Files Changed

**New files:**
- `leaser.go` - Generic `Leaser` interface and `Lease` struct
- `leaser_test.go` - Unit tests for Leaser interface
- `s3/leaser.go` - S3 leaser implementation
- `s3/leaser_test.go` - S3 leaser unit tests with mock server

## Test Plan

- [x] Unit tests for Leaser interface
- [x] S3 leaser mock tests
- [x] All existing tests pass
- [x] Pre-commit hooks pass

## References

- [PR #617](https://github.com/benbjohnson/litestream/pull/617) - Ben's original leasing implementation
- [Leader Election with S3 Conditional Writes](https://www.morling.dev/blog/leader-election-with-s3-conditional-writes/)

### DIFF
--- a/s3/leaser.go
+++ b/s3/leaser.go
@@ -53,8 +53,9 @@ type Leaser struct {
 	SecretAccessKey string
 
 	// S3 bucket information
-	Region         string
-	Bucket         string
+	Region string
+	Bucket string
+	// Path is the lease prefix in the bucket (global across DBs).
 	Path           string
 	Endpoint       string
 	ForcePathStyle bool
@@ -65,6 +66,7 @@ type Leaser struct {
 func NewLeaser() *Leaser {
 	return &Leaser{
 		LeaseTimeout: litestream.DefaultLeaseTimeout,
+		Owner:        defaultOwner(),
 		logger:       slog.Default().WithGroup("s3-leaser"),
 	}
 }
@@ -84,12 +86,23 @@ func (l *Leaser) Init(ctx context.Context) error {
 	if l.Bucket == "" {
 		return fmt.Errorf("s3 leaser: bucket name is required")
 	}
+	if l.Owner == "" {
+		l.Owner = defaultOwner()
+	}
+	if l.Path == "" {
+		l.Path = "leases"
+	} else {
+		l.Path = strings.Trim(l.Path, "/")
+		if l.Path == "" {
+			l.Path = "leases"
+		}
+	}
 
 	region := l.Region
 	if region == "" {
 		if l.Endpoint == "" {
 			var err error
-			if region, err = findBucketRegion(ctx, l.Bucket); err != nil {
+			if region, err = l.findBucketRegion(ctx, l.Bucket); err != nil {
 				return fmt.Errorf("s3 leaser: cannot lookup bucket region: %w", err)
 			}
 		} else {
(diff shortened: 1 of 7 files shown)


benbjohnson:
@corylanou A couple notes:

Handoff is not zero downtime. If a process dies gracefully, then the lease can be released and a standby could detect that next time it checks for the lock. If a process dies ungracefully, then a standby will need to wait for the expiration of the lock.

We want the lock to be globally scoped. We assume that any collection of SQLite databases managed by a process should be locked together. There could be use cases for for per-database locking but that's more complicated and it's out of scope.

For this first PR, let's just get the lease client working and not try to integrate it into the rest of Litestream. I think getting the integration right will take some iteration.

benbjohnson:
@corylanou Does the fix for #927 need to be in this PR for distributed leasing? Or was that a mistake?

corylanou:
Good catch - the #927 fix was mistakenly included in this PR. I've separated it out:

- Created PR #944 for just the #927 fix
- Removed the #927 commit from this branch (force pushed)
- Updated the PR title and description to reflect only the leasing work

This PR now contains only the Phase 1 standalone lease client as requested.

fuchstim:
Hey @corylanou @benbjohnson, Happy Friday!
I'm currently facing pretty much exactly the situation this PR aims to address; a workload running in K8s which, due to HA requirements, cannot strictly be limited to one active replica (to allow for zero-downtime-deploys). 
My initial plan was to implement the Litestream Go lib and wrap it in a leader election / leasing mechanism. But judging from your PR, work on making this an upstream feature seems to already be well underway. Is there anyway I can help implement / test and ship this feature?

s3/leaser.go
```diff
@@ -0,0 +1,480 @@
...
+		Owner:        defaultOwner(),
+		logger:       slog.Default().WithGroup("s3-leaser"),
+	}
+}
+
+// Type returns "s3" as the leaser type.
+func (l *Leaser) Type() string { return ReplicaClientType }
+
+// Init initializes the connection to S3. No-op if already initialized.
+func (l *Leaser) Init(ctx context.Context) error {
```
fuchstim, line 78: There's a lot of overlap between the s3 client init here and the one in the replica client. But they seem to differ in some cases, not sure if intentionally (e.g. debug logging support, default http transport settings, etc.) Would it make sense to reuse the same client init func between both clients to avoid potential issues due to their differing configuration?

fuchstim:
@corylanou if you're swamped I'll also be happy to take a stab at it.

corylanou:
Closing in favor of #1073, which implements a simpler approach using `If-Match` conditional writes instead of epoch-based fencing tokens. The new implementation uses a single `lock.json` file and doesn't require listing or cleanup of old lock files.

PR #926, opened 2025-12-20
feat(heartbeat): add notification hooks for health check services

### LINKED ISSUE #424: Notification systems integration (hooks)
Sorry if this is already in place, couldn't find anything on it.
As a replication service, one should monitor it (as always).

One possibility is to integrate it with any "ping" service like https://healthchecks.io/ or similar so the service can be ping'ed every time sync (replication) has been done **successfully**.

Something like this https://torsion.org/borgmatic/docs/how-to/monitor-your-backups/#healthchecks-hook

corylanou:
## Summary

Adds heartbeat/ping notification support for integration with health check services like:
- healthchecks.io
- deadmanssnitch
- Cronitor
- Pagerduty

The feature sends periodic HTTP GET requests to configured URLs after successful sync operations.

**Key features:**
- Global defaults with per-database overrides
- Configurable interval (minimum 1 minute, default 5 minutes)
- Rate-limited to avoid overwhelming webhook endpoints
- Only sends pings after successful sync operations

**Configuration example:**
```yaml
# Global defaults (applies to all databases)
heartbeat-url: https://hc-ping.com/xxx-xxx
heartbeat-interval: 5m

dbs:
  - path: /path/to/db1.sqlite
    # Uses global heartbeat settings
    replica:
      url: s3://bucket/db1

  - path: /path/to/db2.sqlite
    # Override with different URL
    heartbeat-url: https://hc-ping.com/yyy-yyy
    replica:
      url: s3://bucket/db2

  - path: /path/to/db3.sqlite
    # Disable heartbeat for this DB
    heartbeat-url: ""
    replica:
      url: s3://bucket/db3
```

## Files Changed

- **heartbeat.go** (new): `HeartbeatClient` with ping logic, rate limiting, interval enforcement
- **heartbeat_test.go** (new): Unit tests for heartbeat functionality
- **db.go**: Added `Heartbeat` field and sync tracking (`RecordSuccessfulSync`, `LastSuccessfulSyncAt`)
- **replica.go**: Records successful sync after replication completes
- **store.go**: Added `monitorHeartbeats` goroutine for background heartbeat monitoring
- **cmd/litestream/main.go**: Config parsing for global and per-db heartbeat settings

## Test plan

- [x] Unit tests for HeartbeatClient (ping, rate limiting, interval enforcement)
- [x] Build succeeds
- [x] All existing tests pass
- [x] Linting passes (go vet, staticcheck, goimports)
- [x] Pre-commit hooks pass
- [ ] Manual testing with healthchecks.io (recommended)

Closes #424

### DIFF
--- a/cmd/litestream/main.go
+++ b/cmd/litestream/main.go
@@ -266,22 +266,6 @@ func (c *Config) propagateGlobalSettings() {
 	}
 }
 
-// propagateHeartbeatSettings copies global heartbeat settings to DBConfigs
-// that don't have their own heartbeat settings.
-func (c *Config) propagateHeartbeatSettings() {
-	for _, dbc := range c.DBs {
-		// If per-db heartbeat URL is not set, use global
-		if dbc.HeartbeatURL == nil && c.HeartbeatURL != "" {
-			url := c.HeartbeatURL
-			dbc.HeartbeatURL = &url
-		}
-		// If per-db heartbeat interval is not set, use global
-		if dbc.HeartbeatInterval == nil && c.HeartbeatInterval != nil {
-			dbc.HeartbeatInterval = c.HeartbeatInterval
-		}
-	}
-}
-
 // DefaultConfig returns a new instance of Config with defaults set.
 func DefaultConfig() Config {
 	defaultSnapshotInterval := 24 * time.Hour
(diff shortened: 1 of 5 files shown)


cmd/litestream/main.go
```diff
@@ -260,6 +266,22 @@ func (c *Config) propagateGlobalSettings() {
...
+			url := c.HeartbeatURL
+			dbc.HeartbeatURL = &url
+		}
+		// If per-db heartbeat interval is not set, use global
+		if dbc.HeartbeatInterval == nil && c.HeartbeatInterval != nil {
+			dbc.HeartbeatInterval = c.HeartbeatInterval
+		}
+	}
+}
+
```
benbjohnson: I don't think databases sharing one heartbeat makes sense. That would mean that if any of the databases are replicating correctly then the heartbeat is valid. But really it should be that if any of the databases are failing then the heartbeat should be failing.

benbjohnson (changes requested):

corylanou:
Thanks for the feedback! I've addressed this by refactoring the heartbeat from per-DB to Store level.

**Key changes:**
- Removed per-DB heartbeat configuration (`DBConfig.HeartbeatURL`, `DBConfig.HeartbeatInterval`)
- Moved `HeartbeatClient` from `DB` struct to `Store` struct
- Heartbeat now uses **all-or-nothing semantics**: a ping is only sent when ALL databases have synced successfully within the heartbeat interval
- Added `Store.allDatabasesHealthy()` helper method
- Added tests verifying:
  - No ping when there are no databases
  - Ping fires when all databases have synced
  - Verified one DB not syncing blocks heartbeat

**New config structure:**
```yaml
# Single global heartbeat - applies to all databases
heartbeat-url: https://hc-ping.com/xxx
heartbeat-interval: 5m

dbs:
  - path: /path/to/db1.sqlite
    # No per-DB heartbeat config
  - path: /path/to/db2.sqlite
    # All DBs must be healthy for heartbeat to fire
```

This ensures that if any database fails to replicate, the heartbeat won't fire - correctly indicating a failure condition.

benbjohnson (approved):

tao:
This is a great feature to have, thanks!

Issue #927, opened 2025-12-20
Runaway disk usage


pettijohn:
Apologies for a bug report with such limited details, but litestream somehow ran away and create 200+GB. Yesterday I migrated off of Nextcloud and `down`ed the container, but forgot to stop the litestream process. Since the sqlite file on disk would sit there unchanged without Nextcloud running, I didn't think it would matter. But this morning I woke up to a full hard drive and it looks like litestream is the culprit. I deleted the `.owncloud.db-litestream` folder without any inspection to restore service. 

Screenshot of `du`. 

<img width="644" height="230" alt="Image" src="https://github.com/user-attachments/assets/e90af50e-5123-4bbc-9aef-c9bac3f35eee" />

Limited logs from `docker logs`, it repeats for 10k lines, and I don't have anything before this (I think the docker logs rolled over, but I'm not sure). 

```
litestream-1  | time=2025-12-19T21:06:26.777Z level=INFO msg="initialized db" path=/shares/nextcloud/data/owncloud.db
litestream-1  | time=2025-12-19T21:06:26.777Z level=INFO msg="replicating to" type=s3 sync-interval=1s bucket=litestream.pettijohn.com path=owncloud.db region="" endpoint=""
litestream-1  | time=2025-12-20T00:34:04.770Z level=ERROR msg="sync error" db=owncloud.db error="checkpoint: cannot copy wal after checkpoint: sync: write ltx from db: encode ltx frame (pgno=16168): write page data: write /shares/nextcloud/data/.owncloud.db-litestream/ltx/0/0000000000001185-0000000000001185.ltx.tmp: no space left on device"
litestream-1  | time=2025-12-20T00:34:05.909Z level=ERROR msg="sync error" db=owncloud.db error="sync: write ltx from db: encode ltx frame (pgno=16164): write page data: write /shares/nextcloud/data/.owncloud.db-litestream/ltx/0/0000000000001185-0000000000001185.ltx.tmp: no space left on device"
litestream-1  | time=2025-12-20T00:34:07.083Z level=ERROR msg="sync error" db=owncloud.db error="sync: write ltx from db: encode ltx frame (pgno=16162): write page data: write /shares/nextcloud/data/.owncloud.db-litestream/ltx/0/0000000000001185-0000000000001185.ltx.tmp: no space left on device"
[...]

PR #931, opened 2025-12-24
fix(sync): add exponential backoff for error resilience

corylanou:
## Summary

Adds exponential backoff when sync errors occur repeatedly, providing defense-in-depth for error resilience.

> **Note**: The root cause of issue #927 (runaway disk usage from excessive snapshots) was fixed in PR #944 via the `syncedToWALEnd` flag. This PR provides complementary error handling for general sync failures.

### What this PR does

- Add `DefaultSyncBackoffMax` (5min) and `SyncErrorLogInterval` (30s) constants
- Add `isDiskFullError()` helper to detect disk space errors (ENOSPC, EDQUOT)
- Modify `DB.monitor()` with exponential backoff on errors
- Modify `Replica.monitor()` with same backoff pattern
- Attempt temp file cleanup after 3 consecutive disk full errors
- Log recovery when sync succeeds after previous errors

### Backoff Behavior

| State | Backoff | Action |
|-------|---------|--------|
| First error | 1s | Log error, wait 1s |
| Second error | 2s | Wait 2s before retry |
| Third error (disk full) | 4s | Clean temp files, wait 4s |
| ... | doubles | Backoff doubles each failure |
| Max reached | 5min | Cap at 5 minutes |
| Success after errors | reset | Log "sync recovered", reset counters |

### Related PRs

- **PR #944** (merged): Fixed the root cause - prevents excessive snapshots after checkpoint truncation via `syncedToWALEnd` flag
- **This PR #931**: Adds general error resilience - exponential backoff on ANY repeated sync errors (network issues, disk full, permissions, etc.)

### Why this is still useful

Even with #944 merged, this PR provides value for:
- Network connectivity issues causing repeated failures
- Actual disk full scenarios (not caused by the snapshot bug)
- Permission errors or other transient failures
- Rate-limited logging prevents log spam during persistent errors

## Test plan

- [x] Added `TestIsDiskFullError` with comprehensive error string tests
- [x] Added `TestIsSQLiteBusyError` for existing helper coverage
- [x] All existing tests pass
- [x] Linters pass (go vet, staticcheck, goimports)

Related to #927 (root cause fixed in #944)

### DIFF
--- a/db.go
+++ b/db.go
@@ -35,6 +35,11 @@ const (
 	DefaultTruncatePageN        = 121359 // ~500MB with 4KB page size
 	DefaultShutdownSyncTimeout  = 30 * time.Second
 	DefaultShutdownSyncInterval = 500 * time.Millisecond
+
+	// Sync error backoff configuration.
+	// When sync errors occur repeatedly (e.g., disk full), backoff doubles each time.
+	DefaultSyncBackoffMax = 5 * time.Minute  // Maximum backoff between retries
+	SyncErrorLogInterval  = 30 * time.Second // Rate-limit repeated error logging
 )
 
 // DB represents a managed instance of a SQLite database in the file system.
@@ -831,6 +836,19 @@ func isSQLiteBusyError(err error) bool {
 	errStr := err.Error()
 	return strings.Contains(errStr, "database is locked") ||
 		strings.Contains(errStr, "SQLITE_BUSY")
+}
+
+// isDiskFullError returns true if the error indicates disk space issues.
+// This includes "no space left on device" (ENOSPC) and "disk quota exceeded" (EDQUOT).
+func isDiskFullError(err error) bool {
+	if err == nil {
+		return false
+	}
+	errStr := strings.ToLower(err.Error())
+	return strings.Contains(errStr, "no space left on device") ||
+		strings.Contains(errStr, "disk quota exceeded") ||
+		strings.Contains(errStr, "enospc") ||
+		strings.Contains(errStr, "edquot")
 }
 
 // walFileSize returns the size of the WAL file in bytes.
@@ -1973,6 +1991,9 @@ func (db *DB) EnforceRetentionByTXID(ctx context.Context, level int, txID ltx.TX
 // When a change is detected, we call Sync() which performs full verification
 // and replication. The cost per tick is just one stat() call plus a 32-byte
 // read, which is negligible.
+//
+// Implements exponential backoff on repeated sync errors to prevent disk churn
+// when persistent errors (like disk full) occur. See issue #927.
 func (db *DB) monitor() {
 	ticker := time.NewTicker(db.MonitorInterval)
 	defer ticker.Stop()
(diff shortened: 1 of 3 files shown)


benbjohnson:
Do we know what is filling the disk? Is it temp files? I would expect temp files to be removed on failure immediately.

corylanou:
@benbjohnson Great question - you're right to be suspicious. **It's NOT temp files.** The temp files are correctly cleaned up via `defer os.Remove()` at `db.go:1266`:

```go
ltxFile, err := os.OpenFile(tmpFilename, os.O_RDWR|os.O_CREATE|os.O_TRUNC, mode)
if err != nil {
    return false, fmt.Errorf("open temp ltx file: %w", err)
}
defer func() { _ = os.Remove(tmpFilename) }()  // ← This works correctly
defer func() { _ = ltxFile.Close() }()
```

## The Real Culprit: Permanent L0 LTX Files

The disk fills with **permanent L0 LTX full snapshots** (not temp files). Users reported seeing patterns like:

```
21:43  5.2G  0000000000000005-0000000000000005.ltx
21:45  411K  0000000000000006-0000000000000006.ltx
21:47  5.2G  0000000000000008-0000000000000008.ltx  ← Full snapshot after checkpoint
21:49  305K  0000000000000009-0000000000000009.ltx
21:50  5.2G  000000000000000b-000000000000000b.ltx  ← Another full snapshot
```

## Root Cause

In `verify()` at `db.go:977-1007`, after a checkpoint truncates the WAL:

```go
// If LTX WAL offset is larger than real WAL then the WAL has been truncated.
if fi, err := os.Stat(db.WALPath()); err != nil {
    return info, fmt.Errorf("open wal file: %w", err)
} else if info.offset > fi.Size() {
    info.reason = "wal truncated by another process"
    return info, nil  // snapshotting = true (the default!)
}
```

**The bug**: After sync completes at position 100KB in WAL, then checkpoint truncates WAL to 32 bytes, the next `verify()` sees `100KB > 32B` and triggers a full snapshot - even though we JUST synced all the data moments before the checkpoint. This happens every checkpoint cycle.

## The Fix (committed to this branch)

Added a `syncedToWALEnd` flag to track when sync reached the exact end of WAL. When verify() sees WAL truncation but this flag is true, it recognizes this as expected checkpoint behavior:

**db.go:76-82** - New flag:
```go
// syncedToWALEnd tracks whether the last successful sync reached the
// exact end of the WAL file. When true, a subsequent WAL truncation
// (from checkpoint) is expected and should NOT trigger a full snapshot.
syncedToWALEnd bool
```

**db.go:1345-1353** - Set flag after sync:
```go
if walSize, err := db.walFileSize(); err == nil {
    finalOffset := info.offset + sz
    db.syncedToWALEnd = finalOffset == walSize
}
```

**db.go:984-1004** - Check flag in verify:
```go
if db.syncedToWALEnd {
    db.syncedToWALEnd = false // Clear flag
    hdr, err := readWALHeader(db.WALPath())
    // ... read new salt values ...
    info.offset = WALHeaderSize
    info.snapshotting = false  // ← Key change: NO snapshot needed!
    return info, nil
}
```

## Test Results

Added `TestDB_CheckpointDoesNotTriggerSnapshot` which confirms:
- Before fix: `snapshotting=true` after checkpoint (bug)
- After fix: `snapshotting=false` after checkpoint (correct)

The exponential backoff in this PR remains useful for handling transient errors, but the root cause of disk-filling is now fixed.

PR #945, opened 2026-01-03
optimization(db): skip unnecessary Sync() calls when WAL unchanged

chrismccord:
The monitor loop was calling Sync() on every tick regardless of whether the WAL had changed. This burned CPU unnecessarily when databases were idle.

This adds change detection by checking WAL file size and the 32-byte header before calling Sync(). If unchanged, we skip the call.

Before: ~0.7% CPU per idle database
After: ~0% CPU per idle database

Added TestDB_Monitor_CheapChangeDetection to verify syncs are skipped when WAL is unchanged.

### DIFF
--- a/db.go
+++ b/db.go
@@ -1918,9 +1918,26 @@ func (db *DB) EnforceRetentionByTXID(ctx context.Context, level int, txID ltx.TX
 }
 
 // monitor runs in a separate goroutine and monitors the database & WAL.
+//
+// Change Detection Strategy:
+// To reduce CPU usage when idle, we use cheap change detection to avoid
+// unnecessary Sync() calls when the WAL hasn't changed:
+//
+//  1. stat() the WAL file to get its size (changes on any write)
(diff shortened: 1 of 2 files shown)


db.go
```diff
@@ -1930,7 +1947,39 @@ func (db *DB) monitor() {
...
+				db.Logger.Error("sync error", "error", err)
+			}
+			continue
+		}
+
+		// Skip sync if WAL size and header are unchanged.
+		walSize := fi.Size()
+		if walSize == lastWALSize && bytes.Equal(walHeader, lastWALHeader) {
+			continue
+		}
```
benbjohnson, line 1977: Checking WAL size doesn't always work as SQLite doesn't necessarily truncate the WAL (depending on checkpointing mode).

corylanou:
Closing - this PR's content was already merged via PR #946, which was a follow-up that included the original changes plus additional improvements. Both PRs shared the same branch (`feature/cheap-wal-change-detection`), so when #946 was merged, all changes from #945 were included.

PR #953, opened 2026-01-05
feat(vfs): add write support with periodic sync to remote

benbjohnson:
## Summary

- Add write support to the VFS, allowing it to be the sole writer to a SQLite database with changes periodically synced to a remote replica
- Implement dirty page tracking with read-your-writes consistency
- Add conflict detection (fails if remote has newer TXIDs than expected)
- Add write buffer file for crash recovery/durability
- Add transaction tracking via SQLite lock states to batch writes
- Handle journal files so SQLite can write in rollback journal mode
- Add environment variable configuration for write mode in litestream-vfs

## Configuration

| Variable | Description |
|----------|-------------|
| `LITESTREAM_WRITE_ENABLED` | Set to `true` to enable writes |
| `LITESTREAM_SYNC_INTERVAL` | Sync interval (default: `1s`) |
| `LITESTREAM_LOCAL_PATH` | Directory for write buffer file (crash recovery) |

### 1. Build the VFS extension

```bash
make vfs
```

### 2. Create initial database and replica

```bash
# Create a test database
sqlite3 /tmp/test.db "CREATE TABLE users(id INTEGER PRIMARY KEY, name TEXT); INSERT INTO users VALUES(1, 'Alice');"

# Build litestream

# Replicate to local file backend (run for a few seconds, then Ctrl+C)
./bin/litestream replicate /tmp/test.db file:///tmp/replica
```

### 3. Test writable VFS with file backend

```bash
export LITESTREAM_REPLICA_URL="file:///tmp/replica"
export LITESTREAM_WRITE_ENABLED="true"
export LITESTREAM_SYNC_INTERVAL="1s"
export LITESTREAM_LOCAL_PATH="/tmp"
export LITESTREAM_LOG_LEVEL="DEBUG"

sqlite3 <<EOF
.load /path/to/litestream/dist/litestream-vfs
SELECT litestream_register();
.open file:test.db?vfs=litestream
SELECT * FROM users;
INSERT INTO users VALUES(2, 'Bob');
SELECT * FROM users;
EOF
```

### 4. Test with S3 backend (optional)

```bash
# First replicate to S3
./bin/litestream replicate /tmp/test.db s3://your-bucket/replica

# Then test writable VFS
export LITESTREAM_REPLICA_URL="s3://your-bucket/replica"
export LITESTREAM_WRITE_ENABLED="true"
export LITESTREAM_SYNC_INTERVAL="1s"
export LITESTREAM_LOCAL_PATH="/tmp"

sqlite3 <<EOF
.load /path/to/litestream/dist/litestream-vfs
SELECT litestream_register();
.open file:test.db?vfs=litestream
INSERT INTO users VALUES(3, 'Charlie');
SELECT * FROM users;
EOF
```

## Test plan

- [x] Unit tests for write functionality (9 new tests in `vfs_write_test.go`)
- [x] All existing VFS tests pass
- [x] Manual testing with file backend

### DIFF
--- a/vfs.go
+++ b/vfs.go
@@ -69,9 +69,9 @@ type VFS struct {
 	// Default: 1 second. Set to 0 for manual sync only.
 	WriteSyncInterval time.Duration
 
-	// WriteLocalPath is the path for local write buffer persistence.
-	// If empty, uses temp directory.
-	WriteLocalPath string
+	// WriteBufferPath is the path for local write buffer persistence.
+	// If empty, uses a temp file.
+	WriteBufferPath string
 
 	tempDirOnce sync.Once
 	tempDir     string
@@ -117,8 +117,15 @@ func (vfs *VFS) openMainDB(name string, flags sqlite3vfs.OpenFlag) (sqlite3vfs.F
 		}
 
 		// Set up write buffer path
-		if vfs.WriteLocalPath != "" {
-			f.bufferPath = filepath.Join(vfs.WriteLocalPath, ".litestream-write-buffer")
+		if vfs.WriteBufferPath != "" {
+			f.bufferPath = vfs.WriteBufferPath
+		} else {
+			// Use a temp file if no path specified
+			dir, err := vfs.ensureTempDir()
+			if err != nil {
+				return nil, 0, fmt.Errorf("create temp dir for write buffer: %w", err)
+			}
+			f.bufferPath = filepath.Join(dir, "write-buffer")
 		}
 	}
 
@@ -1044,14 +1051,8 @@ func (f *VFSFile) syncToRemote() error {
 	f.pendingTXID++
 	f.pos = ltx.Pos{TXID: f.expectedTXID}
 
-	// Update index with synced pages and update cache
+	// Update cache with synced pages (index will be populated naturally when pages are fetched)
 	for pgno, bufferOff := range f.dirty {
-		f.index[pgno] = ltx.PageIndexElem{
-			Level:   0,
-			MinTXID: info.MinTXID,
-			MaxTXID: info.MaxTXID,
-		}
-		// Read data from buffer and copy to cache
 		cachedData := make([]byte, f.pageSize)
 		if _, err := f.bufferFile.ReadAt(cachedData, bufferOff); err != nil {
 			return fmt.Errorf("read page %d from buffer for cache: %w", pgno, err)
--- a/cmd/litestream-vfs/main.go
+++ b/cmd/litestream-vfs/main.go
@@ -76,8 +76,8 @@ func LitestreamVFSRegister() {
 			vfs.WriteSyncInterval = d
 		}
 
-		if s := os.Getenv("LITESTREAM_LOCAL_PATH"); s != "" {
-			vfs.WriteLocalPath = s
+		if s := os.Getenv("LITESTREAM_BUFFER_PATH"); s != "" {
+			vfs.WriteBufferPath = s
 		}
 	}
 
--- a/cmd/litestream-vfs/vfs_write_integration_test.go
+++ b/cmd/litestream-vfs/vfs_write_integration_test.go
@@ -987,7 +987,7 @@ func newWritableVFS(tb testing.TB, client litestream.ReplicaClient, syncInterval
 	vfs.PollInterval = 100 * time.Millisecond
 	vfs.WriteEnabled = true
 	vfs.WriteSyncInterval = syncInterval
-	vfs.WriteLocalPath = localPath
+	vfs.WriteBufferPath = localPath
 
 	return vfs
 }


vfs.go
```diff
@@ -731,16 +862,514 @@ func (f *VFSFile) ReadAt(p []byte, off int64) (n int, err error) {
...
+
+	// Update commit count if this extends the database
+	if pgno > f.commit {
+		f.commit = pgno
+	}
+
+	// Write to buffer for durability
+	if err := f.writeToBuffer(pgno, page); err != nil {
+		f.logger.Warn("failed to write to buffer", "error", err)
+	}
```
benbjohnson: This should log and return the error to prevent data loss.

vfs.go
```diff
@@ -731,16 +862,514 @@ func (f *VFSFile) ReadAt(p []byte, off int64) (n int, err error) {
...
+	// Check for conflicts
+	if err := f.checkForConflict(ctx); err != nil {
+		return err
+	}
+
+	// Create LTX file from dirty pages
+	ltxData, err := f.createLTXFromDirty()
+	if err != nil {
+		return fmt.Errorf("create LTX: %w", err)
+	}
```
benbjohnson: createLTXFromDirty() should return an io.Reader rather than a set of bytes. A transaction could get very large and consume too much memory if the LTX data is serialized into bytes.

vfs.go
```diff
@@ -731,16 +862,514 @@ func (f *VFSFile) ReadAt(p []byte, off int64) (n int, err error) {
...
+		f.cache.Add(pgno, cachedData)
+	}
+
+	// Clear dirty pages
+	f.dirty = make(map[uint32][]byte)
+
+	// Clear write buffer after successful sync
+	if err := f.clearWriteBuffer(); err != nil {
+		f.logger.Warn("failed to clear write buffer", "error", err)
+	}
```
benbjohnson, line 1070: This should log and return an error.

vfs.go
```diff
@@ -731,16 +862,514 @@ func (f *VFSFile) ReadAt(p []byte, off int64) (n int, err error) {
...
+	countBuf := make([]byte, 4)
+	binary.BigEndian.PutUint32(countBuf, pageCount)
+	if _, err := f.bufferFile.WriteAt(countBuf, 20); err != nil {
+		return fmt.Errorf("update buffer page count: %w", err)
+	}
+
+	// Sync buffer to disk for durability
+	if err := f.bufferFile.Sync(); err != nil {
+		return fmt.Errorf("sync buffer: %w", err)
+	}
```
benbjohnson: Do not sync the write buffer. Durability only exists once changes are sync'd to the replica. The write buffer only exists to move data to disk rather than storing them in memory.

vfs.go
```diff
@@ -516,6 +580,25 @@ func (f *VFSFile) Open() error {
...
+		f.expectedTXID = pos.TXID
+		f.pendingTXID = pos.TXID + 1
+		f.logger.Debug("write support enabled", "expectedTXID", f.expectedTXID, "pendingTXID", f.pendingTXID)
+
+		// Attempt to recover dirty pages from write buffer
+		if recovered, err := f.recoverWriteBuffer(); err != nil {
+			f.logger.Warn("failed to recover write buffer", "error", err)
+		} else if recovered {
+			f.logger.Info("recovered dirty pages from write buffer")
+		}
```
benbjohnson: Discard write buffer on open than trying to recover it. Any unsync'd changes should be lost. You can remove the `recoverWriteBuffer()` function too.

vfs.go
```diff
@@ -407,6 +460,17 @@ type VFSFile struct {
 	pageSize        uint32
 	commit          uint32
 
+	// Write support fields (only used when writeEnabled is true)
+	writeEnabled  bool              // Whether write support is enabled
+	dirty         map[uint32][]byte // Dirty pages: pgno -> page data
```
benbjohnson: This map should map a pgno to an offset in the write buffer for the data of the page rather than storing the page data in memory.

vfs.go
```diff
@@ -731,16 +862,514 @@ func (f *VFSFile) ReadAt(p []byte, off int64) (n int, err error) {
...
+	}
+
+	// Append page entry at end of file
+	entry := make([]byte, 4+len(data))
+	binary.BigEndian.PutUint32(entry[0:4], pgno)
+	copy(entry[4:], data)
+
+	if _, err := f.bufferFile.WriteAt(entry, writeOffset); err != nil {
+		return fmt.Errorf("write page to buffer: %w", err)
+	}
```
benbjohnson: If the page has already been written to the write buffer, overwrite the page. Otherwise, append page data to the end of the file.

vfs.go
```diff
@@ -731,16 +862,514 @@ func (f *VFSFile) ReadAt(p []byte, off int64) (n int, err error) {
...
+
+	writeOffset := stat.Size()
+	if stat.Size() == 0 {
+		// Write header
+		hdr := make([]byte, writeBufferHdrSize)
+		copy(hdr[0:4], writeBufferMagic)
+		binary.BigEndian.PutUint32(hdr[4:8], writeBufferVersion)
+		binary.BigEndian.PutUint64(hdr[8:16], uint64(f.expectedTXID))
+		binary.BigEndian.PutUint32(hdr[16:20], f.pageSize)
+		binary.BigEndian.PutUint32(hdr[20:24], 0) // pageCount placeholder
```
benbjohnson: Remove the write buffer header entirely. It is not needed since we are no longer recovering the data on restart.

vfs.go
benbjohnson: Fixed in 02e2cb2

vfs.go
benbjohnson: Fixed in 6fc64db

vfs.go
benbjohnson: Fixed in c34e79a

vfs.go
benbjohnson: Fixed in 5ab014f

vfs.go
benbjohnson: Fixed in 5ab014f

vfs.go
benbjohnson: Fixed in 5ab014f

vfs.go
benbjohnson: Fixed in 5ab014f

vfs.go
benbjohnson: Fixed in 5ab014f

vfs.go
benbjohnson: No, this should return a streaming io.Reader that doesn't load all the data into memory. Use an io.Pipe() instead.

benbjohnson: Fixed in 495a882

vfs.go
```diff
@@ -90,11 +106,34 @@ func (vfs *VFS) openMainDB(name string, flags sqlite3vfs.OpenFlag) (sqlite3vfs.F
...
+		f.dirty = make(map[uint32]int64)
+		f.syncInterval = vfs.WriteSyncInterval
+		if f.syncInterval == 0 {
+			f.syncInterval = DefaultSyncInterval
+		}
+
+		// Set up write buffer path
+		if vfs.WriteLocalPath != "" {
+			f.bufferPath = filepath.Join(vfs.WriteLocalPath, ".litestream-write-buffer")
+		}
```
benbjohnson, line 129: Rename WriteLocalPath to WriteBufferPath. If WriteBufferPath is not specified, use a temp file instead. There should always be a buffer file since the dirty map points to offsets in it.

vfs.go
```diff
@@ -731,16 +849,411 @@ func (f *VFSFile) ReadAt(p []byte, off int64) (n int, err error) {
...
+			MinTXID: info.MinTXID,
+			MaxTXID: info.MaxTXID,
+		}
+		// Read data from buffer and copy to cache
+		cachedData := make([]byte, f.pageSize)
+		if _, err := f.bufferFile.ReadAt(cachedData, bufferOff); err != nil {
+			return fmt.Errorf("read page %d from buffer for cache: %w", pgno, err)
+		}
+		f.cache.Add(pgno, cachedData)
+	}
```
benbjohnson: Don't add dirty pages to index. Let index get populated naturally when pages get fetched.

vfs.go
benbjohnson: Fixed in 7573ab1

vfs.go
benbjohnson: Fixed in 7573ab1

ncruces:
What happens when page 1 gets written to?

There are other concerns, but the gravest one: what happens if you try to change the page size with `pragma page_size` (possible with rollback - which you're now fooling SQLite to assume - not possible with WAL)?

Also, in rollback mode, _every_ commit, changes page one to update the change counter. You may - or may not - want to ignore that change.

ncruces:
Also, question, do you envision the VFS doing compaction?

Or would it make sense to decouple and for the CLI to have a mode where it can do asynchronous compaction of a replica that's being written to by a VFS?

benbjohnson:
@ncruces Yes, ideally the VFS would do the compaction itself. I can see some benefits to offloading that to a separate CLI but I think the deployment complexity makes it less than ideal.

PR #970, opened 2026-01-08
feat(vfs): add Dockerfile.vfs for VFS-enabled images

corylanou:
## Summary
- Add Dockerfile.vfs for building Docker images with VFS extension
- Properly propagate VFS init errors through SQLite's error mechanism
- Add CI/CD workflow for VFS Docker builds
- Use Debian base image (Alpine has musl libc TLS issues with Go shared libs)

### Error Handling Improvement
The original PR #964 changed `log.Fatalf` to `log.Printf` + `return` to prevent SQLite crashes. However, this silently failed without propagating errors to scripts.

This PR improves on that by:
- Changing `LitestreamVFSRegister()` to return `*C.char` (error string, like other Go exports)
- Updating C code to check return value and use SQLite's `*pzErrMsg` mechanism
- Now `.load` properly fails with error message and exit code when initialization fails

### Dockerfile.vfs
- Multi-stage build with golang:1.24 builder
- Uses Debian instead of Alpine (musl libc has TLS relocation errors with Go shared libs)
- Includes sqlite3 and wrapper script for loading VFS extension
- Supports linux/amd64 and linux/arm64 (no arm/v7 due to CGO complexity)

## Test plan
- [x] Docker image builds successfully
- [x] litestream binary works
- [x] VFS extension is present at /usr/local/lib/litestream-vfs.so
- [x] Error messages properly propagate when LITESTREAM_REPLICA_URL is missing
- [x] VFS tests pass (one flaky stress test fails on main too)

## References
Closes #964 - incorporates the work from @imjasonh with improvements to error handling

---

### DIFF
--- a/.github/workflows/release.docker.yml
+++ b/.github/workflows/release.docker.yml
@@ -15,7 +15,7 @@ jobs:
   docker:
     runs-on: ubuntu-latest
     env:
-      # VFS extension requires CGO which doesn't cross-compile well for arm/v7
+      # CGO cross-compilation supported platforms
       PLATFORMS: "linux/amd64,linux/arm64"
       VERSION: "${{ github.event_name == 'release' && github.event.release.name || github.sha }}"
 


benbjohnson:
@corylanou Is there any reason not to add this directly to the regular `Dockerfile`?

Dockerfile
```diff
@@ -1,16 +1,74 @@
...
+# Standard builder (static, no CGO)
+FROM builder-base AS builder-standard
 RUN --mount=type=cache,target=/root/.cache/go-build \
 	--mount=type=cache,target=/go/pkg \
 	go build -ldflags "-s -w -X 'main.Version=${LITESTREAM_VERSION}' -extldflags '-static'" -tags osusergo,netgo,sqlite_omit_load_extension -o /usr/local/bin/litestream ./cmd/litestream
 
+# VFS builder (CGO enabled, Debian-based due to musl TLS issues)
+# Alpine's musl libc has TLS (Thread Local Storage) compatibility issues with
+# Go-built shared libraries. When loading the VFS extension on Alpine, you get
+# TLS relocation errors. Debian's glibc works without these issues.
```
benbjohnson: If we have to move off of Alpine to Debian get the VFS to work, that's fine, but I think we should remove all these extra build stages. They seem unnecessary.

Dockerfile
```diff
@@ -1,16 +1,74 @@
...
+FROM builder-base AS builder-vfs
+RUN apt-get update && apt-get install -y gcc libc6-dev && rm -rf /var/lib/apt/lists/*
+RUN --mount=type=cache,target=/root/.cache/go-build \
+	--mount=type=cache,target=/go/pkg \
+	CGO_ENABLED=1 go build \
+	-ldflags "-s -w -X 'main.Version=${LITESTREAM_VERSION}'" \
+	-tags "vfs,SQLITE3VFS_LOADABLE_EXT" \
```
benbjohnson: The `litestream` build should be unchanged.

Dockerfile
```diff
@@ -1,16 +1,74 @@
...
+RUN printf '%s\n' \
+	'#!/bin/sh' \
+	'if [ -z "$LITESTREAM_REPLICA_URL" ]; then' \
+	'    echo "Error: LITESTREAM_REPLICA_URL environment variable required"' \
+	'    echo "Example: export LITESTREAM_REPLICA_URL='"'"'s3://bucket/path'"'"'"' \
+	'    exit 1' \
+	'fi' \
+	'exec sqlite3 -cmd ".load /usr/local/lib/litestream-vfs.so sqlite3_litestreamvfs_init" "$@"' \
+	> /usr/local/bin/litestream-vfs-sqlite && \
+	chmod +x /usr/local/bin/litestream-vfs-sqlite
```
benbjohnson: What is `litestream-vfs-sqlite`?

benbjohnson (changes requested):

Dockerfile
imjasonh: These were artifacts from my PR. This is a wrapper to run SQLite with litestream enabled. If that's wrong or misguided by all means remove it.

PR #991, opened 2026-01-12
fix(file): clean up .ltx.tmp files on all error paths

### LINKED ISSUE #988: File backend leaves .ltx.tmp files on error paths
## Description

The file replica client (`file/replica_client.go`) doesn't clean up `.ltx.tmp` files when errors occur during WriteLTXFile. In high-error scenarios (disk full, permission issues), temp files accumulate faster than the reactive cleanup (v0.5.6) can handle.

## Root Cause

Missing defer-based cleanup for temp files on error paths. The current code creates temp files but only removes them on successful rename. If any error occurs (io.Copy fails, Sync fails, etc.), the temp file remains on disk.

## Current Behavior (v0.5.6)

- Reactive cleanup after 3 consecutive disk full errors
- Temp files can accumulate in high-error scenarios
- Each failed write attempt leaves another orphaned temp file

## Expected Behavior

- Temp files should be cleaned up automatically on ALL error paths
- Use defer-based cleanup similar to WebDAV backend
- Success flag pattern to prevent cleanup after successful rename

## Fix

PR incoming with:
- Success flag pattern for temp file cleanup
- Comprehensive test coverage for error scenarios
- Follows established patterns from WebDAV and PATTERNS.md

## Related Issues

- #927 (primary disk runaway bug is separate and fixed in v0.5.6)
- Split from #927 investigation

---

corylanou:
## Summary

Fixes temp file accumulation in file replica client by adding defer-based cleanup for `.ltx.tmp` files on all error paths.

## Problem

The file replica client (`file/replica_client.go`) creates temp files with `.tmp` suffix but only removes them on successful rename. When errors occur (disk full, I/O errors, permission issues), temp files remain on disk.

In high-error scenarios, these orphaned temp files accumulate faster than the reactive cleanup (v0.5.6) can handle, contributing to disk space exhaustion.

## Solution

Implements **success flag pattern** with defer-based cleanup:
- Temp files removed automatically on ALL error paths
- Success flag prevents cleanup after successful rename
- Follows established patterns from WebDAV backend and PATTERNS.md

### Code Changes

**file/replica_client.go (lines 180-234):**
```go
tmpFilename := filename + ".tmp"
f, err := internal.CreateFile(tmpFilename, fileInfo)
if err != nil {
    return nil, err
}

// Track success to determine whether to clean up temp file
var success bool
defer func() {
    _ = f.Close()
    if !success {
        _ = os.Remove(tmpFilename)
    }
}()

// ... operations ...

// Mark success to prevent defer cleanup
success = true
```

**Key characteristics:**
- Double-close is safe (os.File.Close is idempotent)
- No race conditions (success flag set after atomic rename)
- Best-effort cleanup (errors ignored in defer)

## Test Coverage

Added `TestReplicaClient_WriteLTXFile_ErrorCleanup` with 3 scenarios:

1. **DiskFull**: Verifies cleanup after io.Copy failure
2. **SuccessNoLeaks**: Ensures no temp files after successful write
3. **MultipleErrors**: Tests prevention of accumulation across 5 failed writes

All tests **PASS** ✅

## Verification

```bash
go test -v ./file/... -run TestReplicaClient_WriteLTXFile_ErrorCleanup
# PASS
go test -v ./file/...
# PASS
```

## Impact

- **Users experiencing disk runaway**: Reduces temp file accumulation in error scenarios
- **No breaking changes**: Only affects error paths
- **Performance**: Negligible (defer cleanup is very fast)

## Related Issues

- Fixes #988
- Related to #927 (disk runaway investigation)

## Pattern References

- WebDAV backend: `webdav/replica_client.go:257-260`
- PATTERNS.md: Lines 88-109 (atomic file write pattern)
- DB hydration: `db.go:1023` (defer cleanup pattern)

---

### DIFF
--- a/file/replica_client.go
+++ b/file/replica_client.go
@@ -184,12 +184,11 @@ func (c *ReplicaClient) WriteLTXFile(ctx context.Context, level int, minTXID, ma
 		return nil, err
 	}
 
-	// Track success to determine whether to clean up temp file.
-	// On successful rename, the temp file becomes the final file and should not be removed.
-	var success bool
+	// Clean up temp file on error. On successful rename, the temp file
+	// becomes the final file and should not be removed.
 	defer func() {
 		_ = f.Close()
-		if !success {
+		if err != nil {
 			_ = os.Remove(tmpFilename)
 		}
 	}()
@@ -222,9 +221,6 @@ func (c *ReplicaClient) WriteLTXFile(ctx context.Context, level int, minTXID, ma
 	if err := os.Rename(tmpFilename, filename); err != nil {
 		return nil, err
 	}
-
-	// Mark success to prevent defer cleanup
-	success = true
 
 	// Set file ModTime to preserve original timestamp
 	if err := os.Chtimes(filename, timestamp, timestamp); err != nil {


file/replica_client.go
```diff
@@ -178,11 +178,21 @@ func (c *ReplicaClient) WriteLTXFile(ctx context.Context, level int, minTXID, ma
...
+	tmpFilename := filename + ".tmp"
+	f, err := internal.CreateFile(tmpFilename, fileInfo)
 	if err != nil {
 		return nil, err
 	}
-	defer f.Close()
+
+	// Track success to determine whether to clean up temp file.
+	// On successful rename, the temp file becomes the final file and should not be removed.
+	var success bool
```
benbjohnson: Does this need a `success` variable? Or can we just check that `err != nil`?

corylanou:
@benbjohnson Good catch! Refactored to use the named return `err` instead of the `success` variable. Much cleaner. Thanks for the feedback!

Changes in commit 0d7f2d9:
- Removed `var success bool` declaration
- Defer now checks `err != nil` directly  
- Removed `success = true` statement
- Updated comment for clarity

All tests still pass ✓

PR #1009, opened 2026-01-15
feat(vfs): add PRAGMA litestream_write_enabled for runtime write control

### LINKED ISSUE #1006: feat(vfs): add PRAGMA litestream_write_enabled for runtime write control
## Summary

Add a new PRAGMA `litestream_write_enabled` to enable and disable write support at runtime. Currently, write support can only be enabled at startup via the `LITESTREAM_WRITE_ENABLED` environment variable.

## Motivation

Runtime write control allows applications to:
- Dynamically switch between read-only and read-write modes
- Gracefully disable writes before maintenance or failover
- Enable writes only when needed, reducing risk of accidental modifications

### Reading the current value
```sql
PRAGMA litestream_write_enabled;
-- Returns: "0" (disabled) or "1" (enabled)
```

### Disabling writes
```sql
PRAGMA litestream_write_enabled = 0;
```
- If called during active transaction, waits for transaction to complete
- Syncs all dirty pages to replica before returning
- Blocks until sync completes successfully
- Returns error if sync fails (writes remain enabled)

### Enabling writes
```sql
PRAGMA litestream_write_enabled = 1;
```
- Initializes write buffer file if not already present
- Starts sync loop goroutine if `WriteSyncInterval > 0`
- Works even if VFS was not started with `LITESTREAM_WRITE_ENABLED=true` (cold enable)

### Key files
- `vfs.go` - Add case to `FileControl()` switch (~line 1917)
- `vfs.go` - Add `SetWriteEnabled(bool) error` method to `VFSFile`

### Disable flow
1. Acquire mutex
2. If in active transaction, wait for transaction to complete
3. Call `syncToRemote()` if dirty pages exist
4. Set `f.writeEnabled = false`
5. Release mutex

### Enable flow (cold enable supported)
1. Acquire mutex
2. Initialize buffer file if nil (create temp file)
3. Set `f.writeEnabled = true`
4. Start sync ticker if not running and interval > 0
5. Release mutex

### Error cases
- Disable sync failure: return error, keep writes enabled

## Testing

- Test read returns correct value ("0" or "1")
- Test disable syncs dirty pages before returning
- Test disable during transaction waits for commit
- Test enable after disable works
- Test cold enable (without `LITESTREAM_WRITE_ENABLED` env var)
- Test concurrent read/write operations during state change
- Test disable sync failure keeps writes enabled

corylanou:
## Summary

- Add new PRAGMA `litestream_write_enabled` to enable/disable write support at runtime
- Support "cold enable" - enabling writes on a VFS that was not started with `LITESTREAM_WRITE_ENABLED=true`
- Wait for active transactions to complete before disabling writes
- Sync dirty pages to replica before completing disable

## Motivation

Runtime write control allows applications to:
- Dynamically switch between read-only and read-write modes
- Gracefully disable writes before maintenance or failover
- Enable writes only when needed, reducing risk of accidental modifications

## Usage

```sql
-- Reading current value
PRAGMA litestream_write_enabled;
-- Returns: "0" (disabled) or "1" (enabled)

-- Disabling writes
PRAGMA litestream_write_enabled = 0;
-- Waits for transaction completion, syncs dirty pages

-- Enabling writes (including cold enable)
PRAGMA litestream_write_enabled = 1;
-- Initializes buffer file, starts sync loop
```

## Test plan

- [x] Test read returns correct value ("0" or "1")
- [x] Test disable syncs dirty pages before returning
- [x] Test disable during transaction waits for commit
- [x] Test enable after disable works
- [x] Test cold enable (without `LITESTREAM_WRITE_ENABLED` env var)
- [x] Test no-op when already in requested state
- [x] Test invalid value returns error
- [x] All tests pass with race detection

Fixes #1006

### DIFF
--- a/vfs.go
+++ b/vfs.go
@@ -185,14 +185,19 @@ func (vfs *VFS) openMainDB(name string, flags sqlite3vfs.OpenFlag) (sqlite3vfs.F
 		return nil, 0, err
 	}
 
-	// Always report ReadWrite to SQLite so that cold enable via PRAGMA
-	// litestream_write_enabled works. SQLite permanently marks databases as
-	// read-only based on the output flags from xOpen (pager.c:readOnly,
-	// btree.c:BTS_READ_ONLY), which would prevent write transactions even
-	// after enabling writes at runtime. Read-only enforcement happens at
-	// the VFS layer (WriteAt, Truncate, Lock) when writeEnabled is false.
-	flags &^= sqlite3vfs.OpenReadOnly
-	flags |= sqlite3vfs.OpenReadWrite
+	// When SQLite requests read-write access, always report ReadWrite in the
+	// output flags so that cold enable via PRAGMA litestream_write_enabled
+	// works. SQLite permanently marks databases as read-only based on the
+	// output flags from xOpen (pager.c:readOnly, btree.c:BTS_READ_ONLY),
+	// which would prevent write transactions even after enabling writes at
+	// runtime. Read-only enforcement happens at the VFS layer (WriteAt,
+	// Truncate, Lock) when writeEnabled is false.
+	//
+	// If the caller explicitly requested read-only, we respect that intent.
+	if flags&sqlite3vfs.OpenReadOnly == 0 {
+		flags &^= sqlite3vfs.OpenReadOnly
+		flags |= sqlite3vfs.OpenReadWrite
+	}
 
 	return f, flags, nil
 }
@@ -1412,7 +1417,7 @@ func (f *VFSFile) WriteAt(b []byte, off int64) (n int, err error) {
 
 	// If write support is not enabled, return read-only error
 	if !f.writeEnabled {
-		return 0, fmt.Errorf("litestream is a read only vfs")
+		return 0, sqlite3vfs.ReadOnlyError
 	}
 
 	// Get page data - either from buffer file (if dirty) or from cache/remote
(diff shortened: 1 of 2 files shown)


vfs.go
```diff
@@ -1989,6 +2130,32 @@ func (f *VFSFile) FileControl(op int, pragmaName string, pragmaValue *string) (*
 		result := f.hydrationPath
 		return &result, nil
 
+	case "litestream_write_enabled":
+		if pragmaValue == nil {
+			// READ mode - return current state
+			if f.writeEnabled {
```
benbjohnson: `VFSFile.writeEnabled` needs a mutex around it here (and a couple other spots).

vfs.go
```diff
@@ -1524,15 +1533,147 @@ func (f *VFSFile) Sync(flag sqlite3vfs.SyncType) error {
...
+			f.mu.Lock()
+		}
+
+		// Sync dirty pages if any exist
+		if len(f.dirty) > 0 {
+			f.mu.Unlock()
+			if err := f.syncToRemote(); err != nil {
+				return fmt.Errorf("sync before disable: %w", err)
+			}
+			f.mu.Lock()
```
benbjohnson: Claude's approach to unlocking so the calling function can re-lock is a little insane. We should have wrapper functions for some of these so there's a `syncToRemote()` and a `syncToRemoteWithLock()`

vfs.go
```diff
@@ -1524,15 +1533,147 @@ func (f *VFSFile) Sync(flag sqlite3vfs.SyncType) error {
...
+		}
+	}
+
+	// Initialize buffer file if not present
+	if f.bufferFile == nil {
+		f.mu.Unlock()
+		if err := f.initWriteBuffer(); err != nil {
+			return fmt.Errorf("init write buffer: %w", err)
+		}
+		f.mu.Lock()
```
benbjohnson: This is another weird re-locking situation. We should have two separate versions of `initWriteBuffer()` so we don't have to do this dance.

benbjohnson (changes requested):

benbjohnson (approved):

PR #1008, opened 2026-01-15
Add IPC control commands for dynamic start/stop of database replication

corylanou:
Closes #1002

## Summary

Adds runtime control commands for dynamically managing database replication without restarting the daemon. This enables use cases like:

- **Multi-tenant deployments**: Selectively enable/disable replication per database
- **Failover with external locks**: Replication state doesn't persist across restarts by default
- **Dynamic provisioning**: Start replicating new databases without daemon restart

### Control Server
- Unix socket listener (default: `/var/run/litestream.sock`, permissions: `0600`)
- HTTP API via shared metrics server (`/control/` endpoint)
- JSON-RPC 2.0 protocol for consistent request/response format

### CLI Commands

| Command | Description |
|---------|-------------|
| `litestream start <db>` | Start replication for a database |
| `litestream stop <db>` | Stop replication (always waits for clean shutdown) |
| `litestream sync <db>` | Force immediate sync |
| `litestream status <db>` | Query replication status |
| `litestream list` | List all managed databases |
| `litestream info` | Show daemon information (version, pid, uptime) |

### Command Options
- `--config` - Config file with replica settings (for `start`)
- `--wait` - Block until operation completes
- `--timeout` - Maximum wait time in seconds (default: 30)
- `--socket` - Custom socket path

### Configuration

```yaml
# Control socket (enabled by default)
socket: /var/run/litestream.sock
socket-permissions: 0600

# Optional: persist start/stop state to config file
persist-to-config: false

# Per-database auto-replicate control
dbs:
  - path: /data/app.db
    auto-replicate: false  # Requires manual litestream start
    replicas:
      - url: s3://bucket/path
```

### Store API
New methods for programmatic control:
- `RegisterDB(db)` - Register a database for management
- `FindDB(path)` - Find database by path
- `EnableDB(path)` - Start replication
- `DisableDB(path)` - Stop replication

## Key Files

| File | Purpose |
|------|---------|
| `cmd/litestream/control.go` | Control server implementation |
| `cmd/litestream/config_writer.go` | Atomic config file persistence |
| `cmd/litestream/start.go` | Start command with config loading |
| `cmd/litestream/stop.go` | Stop command (always waits) |
| `cmd/litestream/list.go` | List databases command |
| `cmd/litestream/info.go` | Daemon info command |
| `store.go` | Store control methods |
| `docs/IPC_CONTROL.md` | Comprehensive documentation |

## Testing

```bash
# Unit tests
go test -v ./cmd/litestream/... -run Control
[...]

PR #1010, opened 2026-01-16
feat(replicate): add IPC control commands for dynamic start/stop

corylanou:
## Summary

- Add Unix socket-based control server for runtime database management
- New start and stop CLI commands to control replication dynamically
- HTTP-over-Unix-socket protocol with simple JSON requests (per PR #1008 feedback)
- Store methods: EnableDB, DisableDB, FindDB, RegisterDB
- DB.IsOpen() method to track database open state

## Changes from Original PR #1008

Based on Ben's review feedback:
- Renamed ControlServer to Server
- Uses http.Serve() over Unix socket instead of raw JSON
- Simplified to JSON-over-HTTP (removed JSON-RPC 2.0 envelope)

## Configuration

```yaml
socket:
  path: /var/run/litestream.sock
  permissions: 0600
```

## Usage

```bash
# Start replication for a database
litestream start -socket /var/run/litestream.sock /path/to/db

# Stop replication for a database  
litestream stop -socket /var/run/litestream.sock /path/to/db
```

## Test plan

- [ ] Build succeeds
- [ ] Unit tests pass
- [ ] Manual test: Start replicate with socket config, use start/stop commands
- [ ] Verify socket permissions are applied correctly

Generated with Claude Code

### DIFF
--- a/store.go
+++ b/store.go
@@ -278,6 +278,8 @@ func (s *Store) RemoveDB(ctx context.Context, path string) error {
 }
 
 // EnableDB starts replication for a registered database.
+// The context is checked for cancellation before opening.
+// Note: db.Open() itself does not support cancellation.
 func (s *Store) EnableDB(ctx context.Context, path string) error {
 	db := s.FindDB(path)
 	if db == nil {
@@ -286,6 +288,11 @@ func (s *Store) EnableDB(ctx context.Context, path string) error {
 
 	if db.IsOpen() {
 		return fmt.Errorf("database already enabled: %s", path)
+	}
+
+	// Check for cancellation before starting open
+	if err := ctx.Err(); err != nil {
+		return fmt.Errorf("enable database: %w", err)
 	}
 
 	if err := db.Open(); err != nil {
--- a/db.go
+++ b/db.go
@@ -470,16 +470,19 @@ func (db *DB) Close(ctx context.Context) (err error) {
 		if e := db.db.Close(); e != nil && err == nil {
 			err = e
 		}
+		db.db = nil
 	}
 
 	if db.f != nil {
 		if e := db.f.Close(); e != nil && err == nil {
 			err = e
 		}
+		db.f = nil
 	}
 
 	db.mu.Lock()
 	db.opened = false
+	db.rtx = nil
 	db.mu.Unlock()
 
 	return err
--- a/etc/litestream.yml
+++ b/etc/litestream.yml
@@ -2,6 +2,12 @@
 # access-key-id:     AKIAxxxxxxxxxxxxxxxx
 # secret-access-key: xxxxxxxxxxxxxxxxxxxxxxxxxxxxxx/xxxxxxxxx
 
+# Control socket for runtime start/stop commands
+# socket:
+#   enabled: true                          # Enable control socket (default: false)
+#   path: /var/run/litestream.sock         # Socket path (default: /var/run/litestream.sock)
+#   permissions: 0600                      # Socket file permissions (default: 0600)
+
 # dbs:
 #  - path: /path/to/primary/db            # Database to replicate from
 #    replica:


cmd/litestream/control.go
```diff
@@ -0,0 +1,194 @@
+package main
```
benbjohnson: Can you move this to the root package? You'll probably need to move the `Config` too.

cmd/litestream/control.go
```diff
@@ -0,0 +1,194 @@
...
+
+	ctx    context.Context
+	cancel context.CancelFunc
+	wg     sync.WaitGroup
+
+	logger *slog.Logger
+}
+
+// NewServer creates a new Server instance.
+func NewServer(store *litestream.Store, socketPath string, socketPerms uint32) *Server {
```
benbjohnson: This function signature is going to get long if we eventually keep adding args. Can you have it just accept `store` and then make `socketPath` and `socketPerms` exported so you can set those from the returned `*Server` instance?

cmd/litestream/control.go
```diff
@@ -0,0 +1,194 @@
...
+	}
+	s.wg.Wait()
+	return nil
+}
+
+func (s *Server) handleStart(w http.ResponseWriter, r *http.Request) {
+	if r.Method != http.MethodPost {
+		http.Error(w, "method not allowed", http.StatusMethodNotAllowed)
+		return
+	}
```
benbjohnson: `http.ServeMux` should be able to do this since Go 1.22:

```go
mux.HandleFunc("POST /start", s.handleStart)
```

store.go
```diff
@@ -277,6 +277,72 @@ func (s *Store) RemoveDB(ctx context.Context, path string) error {
...
+	// Check for duplicate path
+	for _, existing := range s.dbs {
+		if existing.Path() == db.Path() {
+			return fmt.Errorf("database already registered: %s", db.Path())
+		}
+	}
+
+	s.dbs = append(s.dbs, db)
+	return nil
+}
```
benbjohnson: This looks like it's unused at the moment. Are you thinking you'd add separate `register` & `unregister` commands that are separate from `start` and `stop`?

cmd/litestream/replicate.go
```diff
@@ -41,6 +41,9 @@ type ReplicateCommand struct {
 	// MCP server
 	MCP *MCPServer
 
+	// Control server for IPC commands.
+	Control *Server
```
benbjohnson: Just name it `Server` :)

cmd/litestream/start.go
```diff
@@ -0,0 +1,101 @@
...
+	timeout := fs.Int("timeout", 30, "timeout in seconds")
+	socketPath := fs.String("socket", "", "control socket path (required)")
+	fs.Usage = c.Usage
+	if err := fs.Parse(args); err != nil {
+		return err
+	}
+
+	if *socketPath == "" {
+		return fmt.Errorf("socket path required; use -socket flag")
+	}
```
benbjohnson: We can have a default on the socket path for the client of `/var/run/litestream.sock`.

Actually, it might be better to have the server socket config have a default path as well but we add an `enabled` flag. e.g.

```yml
socket:
  enabled: false
  path: /var/run/litestream.sock
```

cmd/litestream/start.go
```diff
@@ -0,0 +1,101 @@
...
+		}
+		return fmt.Errorf("start failed: %s", string(body))
+	}
+
+	var result StartResponse
+	if err := json.Unmarshal(body, &result); err != nil {
+		return fmt.Errorf("failed to parse response: %w", err)
+	}
+
+	output, _ := json.MarshalIndent(result, "", "  ")
```
benbjohnson: Why is the error being dropped?

benbjohnson (approved):

Issue #976, opened 2026-01-08
Large quantities of LTX files with v.0.5.X


alongwill:
## Bug Description

I'm using litestream in a Kubernetes Pod to replicate a sqlite db to Cloudflare R2. This is using an image built from [b946989](https://github.com/benbjohnson/litestream/commit/b94698955305a31d4d8f09832758bd712ad9f52a). After for running for 25 hrs we are seeing the following file sizes:

| File  | Size |
| ------------- | ------------- |
| sqlite db `omni.db`  | 8.9M |
| WAL `omni.db-wal` | 8.4M |
| Dir size `.omni.db-litestream/` | 930.0M ⚠️ |
| Num files in `.omni.db-litestream/` | 369 ⚠️ |
| Cloudflare R2: Num LTX files | 3710 ⚠️ |
| Cloudflare R2: Total size of all LTX files | 8986.25 MB ⚠️ |

## Environment

**Litestream version:**

```text
# custom build
v0.5.6-b946989
```

**Operating system & version:**

Talos Linux 1.10.7, Kubernetes 1.33.5
```
uname -r
6.12.43-talos
```

**Installation method:**

Docker image running in Kubernetes

**Storage backend:**

Cloudflare R2

## Steps to Reproduce

1. Run any sqlite backed DB with litestream v0.5.X and Cloudflare R2. We've had similar results with uptime-kuma.

**Expected behavior:**

The shadow WAL and LTX files should remain a similar size to the actual sqlite DB?

**Actual behavior:**

The shadow WAL is almost 1GB
The LTX files in Cloudflare R2 total 8986.25 MB

## Configuration

<details><summary>litestream.yml</summary>

```yaml
dbs:
  - path: /data/omni.db

    replicas:
      - type: s3
        bucket: ****
        path: omni
        endpoint: https://****.r2.cloudflarestorage.com
        force-path-style: true
        sync-interval: 1m
```

</details>

## Logs

<details><summary>Log output</summary>

```text
# Paste relevant logs here
```

</details>

## Additional Context

Add any other context about the problem here (e.g., recent changes, related issues, workarounds you've tried).

corylanou:
Thanks for the detailed report\! I've investigated this issue and found that all the relevant retention fixes (d4dfb33, 8415802, f236f50) are included in your build (b946989), so this appears to be a new issue we need to diagnose further.

[...]

PR #1011, opened 2026-01-16
feat(metrics): add metrics for local and replica LTX files

alongwill:
## Description

Add prometheus metrics to monitor LTX file sizes locally and in replicas:

| Metric                         | Labels                  | Description |
| -------------------------------|-------------------------|------------- |
| `litestream_local_ltx_files`   | db                      | Total LTX files in local metadata |
| `litestream_local_ltx_bytes`   | db                      | Total size of local LTX files |
| `litestream_replica_ltx_files` | db, replica_type, level | LTX files by compaction level |
| `litestream_replica_ltx_bytes` | db, replica_type, level | LTX bytes by compaction level |

Assisted by Claude Opus 4.5.

<details>
<summary>Example prometheus metrics</summary>

```txt
# HELP litestream_local_ltx_bytes Total size of LTX files in local metadata directory
# TYPE litestream_local_ltx_bytes gauge
litestream_local_ltx_bytes{db="/data/fruits.db"} 1991
# HELP litestream_local_ltx_files Total number of LTX files in local metadata directory
# TYPE litestream_local_ltx_files gauge
litestream_local_ltx_files{db="/data/fruits.db"} 4
# HELP litestream_replica_ltx_bytes Size of LTX files by compaction level (replica)
# TYPE litestream_replica_ltx_bytes gauge
litestream_replica_ltx_bytes{db="/data/fruits.db",level="0",replica_type="s3"} 1991
litestream_replica_ltx_bytes{db="/data/fruits.db",level="1",replica_type="s3"} 2290
litestream_replica_ltx_bytes{db="/data/fruits.db",level="2",replica_type="s3"} 1035
litestream_replica_ltx_bytes{db="/data/fruits.db",level="3",replica_type="s3"} 1035
litestream_replica_ltx_bytes{db="/data/fruits.db",level="9",replica_type="s3"} 1470
# HELP litestream_replica_ltx_files Number of LTX files by compaction level (replica)
# TYPE litestream_replica_ltx_files gauge
litestream_replica_ltx_files{db="/data/fruits.db",level="0",replica_type="s3"} 4
litestream_replica_ltx_files{db="/data/fruits.db",level="1",replica_type="s3"} 5
litestream_replica_ltx_files{db="/data/fruits.db",level="2",replica_type="s3"} 2
litestream_replica_ltx_files{db="/data/fruits.db",level="3",replica_type="s3"} 2
litestream_replica_ltx_files{db="/data/fruits.db",level="9",replica_type="s3"} 2
```

</details>

Example Grafana dashboard using these metrics
image

☝️ This example involved making ~2MB of writes to a 13MB sqlite db. Litestream was configured with:
```yaml
    # Compaction levels
    levels:
      - interval: 5m
      - interval: 30m
      - interval: 1h

    # Global snapshot settings
    snapshot:
      interval: 30m
      retention: 1h
```

## Motivation and Context

- Improve observability of LTX files stored locally and in replicas.
  - Locally: so we can size storage according to needs
  - Replica: so we can calculate storage costs
- This information helps understand and optimise settings for compaction levels and snapshots 
  
I implemented this to help debug #976

## How Has This Been Tested?

I ran two instances in a test environment against two accounts with different compaction and snapshot settings. This was using R2 as a replica. Can confirm the metrics were generated and captured successfully and align with reality. 

## Types of changes

## Checklist

### DIFF
--- a/db.go
+++ b/db.go
@@ -333,6 +333,81 @@ func (db *DB) ensureCompactorClient() {
 	}
 }
 
+// LTXStats holds aggregated statistics about LTX files.
+type LTXStats struct {
+	TotalFiles int64
+	TotalBytes int64
+	ByLevel    map[int]LTXLevelStats
+}
+
+// LTXLevelStats holds statistics for a single compaction level.
+type LTXLevelStats struct {
+	Files int64
+	Bytes int64
+}
+
+// LocalLTXStats returns statistics about local LTX files in the metadata directory.
+// Scans the LTX directory and aggregates file counts and sizes.
+func (db *DB) LocalLTXStats() (LTXStats, error) {
+	var stats LTXStats
+
+	ltxDir := db.LTXDir()
+	entries, err := os.ReadDir(ltxDir)
+	if os.IsNotExist(err) {
+		return stats, nil
+	} else if err != nil {
+		return stats, fmt.Errorf("read ltx dir: %w", err)
+	}
+
+	for _, entry := range entries {
+		if !entry.IsDir() {
+			continue
+		}
+
+		// Skip non-numeric directories
+		if _, err := strconv.Atoi(entry.Name()); err != nil {
+			continue
+		}
+
+		subdir := filepath.Join(ltxDir, entry.Name())
+		files, err := os.ReadDir(subdir)
+		if err != nil {
+			continue
+		}
+
+		for _, file := range files {
+			if file.IsDir() {
+				continue
+			}
+			if _, _, err := ltx.ParseFilename(file.Name()); err != nil {
+				continue // skip non-LTX files
+			}
+
+			info, err := file.Info()
+			if err != nil {
+				continue
+			}
+
+			stats.TotalFiles++
+			stats.TotalBytes += info.Size()
+		}
+	}
+
+	return stats, nil
+}
+
+// updateLocalLTXMetrics updates Prometheus metrics for local LTX files.
+func (db *DB) updateLocalLTXMetrics() {
+	stats, err := db.LocalLTXStats()
+	if err != nil {
+		db.Logger.Debug("failed to collect local LTX stats", "error", err)
+		return
+	}
+
+	localLTXFilesGaugeVec.WithLabelValues(db.path).Set(float64(stats.TotalFiles))
+	localLTXBytesGaugeVec.WithLabelValues(db.path).Set(float64(stats.TotalBytes))
+}
+
 // MaxLTX returns the last LTX file written to level 0.
 func (db *DB) MaxLTX() (minTXID, maxTXID ltx.TXID, err error) {
 	ents, err := os.ReadDir(db.LTXLevelDir(0))
@@ -853,6 +928,9 @@ func (db *DB) Sync(ctx context.Context) (err error) {
 		db.dbSizeGauge.Set(float64(fi.Size()))
 	}
 	db.walSizeGauge.Set(float64(newWALSize))
+
+	// Update local LTX file metrics.
+	db.updateLocalLTXMetrics()
 
 	// Notify replicas of WAL changes.
 	// if changed {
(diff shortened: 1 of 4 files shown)


replica_test.go
```diff
@@ -1662,5 +1662,58 @@ func verifyRestoredDB(t *testing.T, path string) {
 	}
 	if result != "ok" {
 		t.Fatalf("integrity check returned: %s", result)
+// TestReplica_LTXStats verifies that LTXStats returns correct file counts and sizes from replica.
```
corylanou, line 1665: **Bug:** This new test function is inserted *inside* `verifyRestoredDB` — the closing `}` for the `if result != "ok"` block and the function itself (lines 1666-1667 in the base) get pushed to the end of the new test. This won't compile.

The new `TestReplica_LTXStats` function needs to be placed *after* the closing `}` of `verifyRestoredDB`.

db.go
```diff
@@ -854,6 +929,9 @@ func (db *DB) Sync(ctx context.Context) (err error) {
 	}
 	db.walSizeGauge.Set(float64(newWALSize))
 
+	// Update local LTX file metrics.
```
corylanou, line 932: **Performance:** `updateLocalLTXMetrics()` does a full `os.ReadDir` + `file.Info()` walk of every LTX file across all level directories. Since `Sync()` fires on every WAL change, this could add noticeable I/O overhead for databases with many LTX files.

Consider either:
- Running this on a timer (e.g., every 30s) rather than every sync
- Implementing the Prometheus `Collector` interface so stats are computed only when the `/metrics` endpoint is scraped

replica.go
```diff
@@ -170,6 +171,9 @@ func (r *Replica) Sync(ctx context.Context) (err error) {
 	// Record successful sync for heartbeat monitoring.
 	r.db.RecordSuccessfulSync()
 
+	// Update replica LTX metrics after successful sync.
+	r.UpdateLTXMetrics(ctx)
```
corylanou, line 175: **Performance:** `UpdateLTXMetrics` iterates levels 0 through 9, calling `r.Client.LTXFiles()` for each. For remote backends (S3, R2, etc.), that's up to **10 LIST API calls per sync per replica**. This adds both latency to the sync path and cost.

Same suggestion as the local side — decouple from the sync hot path with a timer or Prometheus `Collector`.

replica.go
```diff
@@ -221,6 +225,61 @@ func (r *Replica) MaxLTXFileInfo(ctx context.Context, level int) (info ltx.FileI
...
+// Iterates through all compaction levels (0 through SnapshotLevel) and aggregates
+// file counts and sizes using the existing LTXFiles iterator.
+func (r *Replica) LTXStats(ctx context.Context) (LTXStats, error) {
+	stats := LTXStats{ByLevel: make(map[int]LTXLevelStats)}
+
+	// Iterate through all possible levels including snapshot level
+	for level := 0; level <= SnapshotLevel; level++ {
+		itr, err := r.Client.LTXFiles(ctx, level, 0, false)
+		if err != nil {
+			// Skip levels that don't exist or have errors
```
corylanou, line 238: Silently continuing on `LTXFiles` errors could mask real issues (permissions, network problems, etc.). Consider logging at debug level here, consistent with the `itr.Close()` error handling a few lines below:
```go
if err != nil {
    r.Logger().Debug("error listing LTX files", "level", level, "error", err)
    continue
}
```

replica.go
```diff
@@ -221,6 +225,61 @@ func (r *Replica) MaxLTXFileInfo(ctx context.Context, level int) (info ltx.FileI
...
+	}
+
+	stats, err := r.LTXStats(ctx)
+	if err != nil {
+		r.Logger().Debug("failed to collect replica LTX stats", "error", err)
+		return
+	}
+
+	dbPath := r.db.Path()
+	replicaType := r.Client.Type()
```
corylanou, line 274: **Stale metrics:** This only sets gauges for levels that currently have files. If a level previously had files but is now empty (e.g., after compaction), the old non-zero gauge value will persist in Prometheus until process restart.

Consider resetting the gauge vecs for this db/replica before setting new values, or explicitly iterating all levels and setting empty ones to 0.

db.go
```diff
@@ -333,6 +333,81 @@ func (db *DB) ensureCompactorClient() {
 	}
 }
 
+// LTXStats holds aggregated statistics about LTX files.
+type LTXStats struct {
+	TotalFiles int64
```
corylanou, line 338: Nit: `LTXStats` has `TotalFiles`/`TotalBytes` and `ByLevel`, but `LocalLTXStats()` only populates the totals and `Replica.LTXStats()` only populates `ByLevel`. The shared struct is a bit misleading — consider either populating all fields in both cases, or using separate types for local vs. replica stats.

corylanou (changes requested):
Hey @alongwill — thanks for this PR and sorry it took so long to get to! The feature is well-motivated and the metric design is clean. Great work on the detailed description and Grafana example.

I found a few issues ranging from a compile-breaking bug in the test to some performance concerns. See the inline comments for details. The main themes are:

1. **Syntax error** in `replica_test.go` that will prevent compilation
2. **Performance** — collecting stats on every sync (especially the replica side with remote LIST calls) could be expensive. Consider decoupling from the sync hot path
3. **Stale metrics** — emptied compaction levels will retain old non-zero values

Looking forward to the next iteration!

PR #1012, opened 2026-01-16
fix(vfs): resolve ARM64 segfault when loading extension

benbjohnson:
## Summary

- Fixes segmentation fault when loading VFS extension on linux/arm64
- Root cause: duplicate `sqlite3_api` variables in separate C translation units, only one initialized
- Uses forked sqlite3vfs with `extern` declaration to share the pointer
- Adds CI jobs to test extension loading and prevent regression

## Problem

When loading `litestream.so` into sqlite3 CLI on linux/arm64:

```
sqlite> .load ./litestream.so
Segmentation fault
```

**Root cause:** Two separate `sqlite3_api` global variables were created:
1. `litestream-vfs.c` - was using `extern` (expecting sqlite3vfs to define it)
2. `sqlite3vfs.c` (in Go archive) - defined with `SQLITE_EXTENSION_INIT1`

Only one got initialized via `SQLITE_EXTENSION_INIT2(pApi)`, causing NULL pointer dereference.

## Solution

Flip the relationship:
- `litestream-vfs.c` now **defines** `sqlite3_api` via `SQLITE_EXTENSION_INIT1`
- Forked `sqlite3vfs` uses `extern` to **reference** it

## Changes

| File | Change |
|------|--------|
| `src/litestream-vfs.c` | Use `SQLITE_EXTENSION_INIT1` instead of `extern` |
| `go.mod` | Point to forked sqlite3vfs |
| `cmd/litestream-vfs/load_test.go` | New integration tests for extension loading |
| `.github/workflows/commit.yml` | Add load tests to macOS, Linux, and ARM64 CI jobs |

## Test plan

- [x] Extension loads without segfault on darwin-arm64
- [x] New `TestVFS_Load` tests pass locally
- [ ] CI passes on macOS
- [ ] CI passes on Linux amd64
- [ ] CI passes on Linux arm64 (new job)

### DIFF
--- a/.github/workflows/commit.yml
+++ b/.github/workflows/commit.yml
@@ -28,7 +28,7 @@ jobs:
       - uses: pre-commit/action@v3.0.0
 
   vfs-build-test:
-    name: VFS Build Test (macOS)
+    name: VFS Build & Load Test (macOS)
     runs-on: macos-latest
     steps:
       - uses: actions/checkout@v4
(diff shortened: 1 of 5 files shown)


benbjohnson:
This was a red herring, closing. Linux amd64 required the init function passed into `.load` whereas MacOS doesn't allow it:

```
$ sqlite3
sqlite> .load dist/litestream-vfs.so sqlite3_litestream_init
```

Issue #1016, opened 2026-01-19
Unable to restore from Azure on different machine


rphes:
## Bug Description

I have an Ubuntu production system running Litestream to back up our database to Azure. I want to restore the backed up database to my Macbook for development. When I try to restore I get

```
> litestream restore -o database.sqlite abs://storage-account@container/path
time=2026-01-19T12:24:46.684+01:00 level=ERROR msg="failed to run" error="no matching backup files available"
```

Restoring the database on the production system does work:

```
> litestream restore -o test_restore.sqlite /home/user/database.sqlite
```

## Environment

Production system: `0.5.6`
Macbook: `0.5.6`

**Operating system & version:**

Production system: Ubuntu 20.04.6 LTS
Macbook: macOS 15.7.3 (Sequoia)

**Installation method:**

Production system: .deb install
Macbook: binary download

**Storage backend:**
Azure

## Steps to Reproduce

**Ubuntu system:**

`/etc/litestream.yml`:

```yml

account-name: storage-account
account-key: 

dbs:
  - path: /home/user/database.sqlite
    replica:
      url: abs://storage-account@container/path
```

Run Litestream as systemd service.

**Macbook:**

No config file. Logged in to Azure using the Azure CLI. `az login` etcetera.

Run

```
> litestream restore -o database.sqlite abs://storage-account@container/path
```

**Expected behavior:**
A file `database.sqlite` should be created on the Macbook with the latest replica.

**Actual behavior:**
An error is returned saying "no matching backup files available"

## Configuration
See repro.

## Logs
No relevant logs. I'm unable to obtain any output on the Macbook other than the listed error message.

NB: Appreciate you building this! Let me know if I can provide more info.

corylanou:
Thanks for reporting this! We just merged a fix (PR #1017) that should help diagnose the underlying issue. Previously, authentication and permission errors from Azure blob listing were being swallowed, resulting in the unhelpful "no matching backup files available" message. With this fix, the actual error should now be surfaced.

Could you build from `main` and try again? Here's how:

```bash
# Clone the repo
git clone https://github.com/benbjohnson/litestream.git
cd litestream

# Build the binary
go build -o litestream ./cmd/litestream

# Use the local binary
./litestream restore -o database.sqlite abs://storage-account@container/path
```

This requires Go 1.23+ to be installed. If you don't have Go, you can install it from https://go.dev/dl/

[...]

PR #1017, opened 2026-01-19
fix(iterator): surface blob listing errors for better diagnostics

corylanou:
## Summary

Modified iterator `Close()` methods to return `itr.err` instead of `nil`. This ensures errors from blob listing (auth failures, permission errors) are properly surfaced through the standard `Close()` call pattern.

Previously, errors from Azure/S3/GCS blob iteration were silently ignored because `Close()` returned `nil` even when `Err()` had an error. This caused misleading error messages like "no matching backup files available" when the actual issue was authentication failure.

## Important Note

**This is not a complete fix for #1016** - it improves diagnostic information by surfacing the underlying errors. The root cause (Azure CLI authentication not working across machines) still needs investigation. However, with this change, users will now see the actual error:

```
error="cannot calc restore plan: abs: cannot list blobs: ... AuthorizationFailure"
```

Instead of the unhelpful:
```
error="no matching backup files available"
```

## Changes

| Backend | Change |
|---------|--------|
| `abs/replica_client.go` | `Close()` returns `itr.err` |
| `s3/replica_client.go` | `Close()` returns `itr.err` |
| `oss/replica_client.go` | `Close()` returns `itr.err` |
| `nats/replica_client.go` | `Close()` returns `itr.err` |

Note: GCS (`gs/replica_client.go`) already had the correct behavior.

## Test plan

- [x] Verified with Azurite (Azure Blob Storage emulator)
- [x] Backup with correct credentials works
- [x] Restore with correct credentials works  
- [x] Restore with wrong credentials now shows `AuthorizationFailure` instead of "no matching backup files"
- [x] `go test ./...` passes
- [x] `pre-commit run --all-files` passes

Related to #1016

### DIFF
--- a/abs/replica_client.go
+++ b/abs/replica_client.go
@@ -394,7 +394,7 @@ func newLTXFileIterator(ctx context.Context, client *ReplicaClient, level int, s
 func (itr *ltxFileIterator) Close() (err error) {
 	itr.closed = true
 	itr.cancel()
-	return nil
+	return itr.err
 }
 
 func (itr *ltxFileIterator) Next() bool {
--- a/nats/replica_client.go
+++ b/nats/replica_client.go
@@ -503,7 +503,7 @@ func (itr *ltxFileIterator) Err() error {
 	return itr.err
 }
 
-// Close closes the iterator.
+// Close closes the iterator and returns any error that occurred during iteration.
 func (itr *ltxFileIterator) Close() error {
-	return nil
+	return itr.err
 }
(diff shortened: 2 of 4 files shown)


cmd/litestream/ltx.go
```diff
@@ -88,6 +88,11 @@ func (c *LTXCommand) Run(ctx context.Context, args []string) (err error) {
 			info.CreatedAt.Format(time.RFC3339),
 		)
 	}
+
+	if err := itr.Err(); err != nil {
+		return err
+	}
```
benbjohnson: It's better to just return `itr.Err()` from `itr.Close()`. That way we don't forget to return `itr.Err()`. I thought `itr.Close()` would already be doing that actually.

benbjohnson (approved):

PR #1020, opened 2026-01-19
fix(vfs): resolve write buffer race condition causing EOF errors

### LINKED ISSUE #1018: VFS write buffer race condition causes 'read page from buffer: EOF' errors
## Description

When using the Litestream VFS with write support enabled and a short sync interval (1 second), periodic sync errors occur:

```
ERROR periodic sync failed component=litestream-vfs vfs=true name=/var/lib/juicefs/meta.db error="upload LTX: read page 1 from buffer: EOF"
```

## Steps to Reproduce

1. Create a VFS with `WriteEnabled = true` and `WriteSyncInterval = 1 * time.Second`
2. Open a SQLite database through the VFS
3. Perform writes to the database
4. Observe periodic sync errors in the logs

## Expected Behavior

The sync loop should successfully read dirty pages from the write buffer and sync them to the remote replica.

## Actual Behavior

The sync loop attempts to read pages from the buffer file at offsets stored in the `dirty` map, but the buffer file returns EOF, indicating it's empty or shorter than expected.

## Analysis

Looking at the code in `vfs.go`, the issue appears to be a race condition:

1. `WriteAt()` calls `writeToBuffer()` which writes page data to the buffer file and updates `f.dirty[pgno]` with the buffer offset
2. The sync loop in `syncLoop()` calls `syncToRemote()` which calls `createLTXFromDirty()`
3. `createLTXFromDirty()` reads from `bufferFile.ReadAt(data, bufferOff)` using offsets from `f.dirty`

The race appears to be between:
- Writing to the buffer file and updating `f.dirty`
- The sync loop reading from the buffer file

Possible causes:
1. The buffer file write hasn't been flushed to disk yet
2. The `f.dirty` map is updated before the buffer write completes
3. File handle buffering issues

## Workaround

Using a longer sync interval (10 seconds) avoids the issue, suggesting it's timing-related.

## Environment

- Litestream version: v0.5.6
- Go version: 1.24
- OS: Linux (Docker container)
- Architecture: arm64

## Context

Discovered while integrating Litestream VFS with JuiceFS for on-demand SQLite metadata fetching.

benbjohnson:
## Summary
- Pre-read dirty page data from buffer file into memory before spawning the goroutine in `createLTXFromDirty()`
- Prevents race condition where buffer could be truncated while goroutine is reading
- Added test case to verify the fix

## Problem
When using the VFS with write support and a short sync interval (1 second), periodic sync errors occurred:
```
ERROR periodic sync failed error="upload LTX: read page 1 from buffer: EOF"
```

The race condition:
1. `syncToRemote()` calls `createLTXFromDirty()` which spawns a goroutine that reads from the buffer file
2. If `WriteLTXFile` returns early (error/timeout), the lock is released
3. Another sync can complete and truncate the buffer via `clearWriteBuffer()`
4. The original goroutine tries to read from the now-empty buffer → EOF error

## Solution
Changed `createLTXFromDirty()` to pre-read all dirty page data into memory **while the lock is held**, before spawning the goroutine. The goroutine now reads from the in-memory map instead of the buffer file.

## Test plan
- [x] Added `TestVFSFile_CreateLTXFromDirty_BufferClearedAfterCreation` that simulates the race condition
- [x] All existing VFS write tests pass
- [x] Build succeeds with `-tags vfs`

Fixes #1018

### DIFF
--- a/vfs.go
+++ b/vfs.go
@@ -1557,11 +1557,22 @@ func (f *VFSFile) syncToRemote() error {
 		return err
 	}
 
-	// Create LTX file from dirty pages
-	ltxReader := f.createLTXFromDirty()
+	// Create LTX file from dirty pages. The done channel signals when the
+	// goroutine completes - we must wait for it before releasing the lock
+	// to prevent another sync from clearing the buffer while reading.
+	ltxReader, done := f.createLTXFromDirty()
 
 	// Upload LTX file to remote
 	info, err := f.client.WriteLTXFile(ctx, 0, f.pendingTXID, f.pendingTXID, ltxReader)
+
+	// If WriteLTXFile returned early (error or success), the goroutine may still
+	// be running. Close the pipe reader to unblock it, then wait for completion.
+	// This is safe because WriteLTXFile has already finished reading.
+	if rc, ok := ltxReader.(io.Closer); ok {
+		rc.Close()
+	}
+	<-done // Wait for goroutine to finish before releasing lock
+
 	if err != nil {
 		return fmt.Errorf("upload LTX: %w", err)
 	}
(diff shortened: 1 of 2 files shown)


vfs.go
```diff
@@ -1640,30 +1643,42 @@ func (f *VFSFile) checkForConflict(ctx context.Context) error {
...
+	// another sync completing while our goroutine is still reading.
+	pageData := make(map[uint32][]byte, len(pgnos))
+	for _, pgno := range pgnos {
+		bufferOff := f.dirty[pgno]
+		data := make([]byte, f.pageSize)
+		if _, err := f.bufferFile.ReadAt(data, bufferOff); err != nil {
+			return nil, fmt.Errorf("read page %d from buffer: %w", pgno, err)
+		}
+		pageData[pgno] = data
 	}
```
benbjohnson: We should not read the entire buffer into memory. We have the memory buffer so that it doesn't use too much memory. We should instead prevent another sync from clearing out the buffer until this sync is finished.

vfs.go
```diff
@@ -1557,11 +1557,22 @@ func (f *VFSFile) syncToRemote() error {
 		return err
 	}
 
-	// Create LTX file from dirty pages
-	ltxReader := f.createLTXFromDirty()
+	// Create LTX file from dirty pages. The done channel signals when the
+	// goroutine completes - we must wait for it before releasing the lock
+	// to prevent another sync from clearing the buffer while reading.
+	ltxReader, done := f.createLTXFromDirty()
```
benbjohnson, line 1563: We don't need a `done` channel because `WriteLTXFile()` below won't finish until the reader is closed in the goroutine.

benbjohnson:
Closing this because Claude is going down the wrong rabbit hole.

Issue #1019, opened 2026-01-19
VFS expectedTXID not updated after successful sync causes false conflict errors


benbjohnson:
## Description

After a successful sync to the remote replica, the VFS's `expectedTXID` is not being updated, causing subsequent syncs to fail with a false conflict detection error.

## Log Output

```
2026-01-19 17:48:07.238234 I | INFO synced to remote component=litestream-vfs vfs=true name=/var/lib/juicefs/meta.db txid=0000000000000003 pages=3 size=2254
2026-01-19 17:48:07.242710 I | WARN conflict detected component=litestream-vfs vfs=true name=/var/lib/juicefs/meta.db expected=0000000000000002 remote=0000000000000003
2026-01-19 17:48:07.242772 I | ERROR periodic sync failed component=litestream-vfs vfs=true name=/var/lib/juicefs/meta.db error="remote has newer transactions than expected: expected TXID 2 but remote has 3"
```

## Analysis

The sequence shows:
1. Sync succeeds: `txid=0000000000000003` is written to remote
2. Immediately after, conflict is detected: `expected=0000000000000002 remote=0000000000000003`
3. The sync fails because remote (TXID 3) is newer than expected (TXID 2)

The issue is that after successfully syncing TXID 3, the VFS's `expectedTXID` should be updated to 3, but it remains at 2. This causes the next sync attempt to detect a false conflict.

## Expected Behavior

After a successful sync of TXID N, `expectedTXID` should be updated to N so subsequent syncs don't report conflicts.

## Actual Behavior

`expectedTXID` remains at the old value, causing all subsequent syncs to fail with conflict errors until the state is somehow corrected.

## Impact

This error is non-fatal - the data is successfully synced. However, the error messages are confusing and subsequent sync attempts fail unnecessarily.

## Environment

- Litestream version: v0.5.6
- Go version: 1.24
- OS: Linux (Docker container)
- Architecture: arm64

## Related

This may be related to #1018 (write buffer race condition) as both involve sync timing issues.

PR #1022, opened 2026-01-19
fix(vfs): share expectedTXID state across multiple VFSFile instances

benbjohnson:
## Summary
- Fixes issue #1019: After a successful sync, other VFSFile instances for the same database now see the updated expectedTXID
- Adds shared `vfsWriteState` struct with mutex-protected state per database path
- Implements reference-counted state lifecycle management via `acquireWriteState`/`releaseWriteState`

## Problem
Multiple VFSFile instances for the same database had independent `expectedTXID` and `pendingTXID` fields. When one connection synced and updated its state, other connections still had the old value, causing false conflict detection errors like:
```
remote has newer transactions than expected: expected TXID 2 but remote has 3
```

## Solution
Share `expectedTXID` and `pendingTXID` state across all VFSFile instances for the same database path using a reference-counted shared state structure managed at the VFS level.

## Test plan
- [x] New test `TestVFSFile_SharedWriteStateAcrossInstances` verifies multiple VFSFile instances share state correctly
- [x] All existing VFS write tests pass
- [x] Build passes with no errors

### DIFF
--- a/vfs.go
+++ b/vfs.go
@@ -108,6 +108,22 @@ type VFS struct {
 	tempDirErr  error
 	tempFiles   sync.Map // canonical name -> absolute path
 	tempNames   sync.Map // canonical name -> struct{}{}
+
+	// writeStates tracks shared write state for each database path.
+	// Multiple VFSFile instances for the same database share state to prevent
+	// false conflict detection when multiple connections are open.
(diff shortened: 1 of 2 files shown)


benbjohnson:
This is the wrong approach, Claude.

PR #1030, opened 2026-01-21
feat(logging): add process ID to log output

benbjohnson:
## Summary
- Adds `pid` attribute to all log messages for process identification
- Helps correlate logs when multiple litestream processes are running

## Test plan
- [ ] Build and run litestream, verify pid appears in log output
- [ ] Test with both text and JSON log formats


benbjohnson:
Closing as this was not the underlying reason for the bug. It was an issue with connection pooling spawning multiple VFSes.

PR #1055, opened 2026-01-27
fix(dir-watch): prevent thundering herd at startup with many databases

### LINKED ISSUE #1051: Directory Watcher w/ ~400 DBs crashes litestream
## Bug Description

When running litestream 0.5.3 in the directory wacthing mode on 383 DBs, litestream becomes unstable rapidly (CPU exceeding 100%; Memory exceeding 1.5G; both of these limits are what Litestream reached before I killed it to avoid it taking down the server, I have no reason to think litestream would otherwise stop there).

This is in the context of migrating from the 0.3.x version of litestream. Having each of these databases individually watched (and editing the configuration file and doing a reload) litestream's CPU and Memory usage would be negligible/acceptable (i.e. small enough not to have measured them).

This is also in the context of "bootstrapping Litestream", i.e. I have not pointed the new setup to the same location as the old one, and am (on purpose) starting from fresh.

If I take a small  (22 DBs) sample of the available DBs by pointing my path at `/path/to/tenants/a*.db` (note the `a`) litestream is able to get going. However, this is hardly a general solution and would seem to defeat the purpose (at least partially) of the directory watching setup.

## Environment

**Litestream version:**

```text
0.5.3
```

```
Ubuntu 24.04
```

**Installation method:**
```
apt / dpkg
```

**Storage backend:**
S3 (Hetzner)

## Steps to Reproduce

1. Put lots of DBs in the directory to watch (for some value of lots)
2. Configure directory mode
3. Watch it burn

**Expected behavior:**
I would expect this to either "magically work", or, if this is not an option, have a documented mechanism (with known trade-offs) to "ease in" to an existing multi-tenant setup with many tenants.

## Configuration

```yaml
snapshot:
  interval: 24h
  retention: 672h
dbs:
  - dir: /path/to/tenants
    pattern: "*.db.sqlite3"
    watch: true
    monitor-interval: 10s
    replica:
      url: s3://bucket-name/tenants
      endpoint: https://fsn1.your-objectstorage.com
      sign-payload: true
      require-content-md5: false
      access-key-id: ...
      secret-access-key: ...
```

</details>

corylanou:
## Summary

Fixes #1051

Fixes a "thundering herd" problem when Litestream starts with many pre-existing databases. When starting with 500+ databases, all monitor goroutines would fire their first sync within a 1-second window, causing CPU spikes of 900%+ and potential resource exhaustion.

## Root Cause

When databases are loaded at startup:
1. All `db.Open()` calls start a monitor goroutine
2. All monitors use the same `MonitorInterval` (1 second default)
3. After 1 second, **all monitors fire simultaneously**
4. Each calls `db.Sync()` → `db.init()` which is CPU-intensive (opens SQLite, creates tables, acquires locks)

### Core Fixes

1. **Jittered monitor starts** (`db.go`)
   - Added random delay (0 to 3x MonitorInterval) before first tick
   - Spreads initial syncs over 3 seconds instead of 1 second
   - Most impactful fix for the thundering herd

2. **Rate-limited store opening** (`store.go`)
   - Added `openDatabasesConcurrently()` with semaphore
   - Limits to 10 concurrent `db.Open()` calls during startup

3. **Increased debounce interval** (`directory_watcher.go`)
   - Changed from 100ms to 250ms
   - Reduces event processing overhead at scale

4. **Reduced concurrent init limit** (`directory_watcher.go`)
   - Changed from 10 to 3 concurrent database initializations
   - Lowers peak CPU during mass database discovery

### Test Infrastructure

- **`cmd/stress-test/`** - Standalone stress test tool with:
  - Resource monitoring (CPU/memory via `ps`)
  - Auto-kill on threshold exceeded
  - Pre-created and dynamic scaling test modes
  - Real-time metrics display

- **`cmd/litestream/directory_watcher_stress_test.go`** - Integration tests

### Before Fix (500 databases)
| Metric | Value | Status |
|--------|-------|--------|
| Peak CPU | 969% | ✗ Failed |
| Sustained CPU (10s avg) | 237% | ✗ Failed |
| Detection time | ~5s | ✓ |
| Memory | 241MB | ✓ |

### After Fix (500 databases)
| Metric | Value | Status |
|--------|-------|--------|
| Peak CPU | ~150% | ✓ Improved |
| Sustained CPU (10s avg) | <100% | ✓ Passes |
| Detection time | ~6s | ✓ |
| Memory | ~300MB | ✓ |

## How to Test

```bash
# Build
go build -o bin/stress-test ./cmd/stress-test

# Pre-created 500 test (main scenario)
./bin/stress-test --test=precreated --dbs=500 --duration=60s

# Dynamic scaling test
./bin/stress-test --test=dynamic --initial-dbs=20 --final-dbs=500 --batch-size=20

# Integration tests
go test -v -run "TestDirectoryWatcher_Stress" ./cmd/litestream
```

## Test plan

- [x] Pre-created 500 databases stress test passes
- [x] Dynamic scaling stress test runs
- [x] Integration tests pass
- [x] Existing unit tests pass
- [ ] CI checks pass

### DIFF
--- a/store.go
+++ b/store.go
@@ -10,6 +10,7 @@ import (
 	"time"
 
 	"github.com/superfly/ltx"
+	"golang.org/x/sync/errgroup"
 )
 
 var (
@@ -153,19 +154,21 @@ func (s *Store) Open(ctx context.Context) error {
 		return err
 	}
 
-	// Stagger db.Open() calls to prevent thundering herd when many databases
-	// start their monitor goroutines simultaneously.
-	var openDelay time.Duration
-	if n := len(s.dbs); n > 1 {
-		openDelay = DefaultMonitorInterval / time.Duration(n)
+	g, ctx := errgroup.WithContext(ctx)
+	g.SetLimit(50)
+	for _, db := range s.dbs {
+		db := db
+		g.Go(func() error {
+			select {
+			case <-ctx.Done():
+				return ctx.Err()
+			default:
+				return db.Open()
+			}
+		})
 	}
-	for i, db := range s.dbs {
-		if err := db.Open(); err != nil {
-			return err
-		}
-		if openDelay > 0 && i < len(s.dbs)-1 {
-			time.Sleep(openDelay)
-		}
+	if err := g.Wait(); err != nil {
+		return err
 	}
 
 	// Start monitors for compactions & snapshots.


corylanou:
## Stress Test Results

Ran the full stress test suite to validate this fix against the conditions reported in #1051 (~400 databases causing instability). Each scenario was tested at **100, 250, 500, 1000, and 2500** databases.

### Pre-Created Databases (all exist before monitor starts)
| Count | Result | Time |
|-------|--------|------|
| 100   | PASS   | <1s  |
| 250   | PASS   | <1s  |
| 500   | PASS   | ~1s  |
| 1000  | PASS   | ~2s  |
| 2500  | PASS   | ~5s  |

### Dynamic Scaling (databases added in batches while monitor runs)
| Final Count | Batch Size | Result | Time |
|-------------|-----------|--------|------|
| 100         | 10        | PASS   | ~3s  |
| 250         | 25        | PASS   | ~8s  |
| 500         | 50        | PASS   | ~18s |
| 1000        | 100       | PASS   | ~40s |
| 2500        | 250       | PASS   | ~109s|

### Concurrent Writes (sustained 5 writes/sec/db for 10s)
| DB Count | Total Writes | Result | Time |
|----------|-------------|--------|------|
| 100      | 4,928       | PASS   | ~12s |
| 250      | 12,304      | PASS   | ~25s |
| 500      | 19,385      | PASS   | ~45s |
| 1000     | 26,598      | PASS   | ~80s |
| 2500     | 17,112      | PASS   | ~196s|

All 15 subtests pass. The fix handles 6.5x the database count reported in the original issue without instability.

**Run locally with:** `go test -v -tags stress ./cmd/litestream/ -run TestDirectoryWatcher`

db.go
```diff
@@ -2041,6 +2044,17 @@ func (db *DB) EnforceRetentionByTXID(ctx context.Context, level int, txID ltx.TX
...
+	// Jitter the first tick to prevent thundering herd when many databases start simultaneously.
+	if db.MonitorInterval > 0 && !db.monitorDelayDisabled {
+		jitterWindow := 3 * db.MonitorInterval
+		jitter := time.Duration(rand.Int63n(int64(jitterWindow)))
+		select {
+		case <-db.ctx.Done():
+			return
+		case <-time.After(jitter):
+		}
+	}
```
benbjohnson: I think it'd be better to rate limit the `db.Open()` calls in `Store.Open()` instead. With this, we'll always have a random 0-3 second delay even if there's only one database (which is typically the case).

cmd/litestream/directory_watcher.go
```diff
@@ -111,7 +123,11 @@ func (dm *DirectoryMonitor) run() {
 			if !ok {
 				return
 			}
-			dm.handleEvent(event)
+			if dm.handleEvent(event) {
+				debounceTimer.Reset(debounceInterval)
+			}
```
benbjohnson: This seems like this would debounce indefinitely as long as there are new writes coming in. I would expect that the `debounceTimer` would fire `debounceInterval` after the first event occurred.

db.go
corylanou: Good call. Done in b60aa49 — removed the per-DB jitter from `monitor()` entirely and moved staggering into `Store.Open()`:

```go
var openDelay time.Duration
if n := len(s.dbs); n > 1 {
    openDelay = DefaultMonitorInterval / time.Duration(n)
}
for i, db := range s.dbs {
    if err := db.Open(); err != nil {
        return err
    }
    if openDelay > 0 && i < len(s.dbs)-1 {
        time.Sleep(openDelay)
    }
}
```

With 400 DBs at 1s interval each open is ~2.5ms apart, naturally spreading first monitor ticks across the full interval. Single-DB case gets zero delay.

Also removed the `monitorDelayDisabled` test field since it's no longer needed.

cmd/litestream/directory_watcher.go
corylanou: You're right — fixed in b60aa49. The timer now only starts on the first event and won't reset on subsequent events:

```go
if dm.handleEvent(event) && \!dm.debounceActive {
    debounceTimer.Reset(debounceInterval)
    dm.debounceActive = true
}

// In flushPendingEvents:
dm.debounceActive = false
```

Both debounceActive and pendingEvents are only accessed from the run() goroutine, so no mutex needed.

corylanou: Fixed in b60aa49. Added a debounceActive flag — timer only starts on the first event and does not reset on subsequent events. Flag is cleared when flushPendingEvents runs.

corylanou:
Addressed both review comments in b60aa49:

**1. Thundering herd prevention moved to Store.Open()** — Removed per-DB jitter from `monitor()` entirely. `Store.Open()` now staggers `db.Open()` calls with `DefaultMonitorInterval / len(dbs)` spacing. Single-DB case gets zero delay. With 400 DBs at 1s interval, each open is ~2.5ms apart.

**2. Debounce timer fixed** — Added `debounceActive` flag so the timer only starts on the first event and does not reset on subsequent events. Flag is cleared when `flushPendingEvents()` runs. Both fields are only accessed from `run()` goroutine so no mutex needed.

Also fixed a pre-existing integration test that used hardcoded `"snapshot"` directory name instead of `strconv.Itoa(litestream.SnapshotLevel)`.

All tests pass including stress tests up to 2500 databases.

cmd/litestream/directory_watcher.go
corylanou: Fixed in b60aa49. Added a debounceActive flag so the timer only starts on the first event and does not reset on subsequent events. Flag is cleared when flushPendingEvents runs.

corylanou:
@benbjohnson Ready for re-review - rebased onto main (28 commits) with no conflicts. Both review threads were already resolved in the previous commit:

1. **Store.Open() staggering**: Moved thundering herd prevention from per-DB jitter to evenly-spaced delays in `Store.Open()`. Single-DB case gets zero delay.

2. **debounceActive flag**: Timer now fires after fixed interval from first event instead of resetting on each event.

store.go
```diff
@@ -153,10 +153,19 @@ func (s *Store) Open(ctx context.Context) error {
...
+		openDelay = DefaultMonitorInterval / time.Duration(n)
+	}
+	for i, db := range s.dbs {
 		if err := db.Open(); err != nil {
 			return err
 		}
+		if openDelay > 0 && i < len(s.dbs)-1 {
+			time.Sleep(openDelay)
+		}
 	}
```
benbjohnson: I don't think it should be a fixed delay. We should allow a certain number of concurrent database opens at a time (e.g. 50) and once one database finishes opening then another one can open. You can use a semaphore and a goroutine for each worker.

corylanou: Done in 393d04e — replaced fixed delays with `errgroup.SetLimit(50)`:

```go
g, ctx := errgroup.WithContext(ctx)
g.SetLimit(50)
for _, db := range s.dbs {
    db := db
    g.Go(func() error {
        select {
        case <-ctx.Done():
            return ctx.Err()
        default:
            return db.Open()
        }
    })
}
if err := g.Wait(); err != nil {
    return err
}
```

Up to 50 databases open concurrently, new ones start as soon as a slot frees. Remaining opens cancel on first error.

store.go
```diff
@@ -153,10 +154,21 @@ func (s *Store) Open(ctx context.Context) error {
 		return err
 	}
 
+	g, ctx := errgroup.WithContext(ctx)
+	g.SetLimit(50)
```
benbjohnson, line 158: I didn't know that `errgroup` had a built in semaphore. That's cool. 👍

PR #1057, opened 2026-01-27
add replication monitoring endpoint to unix sock

mrkurt:
Add a monitoring endpoint to the unix socket server that streams replication status  when LTX files are uploaded. This enables real-time monitoring of replication progress.

Changes:
- Add OnSync callback to Replica that fires when LTX files are uploaded
- Add StatusMonitor for coordinating event streaming to subscribers
- Add GET /monitor endpoint to server with NDJSON streaming
- Add litestream monitor CLI command to consume the stream
- Add comprehensive tests for all new functionality

Usage:
  litestream monitor -socket /var/run/litestream.sock

### DIFF
--- a/replica.go
+++ b/replica.go
@@ -53,6 +53,11 @@ type Replica struct {
 	// the position file and removing local LTX files, forcing a fresh sync.
 	// Disabled by default to prevent silent data loss scenarios.
 	AutoRecoverEnabled bool
+
+	// OnSync is called after LTX files are uploaded to the replica.
+	// Only fires when actual replication occurs (TXID advances).
+	// May be nil.
+	OnSync func(pos ltx.Pos)
 }
 
 func NewReplica(db *DB) *Replica {
@@ -164,6 +169,11 @@ func (r *Replica) Sync(ctx context.Context) (err error) {
 			return err
 		}
 		r.SetPos(ltx.Pos{TXID: txID})
+
+		// Notify listeners of actual replication
+		if r.OnSync != nil {
+			r.OnSync(r.Pos())
+		}
 	}
 
 	// Record successful sync for heartbeat monitoring.
--- a/replica.go
+++ b/replica.go
@@ -54,10 +54,10 @@ type Replica struct {
 	// Disabled by default to prevent silent data loss scenarios.
 	AutoRecoverEnabled bool
 
-	// OnSync is called after LTX files are uploaded to the replica.
+	// AfterSync is called after LTX files are uploaded to the replica.
 	// Only fires when actual replication occurs (TXID advances).
 	// May be nil.
-	OnSync func(pos ltx.Pos)
+	AfterSync func(pos ltx.Pos)
 }
 
 func NewReplica(db *DB) *Replica {
@@ -171,8 +171,8 @@ func (r *Replica) Sync(ctx context.Context) (err error) {
 		r.SetPos(ltx.Pos{TXID: txID})
 
 		// Notify listeners of actual replication
-		if r.OnSync != nil {
-			r.OnSync(r.Pos())
+		if r.AfterSync != nil {
+			r.AfterSync(r.Pos())
 		}
 	}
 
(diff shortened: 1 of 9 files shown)


replica.go
```diff
@@ -53,6 +53,11 @@ type Replica struct {
 	// the position file and removing local LTX files, forcing a fresh sync.
 	// Disabled by default to prevent silent data loss scenarios.
 	AutoRecoverEnabled bool
+
+	// OnSync is called after LTX files are uploaded to the replica.
+	// Only fires when actual replication occurs (TXID advances).
+	// May be nil.
+	OnSync func(pos ltx.Pos)
```
benbjohnson: Can you rename this to `AfterSync`? I could it being useful to have a `BeforeSync` as well.

server.go
```diff
@@ -211,6 +216,50 @@ func (s *Server) handleStop(w http.ResponseWriter, r *http.Request) {
...
+		Databases: s.StatusMonitor.GetFullStatus(),
+	}
+	if err := json.NewEncoder(w).Encode(event); err != nil {
+		return
+	}
+	flusher.Flush()
+
+	// Subscribe to updates
+	ch := s.StatusMonitor.Subscribe()
+	defer s.StatusMonitor.Unsubscribe(ch)
```
benbjohnson: You should subscribe before you send the full status so you don't lose any incremental statuses in between.

status.go
```diff
@@ -0,0 +1,157 @@
...
+
+	"github.com/superfly/ltx"
+)
+
+// StatusMonitor manages replication status monitoring and event streaming.
+type StatusMonitor struct {
+	store *Store
+
+	mu          sync.RWMutex
+	subscribers map[chan *StatusEvent]struct{}
```
benbjohnson: It'd be a little more flexible to have these be a type like `*StatusSubscriber` and then have a `Ch` field or method that has the channel. Or we could implement that later if we need to expand the functionality.

status.go
```diff
@@ -0,0 +1,157 @@
...
+// Events are dropped for slow subscribers to avoid blocking.
+func (m *StatusMonitor) broadcast(event *StatusEvent) {
+	m.mu.RLock()
+	defer m.mu.RUnlock()
+
+	for ch := range m.subscribers {
+		select {
+		case ch <- event:
+		default:
+			// Drop if subscriber is slow
```
benbjohnson: You should close the subscriber's channel and remove it rather than dropping the event. Otherwise we won't know if we're missing events in a stream under load.

status.go
```diff
@@ -0,0 +1,157 @@
...
+	return status
+}
+
+// StatusEvent represents a single NDJSON event sent to monitoring clients.
+type StatusEvent struct {
+	Type      string           `json:"type"` // "full" or "sync"
+	Timestamp time.Time        `json:"timestamp"`
+	Database  *DatabaseStatus  `json:"database,omitempty"`  // for "sync" events
+	Databases []DatabaseStatus `json:"databases,omitempty"` // for "full" events
+}
```
benbjohnson, line 158: Do you think it makes more sense to have each `StatusEvent` be for a single database and just issue one for each database when we do our initial "full" broadcast?

benbjohnson (changes requested):
Mostly looks good. A couple tweaks needed.

status.go
mrkurt: yeah that makes sense to me.

benbjohnson (approved):

corylanou:
Hey @mrkurt! Thanks so much for putting this together — the replication monitoring endpoint is a great addition and we really appreciate the effort you've put into it.

Just wanted to check in and see if you're still interested in continuing with the requested changes? No pressure at all — if you're busy or have moved on, we'd be happy to pick it up from here and take it across the finish line.

Either way, thanks again for the contribution!

corylanou:
Thanks again for putting this together, @mrkurt — picking it up to get it across the finish line.

Superseded by #1299, which rebases your branch onto current main and re-integrates the `/monitor` endpoint with the post-#1258/#1260/#1261 control-socket dispatch. Both of your commits (`feat: add replication monitoring endpoint with NDJSON streaming` and `pr feedback fixes`) are preserved on the new branch with your authorship intact.

Closing this one in favor of #1299.

PR #1073, opened 2026-01-30
feat(s3): add distributed leasing with If-Match conditional writes

corylanou:
## Summary

- Add simplified `Leaser` interface with `AcquireLease`, `RenewLease`, `ReleaseLease` methods
- Implement S3 leaser using `If-Match`/`If-None-Match` headers for atomic conditional writes
- Use a single `lock.json` file instead of multiple epoch files (simpler, no listing/cleanup needed)

## Background

This is an alternative implementation to PR #922 (which used `If-None-Match: *` with epoch-based fencing tokens). Ben requested a simpler approach using `If-Match` as described in [this article](https://quanttype.net/posts/2025-02-25-leader-election-with-s3-and-if-match.html).

## Key Differences from PR #922

| Aspect | PR #922 | This PR |
|--------|---------|---------|
| Lock files | Multiple epoch files (`0000000001.lock`, etc.) | Single file (`lock.json`) |
| Atomicity | `If-None-Match: *` only | `If-Match: <etag>` for updates, `If-None-Match: *` for creation |
| Cleanup | Background reaping of old files | None needed |
| Listing | Required on every operation | Not needed |
| Complexity | Higher | Lower |

## Algorithm

1. **Read existing lock**: GET `lock.json`, capture ETag and content
2. **Check expiration**: If lock exists and not expired, fail with `LeaseExistsError`
3. **Conditional write**:
   - If no lock existed: PUT with `If-None-Match: *`
   - If lock existed but expired: PUT with `If-Match: <etag>`
4. Success = acquired leadership, failure (412) = retry

## Lock File Format

```json
{"token": 1, "expires_at": 1740151473.206179, "owner": "hostname:pid"}
```

- `token`: Monotonically increasing fencing token
- `expires_at`: Unix timestamp when lease expires
- `owner`: Human-readable identifier for debugging

## Test Plan

- [x] Test initial acquisition with `If-None-Match: *`
- [x] Test renewal with `If-Match: <etag>`
- [x] Test race condition handling (412 response)
- [x] Test expired lease takeover
- [x] Test release and already-deleted scenarios
- [x] Test concurrent acquisition (only one winner)
- [x] Run all existing tests to ensure no regressions

## References

- [Leader Election with S3 and If-Match](https://quanttype.net/posts/2025-02-25-leader-election-with-s3-and-if-match.html)
- [Original blog post (If-None-Match)](https://www.morling.dev/blog/leader-election-with-s3-conditional-writes/)
- [PR #922](https://github.com/benbjohnson/litestream/pull/922) (previous implementation)

### DIFF
--- a/s3/leaser.go
+++ b/s3/leaser.go
@@ -27,8 +27,9 @@ const (
 var (
 	_ litestream.Leaser = (*Leaser)(nil)
 
-	ErrLeaseRequired     = errors.New("lease required")
-	ErrLeaseETagRequired = errors.New("lease etag required")
+	ErrLeaseRequired        = errors.New("lease required")
+	ErrLeaseETagRequired    = errors.New("lease etag required")
+	ErrLeaseAlreadyReleased = errors.New("lease already released")
 )
 
 // S3API is the interface for S3 operations needed by Leaser.
@@ -179,10 +180,7 @@ func (l *Leaser) ReleaseLease(ctx context.Context, lease *litestream.Lease) erro
 	})
 	if err != nil {
 		if isNotExists(err) || isNotFoundError(err) {
-			l.logger.Warn("lease already deleted unexpectedly",
-				"generation", lease.Generation,
-				"owner", lease.Owner)
-			return nil
+			return ErrLeaseAlreadyReleased
 		}
 		if isPreconditionFailed(err) {
 			return litestream.ErrLeaseNotHeld
--- a/s3/leaser_test.go
+++ b/s3/leaser_test.go
@@ -406,8 +406,8 @@ func TestLeaser_ReleaseLease_AlreadyDeleted(t *testing.T) {
 	}
 
 	err := leaser.ReleaseLease(ctx, lease)
-	if err != nil {
-		t.Fatalf("ReleaseLease() should not error for already-deleted lock: %v", err)
+	if !errors.Is(err, ErrLeaseAlreadyReleased) {
+		t.Fatalf("expected ErrLeaseAlreadyReleased, got %v", err)
 	}
 }
 


s3/leaser.go
```diff
@@ -0,0 +1,309 @@
...
+	}
+
+	if lease == nil || lease.ETag == "" {
+		return litestream.ErrLeaseNotHeld
+	}
+
+	key := l.lockKey()
+	_, err := l.client.s3.DeleteObject(ctx, &s3.DeleteObjectInput{
+		Bucket: aws.String(l.client.Bucket),
+		Key:    aws.String(key),
```
fuchstim: Should we be checking the ETag here as well? Otherwise we might end up releasing a Lease we do not own.

leaser.go
```diff
@@ -0,0 +1,44 @@
...
+
+type Leaser interface {
+	Type() string
+	AcquireLease(ctx context.Context) (*Lease, error)
+	RenewLease(ctx context.Context, lease *Lease) (*Lease, error)
+	ReleaseLease(ctx context.Context, lease *Lease) error
+}
+
+type Lease struct {
+	Token     int64     `json:"token"`
```
fuchstim: nit: 'generation' may be a more fitting description here

s3/leaser.go
```diff
@@ -0,0 +1,309 @@
...
+
+	newETag := ""
+	if out.ETag != nil {
+		newETag = *out.ETag
+	}
+
+	return newETag, nil
+}
+
+func timeToUnix(t time.Time) float64 {
```
fuchstim: if you implement the (Un)Marshal methods on litestream.Lease itself you can make it more portable and save yourself the conversions here. time.Time also implements JSON marshalling by default so you may not need any custom marshalling

corylanou:
## Addressing Review Feedback

Thanks for the review @fuchstim! I've addressed your comments in commit b6b4aac:

### 1. Conditional Delete with If-Match (`s3/leaser.go:154`)

> Should we be checking the ETag here as well? Otherwise we might end up releasing a Lease we do not own.

Good catch! Now using conditional delete with `If-Match: lease.ETag`. If the ETag doesn't match (412 Precondition Failed), we return `ErrLeaseNotHeld` to prevent stale holders from deleting another leader's lease.

AWS S3 now supports conditional deletes: https://docs.aws.amazon.com/AmazonS3/latest/userguide/conditional-deletes.html

### 2. Rename 'token' to 'generation' (`leaser.go:32`)

> nit: 'generation' may be a more fitting description here

Good suggestion - I'll update this in a follow-up commit.

### 3. Custom JSON marshalling on Lease (`s3/leaser.go:263`)

> if you implement the (Un)Marshal methods on litestream.Lease itself you can make it more portable and save yourself the conversions here

The reason for the `lockFile` struct with float64 `expires_at` is to match the format in the reference article (Unix timestamp as float). However, I can look into using `time.Time` directly with custom JSON marshalling if preferred. What do you think?

s3/leaser.go
```diff
@@ -0,0 +1,310 @@
...
+	logger *slog.Logger
+
+	TTL   time.Duration
+	Owner string
+}
+
+func NewLeaser(client *ReplicaClient) *Leaser {
+	owner, _ := os.Hostname()
+	if owner == "" {
+		owner = fmt.Sprintf("pid-%d", os.Getpid())
```
fuchstim, line 55: PIDs aren't globally unique, e.g. in k8s the main process always runs as PID 0. Consider making the owner explicitly configurable via env var / config property. But since this value is only useful for debugging feel free to ignore.

s3/leaser.go
```diff
@@ -0,0 +1,310 @@
...
+
+	key := l.lockKey()
+	_, err := l.client.s3.DeleteObject(ctx, &s3.DeleteObjectInput{
+		Bucket:  aws.String(l.client.Bucket),
+		Key:     aws.String(key),
+		IfMatch: aws.String(lease.ETag),
+	})
+	if err != nil {
+		if isNotExists(err) || isNotFoundError(err) {
+			return nil
```
fuchstim: Consider adding a log here, releasing a lease that doesn't exist means it must have been deleted unexpectedly

fuchstim:
>>3. Custom JSON marshalling on Lease (s3/leaser.go:263)
if you implement the (Un)Marshal methods on litestream.Lease itself you can make it more portable and save yourself the conversions here

>The reason for the lockFile struct with float64 expires_at is to match the format in the reference article (Unix timestamp as float). However, I can look into using time.Time directly with custom JSON marshalling if preferred. What do you think?

Maybe this is personal preference. IMO, debugging lock files would be easier using the time.Time JSON format (RFC3339) rather than unix epoch. Implementing the marshalling on the litestream.Lease level (or ideally doing away with any custom serialization logic) also makes it easier to implement future lease backends without re-implementing the serialization for each provider. I'll leave it up to you.

s3/leaser.go
corylanou: Fixed in commit b6b4aac: Added `If-Match: lease.ETag` to the DeleteObject call. If the ETag doesn't match (412 Precondition Failed), we now return `ErrLeaseNotHeld` to prevent stale holders from deleting another leader's lease.

leaser.go
corylanou: Fixed in commit 619faf4: Renamed `Token` to `Generation` throughout the codebase.

s3/leaser.go
corylanou: Fixed in commit 9ac54e3: Removed the `lockFile` struct and now marshal/unmarshal `litestream.Lease` directly. The lock file now uses RFC3339 format (e.g., `"expires_at": "2026-01-30T15:24:33Z"`) which is easier to debug than Unix timestamps.

s3/leaser.go
corylanou: The `Owner` field is already exported on the `Leaser` struct, so callers can override it after `NewLeaser()`:
```go
leaser := s3.NewLeaser(client)
leaser.Owner = "my-custom-owner"
```
This allows users to set a meaningful identifier in k8s (e.g., pod name) without requiring env var support in the library itself.

s3/leaser.go
corylanou: Fixed in commit 9ac54e3: Added `l.logger.Warn("lease already deleted unexpectedly", ...)` when the lease file is not found during release.

fuchstim:
PR lgtm, thanks for entertaining my comments. Lets see what Ben thinks

corylanou:
> PR lgtm, thanks for entertaining my comments. Lets see what Ben thinks

Thanks for the reviews and assistance!

s3/leaser.go
benbjohnson: @corylanou I'd suggest going one step further and returning an `ErrLeaseAlreadyReleased` error but having the caller log the error.

s3/leaser.go
```diff
@@ -0,0 +1,294 @@
...
+}
+
+func (l *Leaser) ReleaseLease(ctx context.Context, lease *litestream.Lease) error {
+	if err := l.client.Init(ctx); err != nil {
+		return fmt.Errorf("init s3 client: %w", err)
+	}
+
+	if lease == nil || lease.ETag == "" {
+		return litestream.ErrLeaseNotHeld
+	}
```
benbjohnson: These seem like different errors from the precondition failure that returns `ErrLeaseNotHeld` below. These should return `lease required` and `lease etag required` instead.

s3/leaser.go
```diff
@@ -0,0 +1,294 @@
...
+const (
+	DefaultLeaseTTL  = 30 * time.Second
+	DefaultLeasePath = "lock.json"
+	LeaserType       = "s3"
+)
+
+var _ litestream.Leaser = (*Leaser)(nil)
+
+type Leaser struct {
+	client *ReplicaClient
```
benbjohnson: I'm not sure we want to tie the Leaser to the ReplicaClient. I think it should be configured separately from the replica as you could have a case where you're replicating to something like SFTP but you're using S3 for actual storage (since SFTP can't support leasing).

s3/leaser.go
```diff
@@ -0,0 +1,294 @@
...
+	}
+
+	newETag, err := l.writeLease(ctx, newLease, lease.ETag)
+	if err != nil {
+		var leaseErr *litestream.LeaseExistsError
+		if errors.As(err, &leaseErr) {
+			return nil, litestream.ErrLeaseNotHeld
+		}
+		return nil, err
+	}
```
benbjohnson, line 155: You don't need to change this but it annoys me that AI doesn't write this as:

```go
var leaseErr *litestream.LeaseExistsError
if errors.As(err, &leaseErr) {
	return nil, litestream.ErrLeaseNotHeld
} else if err != nil {
	return nil, err
}
```

It's not a hill I'm going to die on. Just venting. :)

s3/leaser.go
```diff
@@ -0,0 +1,294 @@
...
+
+	data, err := io.ReadAll(out.Body)
+	if err != nil {
+		return nil, "", fmt.Errorf("read lock file: %w", err)
+	}
+
+	var lease litestream.Lease
+	if err := json.Unmarshal(data, &lease); err != nil {
+		return nil, "", fmt.Errorf("unmarshal lock file: %w", err)
+	}
```
benbjohnson: I wonder why AI did it this way rather than use `json.Decoder`

benbjohnson (changes requested):

corylanou:
Addressed the remaining feedback in fbf6364:

1. **Line 158 (error messages)**: Added `ErrLeaseRequired` and `ErrLeaseETagRequired` for more specific error messages instead of returning `ErrLeaseNotHeld` for input validation.

2. **Line 31 (decouple from ReplicaClient)**: Leaser now defines its own `S3API` interface and has `Bucket` and `Path` fields. The `SetClient()` method accepts any implementation of the interface, allowing independent configuration for use cases like replicating to SFTP while using S3 for leasing.

3. **Line 139 (code style)**: Acknowledged! Fair point about the style - I'll keep that pattern in mind.

4. **Line 209 (json.Decoder)**: Good catch! Changed to use `json.NewDecoder(out.Body).Decode(&lease)` instead of `io.ReadAll` + `json.Unmarshal`.

corylanou:
Fixed in fbbdc8a: Now returns `ErrLeaseAlreadyReleased` instead of logging internally, allowing the caller to decide how to handle it.

benbjohnson (approved):

PR #1076, opened 2026-02-02
Stream restore plan selection across levels

benbjohnson:
Summary
- refactor `CalcRestorePlan` to stream iterators per level, picking the next file with the longest contiguous TXID range instead of collecting all level files first
- add level cursors that concurrently advance through iterators, allowing gaps at higher levels to be filled by longer ranges from lower levels
- expand test coverage to ensure the new planner prefers longer candidates and can recover from gaps using lower levels

Testing
- Not run (not requested)

### DIFF
--- a/restore_fuzz_test.go
+++ b/restore_fuzz_test.go
@@ -1,0 +1,135 @@
+package litestream_test
+
+import (
+	"math/rand"
+	"path/filepath"
+	"sync"
+	"testing"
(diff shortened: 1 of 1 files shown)


benbjohnson:
I tested this against 350+ databases internally and they all pass with `PRAGMA integrity_check`. This PR also include a fuzz test which was run for a while too.

PR #1078, opened 2026-02-03
fix: detect WAL changes by mtime

### LINKED ISSUE #1037: Writes silently fail to replicate when WAL size unchanged
## Summary

Litestream 0.5.6 silently stops replicating databases to S3 while writes continue to occur. No errors are logged. The only recovery is restarting the litestream process.

## Root Cause

PR #946 (commit 9d264870, merged January 6, 2026) added a change detection optimization to `DB.monitor()` that skips calling `Sync()` when WAL size and header are unchanged. The commit comment states:

> "stat() the WAL file to get its size (changes on any write)"

This assumption is incorrect. **WAL size does not change on every write.** SQLite can reuse space within the WAL, keeping size constant while writes occur.

**Before #946:** `Sync()` was called unconditionally on every monitor tick.

**After #946:** `Sync()` is skipped when `walSize == lastWALSize && walHeader == lastWALHeader`.

The fix is to use `mtime` instead of size, since mtime reliably updates on every write.

## Environment

- **Litestream version**: 0.5.6
- **Deployment**: Podman container on Linux (production) and macOS (local)
- **Storage backend**: S3-compatible (Cloudflare R2, MinIO)
- **Bug reproduced on both environments**

## Symptoms

1. `litestream_sync_count` metric stops incrementing
2. `litestream_txid` metric stops advancing
3. VFS queries return stale data
4. No errors in litestream logs
5. Database writes continue normally (verified via WAL file mtime changes)
6. Restarting litestream temporarily fixes the issue, but it recurs

---

## Reproduction Test Results

I built a patched version adding `mtime` to the change detection and compared against 0.5.6.

### Patched Code

```go
var lastWALSize int64
var lastWALTime time.Time
var lastWALHeader []byte

// ...

walSize := fi.Size()
walModTime := fi.ModTime()

if walSize == lastWALSize &&
    walModTime.Equal(lastWALTime) &&
    bytes.Equal(walHeader, lastWALHeader) {
    continue
}

lastWALSize = walSize
lastWALTime = walModTime
lastWALHeader = walHeader
```

### Test Setup

1. Create a test table:
   ```sql
   CREATE TABLE sync_test (id INTEGER PRIMARY KEY, updated_at TEXT);
   INSERT INTO sync_test (id, updated_at) VALUES (1, datetime('now'));
   ```
2. Start litestream replicating the database
3. Perform 10 writes (2 second intervals)
4. Record `sync_count`, `txid`, WAL size, and WAL header after each write

```bash
for i in $(seq 1 10); do
  sqlite3 /path/to/test.db "UPDATE sync_test SET updated_at = datetime('now') WHERE id = 1"
  sleep 2
  SYNC=$(wget -qO- http://localhost:9090/metrics | grep 'litestream_sync_count{db=\"/path/to/test.db\"}' | awk '{print $2}')
  TXID=$(wget -qO- http://localhost:9090/metrics | grep 'litestream_txid{db=\"/path/to/test.db\"}' | awk '{print $2}')
  SIZE=$(stat /path/to/test.db-wal | grep "Size:" | awk '{print $2}')
  HEADER=$(xxd -l 32 -p /path/to/test.db-wal | tr -d '\n')
  printf "[%02d] sync=%s txid=%s size=%s header=%s\n" $i "$SYNC" "$TXID" "$SIZE" "$HEADER"
done
```

### Results: Litestream 0.5.6

```
Initial: sync=1 txid=47

[01] sync=1 txid=47 size=201912 header=377f0682002de218...a12af3fd
[02] sync=1 txid=47 size=201912 header=377f0682002de218...a12af3fd
[03] sync=1 txid=47 size=201912 header=377f0682002de218...a12af3fd
[04] sync=1 txid=47 size=201912 header=377f0682002de218...a12af3fd
[05] sync=1 txid=47 size=201912 header=377f0682002de218...a12af3fd
[06] sync=1 txid=47 size=201912 header=377f0682002de218...a12af3fd
[07] sync=1 txid=47 size=201912 header=377f0682002de218...a12af3fd
[08] sync=1 txid=47 size=201912 header=377f0682002de218...a12af3fd
[09] sync=1 txid=47 size=201912 header=377f0682002de218...a12af3fd
[10] sync=3 txid=49 size=206032 header=377f0682002de218...74107f10
```

**STALLED** - 9 writes missed. Replication only occurred on iteration 10 when WAL grew and header changed.

### Results: Patched Litestream (with mtime check)

```
Initial: sync=1 txid=49

[01] sync=2  txid=50 size=206032 header=377f0682002de218...74107f10
[02] sync=3  txid=51 size=206032 header=377f0682002de218...74107f10
[03] sync=4  txid=52 size=206032 header=377f0682002de218...74107f10
[04] sync=5  txid=53 size=206032 header=377f0682002de218...74107f10
[05] sync=6  txid=54 size=206032 header=377f0682002de218...74107f10
[06] sync=7  txid=55 size=206032 header=377f0682002de218...74107f10
[07] sync=8  txid=56 size=206032 header=377f0682002de218...74107f10
[08] sync=9  txid=57 size=206032 header=377f0682002de218...74107f10
[09] sync=10 txid=58 size=206032 header=377f0682002de218...74107f10
[10] sync=12 txid=60 size=206032 header=377f0682002de218...964bede0
```

**WORKING** - Every write replicated. Iterations 1-9 had constant size and header, but mtime detection triggered replication.

### Key Observation

Iterations 1-9: WAL size constant, WAL header constant. Only mtime changed.

---

## Proposed Fix

Use `mtime` instead of (or in addition to) WAL size for change detection. mtime reliably updates on every write.

Potential considerations with `mtime`:

- **Timestamp resolution**: Some filesystems have low-resolution timestamps. Rapid writes within the resolution window may not update mtime.
- **Mount options**: Certain mount options or container configurations may affect timestamp behavior.

corylanou:
## Summary
Use WAL mtime in the monitor loop’s cheap change detection and add a regression test for mtime-only WAL changes.

## Problem
Issue #1037 reports replication stalling when WAL size and header remain unchanged even though writes continue. SQLite can reuse WAL space, so size/header can stay constant while WAL mtime changes on each write. The current optimization skips Sync() in that case, so replication silently stalls.

## Solution
Track WAL mtime alongside size and header. Skip Sync() only if size, header, and mtime are unchanged.

## Scope
**In scope:**
- Add WAL mtime to monitor loop change detection
- Add a regression test for mtime-only changes

**Not in scope:**
- WAL reader/format changes
- Replica backend changes
- New metrics/logging

## Test Plan
```bash
go test -run TestDB_Monitor_ -v ./...
```

## Related
- Fixes #1037

### DIFF
--- a/db.go
+++ b/db.go
@@ -1137,6 +1137,7 @@ func (db *DB) verify(ctx context.Context) (info syncInfo, err error) {
 	info.offset = dec.Header().WALOffset + dec.Header().WALSize
 	info.salt1 = dec.Header().WALSalt1
 	info.salt2 = dec.Header().WALSalt2
+	info.commit = dec.Header().Commit
 
 	// If LTX WAL offset is larger than real WAL then the WAL has been truncated.
 	if fi, err := os.Stat(db.WALPath()); err != nil {
@@ -1323,6 +1324,7 @@ type syncInfo struct {
 	offset       int64 // end of the previous LTX read
 	salt1        uint32
 	salt2        uint32
+	commit       uint32
 	snapshotting bool   // if true, a full snapshot is required
 	reason       string // reason for snapshot
 }
@@ -1469,7 +1471,7 @@ func (db *DB) sync(ctx context.Context, checkpointing bool, info syncInfo) (sync
 			return false, fmt.Errorf("write ltx from db: %w", err)
 		}
 	} else {
-		if err := db.writeLTXFromWAL(ctx, enc, walFile, pageMap); err != nil {
+		if err := db.writeLTXFromWAL(ctx, enc, walFile, pageMap, info.commit); err != nil {
 			return false, fmt.Errorf("write ltx from wal: %w", err)
 		}
 	}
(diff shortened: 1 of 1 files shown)


corylanou:
Implemented WAL change detection without relying on filesystem mtime granularity.

Instead of mtime, DB.monitor() now compares the WAL-index (-shm) mxFrame value (plus WAL size & WAL header). mxFrame advances on each commit even when SQLite reuses WAL space, and it avoids missing writes on coarse mtime filesystems.

Also fixed a related restore/compaction edge case exposed by the integration suite: when SQLite bumps the post-commit DB size by multiple pages but only some of the new pages appear in the WAL, we now encode zero-filled pages for the missing newly-added page numbers so compaction can always rebuild a valid snapshot.

Re: your question: there still isn't an fsnotify-based watcher on the WAL itself (only directory watching in cmd/litestream/directory_watcher.go); this remains polling-based via the monitor loop.

corylanou:
Superseded by #1087 which implements the same SHM mxFrame approach (avoiding mtime granularity issues) and also closes the duplicate issue #1083. Thank you for the initial implementation work on this!

PR #1079, opened 2026-02-03
Preserve compaction timestamps for PITR

### LINKED ISSUE #1038: PITR fails after L0 retention expires, even when data exists in compacted files
## Summary

Timestamp-based restore fails with "timestamp does not exist" after L0 files are compacted, even when the data exists in L1+ files.

## Root Cause

`TimeBounds()` only queries level 0:
```go
itr, err := r.Client.LTXFiles(ctx, 0, 0, false)  // Level 0 only
```

`CalcRestorePlan()` correctly iterates all levels:
```go
for level := maxLevel; level >= 0; level-- {
    a, err := FindLTXFiles(ctx, client, level, ...)
}
```

## Behavior

1. L0 files exist → restore works
2. L0 files compacted to L1+ → restore rejected
3. Data still exists in L1+ and `CalcRestorePlan()` would succeed if allowed to run

## Reproduction

1. Start litestream replicating a database with default L0 retention (5 minutes)
2. Write data and note the timestamp
3. Wait for L0 → L1 compaction and L0 retention to expire (~6 minutes)
4. Verify L1+ files exist:
   ```bash
   aws s3 ls s3://bucket/db/ltx/L1/ --recursive
   ```
5. Attempt restore to the earlier timestamp:
   ```bash
   litestream restore -timestamp "2026-01-24T10:00:00Z" -o restored.db s3://bucket/db
   ```
6. Observe failure: `timestamp does not exist`

## Proposed Fix

Modify `TimeBounds()` to iterate all compaction levels:

```go
func (r *Replica) TimeBounds(ctx context.Context) (min, max time.Time, err error) {
    // ... setup code ...

    // Check all compaction levels, not just L0
    for level := 0; level <= SnapshotLevel; level++ {
        itr, err := r.Client.LTXFiles(ctx, level, 0, false)
        if err != nil {
            return min, max, err
        }

        for itr.Next() {
            info := itr.Item()
            if min.IsZero() || info.CreatedAt.Before(min) {
                min = info.CreatedAt
            }
            if info.CreatedAt.After(max) {
                max = info.CreatedAt
            }
        }
        if err := itr.Close(); err != nil {
            return min, max, err
        }
    }
    return min, max, nil
}
```

corylanou:
## Summary
- Preserve earliest LTX timestamps during compaction so PITR bounds survive L0 retention.
- Use metadata timestamps when computing replica time bounds.

## Problem
Timestamp-based restore rejects valid times once L0 files expire even though L1+ contains the data. Investigation evidence:
- `ltx.Compactor` encodes the output header timestamp from the last input header (`maxHdr.Timestamp`), so compacted files report compaction time instead of the earliest source timestamp.
- `Replica.TimeBounds()` uses `CreatedAt` from `LTXFiles()` to validate restore timestamps.

## Solution
- Add a header timestamp override to the LTX compactor and set it to the oldest source file timestamp during compaction.
- Switch `Replica.TimeBounds()` to request metadata timestamps so remote backends use the preserved timestamps.

## Scope
**In scope:**
- Preserve earliest `CreatedAt` for compacted LTX files.
- Accurate time bounds for restore validation.

**Not in scope:**
- Changes to retention policy or restore planning beyond time-bound validation.

## Test Plan
- `go test ./...`

## Related
- Fixes #1038

### DIFF
--- a/compactor.go
+++ b/compactor.go
@@ -125,6 +125,7 @@ func (c *Compactor) Compact(ctx context.Context, dstLevel int) (*ltx.FileInfo, e
 	}()
 
 	var minTXID, maxTXID ltx.TXID
+	var oldestCreatedAt time.Time
 	for itr.Next() {
 		info := itr.Item()
 
@@ -133,6 +134,9 @@ func (c *Compactor) Compact(ctx context.Context, dstLevel int) (*ltx.FileInfo, e
 		}
 		if maxTXID == 0 || info.MaxTXID > maxTXID {
 			maxTXID = info.MaxTXID
+		}
+		if !info.CreatedAt.IsZero() && (oldestCreatedAt.IsZero() || info.CreatedAt.Before(oldestCreatedAt)) {
+			oldestCreatedAt = info.CreatedAt
 		}
 
 		if c.LocalFileOpener != nil {
@@ -162,6 +166,10 @@ func (c *Compactor) Compact(ctx context.Context, dstLevel int) (*ltx.FileInfo, e
 			return
 		}
 		comp.HeaderFlags = ltx.HeaderFlagNoChecksum
+		if !oldestCreatedAt.IsZero() {
+			timestamp := oldestCreatedAt.UTC().UnixMilli()
+			comp.HeaderTimestamp = &timestamp
+		}
 		_ = pw.CloseWithError(comp.Compact(ctx))
 	}()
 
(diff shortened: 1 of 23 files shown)


corylanou:
Closing here; root fix should land in superfly/ltx (compactor timestamp preservation). Opened ltx issue: https://github.com/superfly/ltx/issues/74

PR #1086, opened 2026-02-07
fix(compactor): add mutex to protect Compactor.client from concurrent access

### LINKED ISSUE #1085: fix: FuzzRestoreWithMissingCompactedFile race condition detected with -race
## Description

The `FuzzRestoreWithMissingCompactedFile` fuzz test has a race condition detected when running with `-race` on the `main` branch.

## Reproduction

```bash
go test -race -count=1 -run FuzzRestoreWithMissingCompactedFile -v .
```

## Output

```
--- FAIL: FuzzRestoreWithMissingCompactedFile (3.76s)
    --- PASS: FuzzRestoreWithMissingCompactedFile/seed#0 (1.11s)
    --- PASS: FuzzRestoreWithMissingCompactedFile/seed#1 (1.42s)
    --- FAIL: FuzzRestoreWithMissingCompactedFile/seed#2 (1.24s)
        testing.go:1617: race detected during execution of test
FAIL
```

The race is intermittent - `seed#0` sometimes passes and sometimes fails, while `seed#2` consistently triggers it. The race appears to occur during the restore/compaction flow when multiple operations are happening concurrently.

## Environment

- Go: 1.24
- OS: macOS (Darwin 25.1.0)
- Branch: `main` (commit `90fc5b6`)

## Notes

This was discovered while running the full test suite with `-race` during work on PR #1021. The issue exists on the `main` branch and is not introduced by any PR changes.

corylanou:
## Summary

- Fix data race in `Compactor.client` field detected by `-race` in `FuzzRestoreWithMissingCompactedFile`
- Root cause: `ensureCompactorClient()` lazily syncs the compactor's client on every operation, creating a race when multiple goroutines call it concurrently from `Store.Open()` monitors
- Fix: set the compactor client once in `DB.Open()` (where `Replica` is already assigned), eliminating `ensureCompactorClient()`, `SetClient()`, `Client()`, and the mutex entirely
- Tests that previously overwrote `db.Replica` after `Open()` are restructured to set Replica before `Open()`

**Race trace (on main):** Goroutine reading `Compactor.Client()` via compaction level monitor races with goroutine writing `Compactor.SetClient()` via snapshot level monitor — both enter through `DB.ensureCompactorClient()`.

Fixes #1085

## Test plan

- [x] `go test -race -count=5 -run FuzzRestoreWithMissingCompactedFile -v .` — Compactor.client race eliminated
- [x] `go test -race ./...` — full test suite passes
- [x] Pre-commit hooks pass (go-imports, go-vet, staticcheck)

### DIFF
--- a/compactor.go
+++ b/compactor.go
@@ -163,6 +163,7 @@ func (c *Compactor) Compact(ctx context.Context, dstLevel int) (*ltx.FileInfo, e
 		c.CacheSetter(dstLevel, info)
 	}
 
+	// Verify level consistency if enabled
 	if c.VerifyCompaction {
 		if err := c.VerifyLevelConsistency(ctx, dstLevel); err != nil {
 			c.logger.Warn("post-compaction verification failed",
@@ -191,11 +192,13 @@ func (c *Compactor) VerifyLevelConsistency(ctx context.Context, level int) error
 	for itr.Next() {
 		info := itr.Item()
 
+		// Skip first file - nothing to compare against
 		if prevInfo == nil {
 			prevInfo = info
 			continue
 		}
 
+		// Check for TXID contiguity: prev.MaxTXID + 1 should equal curr.MinTXID
 		expectedMinTXID := prevInfo.MaxTXID + 1
 		if info.MinTXID != expectedMinTXID {
 			if info.MinTXID > expectedMinTXID {


benbjohnson (changes requested):
@corylanou This one seems weird. Do you know why we're updating the Compactor's `client` field at all? It seems like we should only be setting its client once immediately after the Replica is set. Or we could move `compactor` to the `Replica` instead of having it on `DB` and then it can be initialized at the same time.

corylanou:
> @corylanou This one seems weird. Do you know why we're updating the Compactor's `client` field at all? It seems like we should only be setting its client once immediately after the Replica is set. Or we could move `compactor` to the `Replica` instead of having it on `DB` and then it can be initialized at the same time.

That would be a better firx for sure.  I wasn't sure if you needed the ability to set the client once it was created.  It's possible that the "SetClient" is only for testing?  If so, we should be able to hopefully change the way we test.  This came up as a fuzz test failing a race condition.  I'll take another look if we don't think we need the SetClient method.

corylanou:
@benbjohnson turns out the real fix was to change how we tested it.  Removed all of the race conditions, etc. Much cleaner now.

benbjohnson (approved):
Yeah, that seems way better :+1:

PR #1087, opened 2026-02-08
fix(db): detect WAL changes via shm mxFrame

### LINKED ISSUE #1037: Writes silently fail to replicate when WAL size unchanged

### LINKED ISSUE #1083: Silent Replication Failure in v0.5.6+ When SQLite Reuses WAL Space
# Silent Replication Failure in v0.5.6+ When SQLite Reuses WAL Space

## Summary

Litestream v0.5.6 and v0.5.7 contain a critical bug that causes complete replication failure with no error messages. The bug occurs when SQLite reuses space in the WAL file after checkpointing, causing Litestream to incorrectly conclude that no writes have occurred and skip sync operations entirely.

This results in:
- Frozen metrics (`sync_count`, `checkpoint_count`, `txid` stop advancing)
- Stale backups (no new data replicated to S3)
- No error messages in logs (silent failure)
- Only recovery method is restarting Litestream

## Affected Versions

- ❌ **v0.5.6** - Broken
- ❌ **v0.5.7** - Broken
- ✅ **v0.5.5** - Works correctly

## Environment

- **Deployment:** ECS (standalone Litestream service, separate from application)
- **Configuration:**
  - `checkpoint-interval: 1m`
  - `sync-interval: 10s` (S3)
  - `snapshot-interval: 5m`
  - Directory mode with `watch: true` and `recursive: true`
- **Application:** C# .NET with `PRAGMA wal_autocheckpoint=0` set on every connection
- **Storage:** S3 (us-east-2)

### Setup
1. Deploy Litestream v0.5.7 with directory monitoring
2. Configure application to set `PRAGMA wal_autocheckpoint=0` on all connections
3. Application writes data continuously for several minutes
4. Stop application writes

### Observed Behavior (v0.5.7)

**Initial state (while app is writing):**
```
litestream_checkpoint_count: 7
litestream_sync_count: 17
litestream_txid: 9298
litestream_wal_size: 0 bytes
Actual WAL file: 5.7 MB
```

**After app stops writing (waited 15+ minutes):**
```
litestream_checkpoint_count: 7 (unchanged)
litestream_sync_count: 17 (unchanged)
litestream_txid: 9298 (unchanged)
litestream_wal_size: 0 bytes (unchanged)
Actual WAL file: 5.7 MB (actively being written to)
```

**Key observations:**
- Metrics completely frozen
- WAL file continues to grow and be modified (mtime changes)
- No errors in Litestream logs
- Restore from S3 is missing 14+ minutes of data
- Only fix is restarting Litestream

### Expected Behavior (v0.5.5)

**Initial state (while app is writing):**
```
litestream_checkpoint_count: 234
litestream_sync_count: 234
litestream_txid: 9553
litestream_wal_size: 6,035,832 bytes (6.0 MB)
Actual WAL file: 5.8 MB
```

**After app stops writing (waited 6 minutes):**
```
litestream_checkpoint_count: 722 (increased by 488)
litestream_sync_count: 721 (increased by 487)
litestream_txid: 10055 (increased by 502)
litestream_wal_size: 6,035,832 bytes (still accurate)
Actual WAL file: 5.8 MB
```

**Key observations:**
- Metrics continue advancing even after writes stop
- WAL size tracking is accurate
- Restore from S3 contains all data up to the last write
- Replication continues working normally

## Root Cause Analysis

The bug appears to be related to an optimization added in v0.5.6 that checks WAL file size and header to determine if sync is needed. The problematic logic likely looks like:

```go
if walSize == lastWALSize && bytes.Equal(walHeader, lastWALHeader) {
    continue  // Skip Sync()
}
```

**The problem:** SQLite reuses WAL space after checkpointing old frames. When this happens:
1. New writes occur to the database
2. WAL file size stays the same (SQLite reused space from checkpointed frames)
3. WAL mtime changes (file is being written to)
4. Litestream sees "size unchanged" and skips `Sync()`
5. No sync = no notification to `Replica.monitor` goroutines
6. Replication stops completely

This optimization incorrectly assumes that unchanged WAL size means no writes occurred, which is false when SQLite reuses space.

### Production Risk
- **Silent data loss:** Backups become stale with no indication
- **No alerting possible:** No errors logged, metrics appear "healthy" (non-zero values)
- **Point-in-time recovery broken:** Cannot restore to recent states
- **Manual intervention required:** Must restart Litestream to recover

### Affected Use Cases
- Applications with continuous writes (most common case)
- Applications with bursty write patterns
- Any scenario where SQLite checkpoints and reuses WAL space
- Multi-tenant applications with many databases

## Comparison: v0.5.5 vs v0.5.7

| Metric | v0.5.7 (Broken) | v0.5.5 (Working) |
|--------|-----------------|------------------|
| Checkpoints in 6 min | 0 (frozen at 7) | 488 (234 → 722) |
| Syncs in 6 min | 0 (frozen at 17) | 487 (234 → 721) |
| Transaction ID | Frozen at 9298 | Advanced to 10055 |
| WAL size tracking | 0 bytes (wrong) | 6.0 MB (correct) |
| Restore freshness | Missing 14+ min | All data present |
| Error messages | None (silent) | N/A (working) |

## Suggested Fix

The optimization should include WAL file modification time (mtime) in the check:

```go
if walSize == lastWALSize && walMtime == lastWALMtime && bytes.Equal(walHeader, lastWALHeader) {
    continue  // Skip Sync()
}
```

Alternatively, remove the optimization entirely and revert to the v0.5.5 behavior of calling `Sync()` on every monitor tick. Since Litestream already skips actual replication when there are no changes, the optimization provides minimal benefit while introducing this critical bug.

## Workaround

**Downgrade to v0.5.5** until this bug is fixed:

```dockerfile
FROM litestream/litestream:0.5.5
```

Or for binary installations:
```bash
wget https://github.com/benbjohnson/litestream/releases/download/v0.5.5/litestream-v0.5.5-linux-amd64.deb
dpkg -i litestream-v0.5.5-linux-amd64.deb
```

## Additional Notes

- This bug makes v0.5.6+ unsuitable for production use
- The bug is deterministic and reproducible
- No configuration changes can work around it (tried various `checkpoint-interval` and `sync-interval` values)
- The bug affects all storage backends (S3, file, etc.)
- `validation-interval` is documented as non-functional in v0.5.x, so it cannot detect/correct this issue

## Request

Please prioritize fixing this critical bug in v0.5.8. The silent nature of the failure makes it particularly dangerous for production deployments.

Thank you for maintaining this excellent tool!

### LINKED ISSUE #1104: L0 files accumulate indefinitely and fill disk in v0.5.5
## Bug Description

 L0 files (.ltx files in ltx/0/ directory) accumulate continuously in local storage and are never cleaned up, eventually filling the disk completely. This happens even though:

Files are successfully uploaded to S3 (0001/ directory)
sync_count metrics show successful syncs
sync_error_count is 0
L0 retention is configured (5m)

## Environment

OS: Linux (Alpine in Docker)
Deployment: ECS on EC2
Storage: 98 GB EBS volume
Database: SQLite with WAL mode
Replica: S3 (us-east-2)

**Litestream version:**

Litestream v0.5.5

Linux (Alpine in Docker)

**Installation method:**
Docker

**Storage backend:**
s3

## Steps to Reproduce

Litestream configured in directory watcher mode. 1/5 Database that is having issues is largish (about 270MB).

**Expected behavior:**

L0 files should be deleted after:

Successful upload to S3, AND
L0 retention period (5m) has elapsed

**Actual behavior:**

L0 files (.ltx files in ltx/0/ directory) accumulate continuously in local storage and are never cleaned up, eventually filling the disk completely. This happens even though:

Files are successfully uploaded to S3 (0001/ directory)
sync_count metrics show successful syncs
sync_error_count is 0
L0 retention is configured (5m)

L0 files created every checkpoint interval (1m)
Each L0 file is ~200 MB
Files are uploaded to S3 successfully
Files remain in local ltx/0/ directory indefinitely
No local L1 directory (ltx/1/) is ever created
Disk fills at rate of ~200 MB/minute
After ~8 hours, 98 GB disk is 100% full

## Configuration

```yaml
addr: ":9090"
http-url: "http://localhost:8080"

# L0 Retention Configuration
# Controls how long L0 (level 0) files are kept after compaction to L1
# Only needed if using VFS read replicas - can be minimal for backup-only use
l0-retention: 5m
l0-max-retention: 24h

# Compaction levels configuration
# Level 1: L0 → L1 compaction
# Level 2: L1 → L2 compaction
# Level 3+: Higher level compaction
levels:
  - interval: 30s
  - interval: 24h
  - interval: 168h

dbs:
  - dir: /mnt/ebs
    pattern: "*.metadata"
    recursive: true
    watch: true
    checkpoint-interval: 1m
    
    replicas:
      - type: s3
        bucket: ${S3_BACKUP_BUCKET}
        path: ${HOSTNAME}/${DATABASE_NAME}
        region: ${AWS_REGION}
        retention: 168h
        sync-interval: 1m
        snapshot-interval: 1h
```

</details>

## Logs

<details><summary>No errors related to L0 retention or cleanup.</summary>

```text
Level=INFO msg="compaction complete" level=1 txid.min=... txid.max=... size=219076102
level=INFO msg="replicating to" type=s3 sync-interval=1m0s bucket=...
```

</details>

## Additional Context

This occurs during idle periods when the database has no new writes. The WAL file modification time updates during checkpoints (even with no new transactions), which may trigger Litestream to create new L0 files. However, the core issue is that successfully uploaded L0 files are never cleaned up locally.

corylanou:
## Summary
- Fix silent replication failure when SQLite reuses WAL space after checkpointing
- Add WAL-index (-shm) `mxFrame` to the monitor loop's cheap change detection, which advances on each commit even when WAL size and header remain unchanged
- Fix compaction edge case where SQLite bumps DB size by multiple pages but only some appear in the WAL — encode zero-filled pages for missing newly-added page numbers

Closes #1037
Closes #1083
Closes #1104
Supersedes #1078

## Context

The monitor loop's cheap change detection (added in v0.5.6) compared only WAL file size and 32-byte header to skip `Sync()` calls. When SQLite performs a checkpoint and reuses WAL space, both can remain unchanged despite new writes, causing replication to silently stop with no error messages.

The fix reads 4 bytes from the WAL-index (`-shm`) file at offset 16 (`mxFrame` — the count of valid WAL frames). This is updated atomically by SQLite on every write transaction, making it a reliable change indicator regardless of WAL space reuse. Both copies of `mxFrame` (offsets 16 and 64) are verified for consistency.

## Test plan
- [x] `TestReadSHMMxFrameKey` — verifies reading matching mxFrame copies
- [x] `TestReadSHMMxFrameKeyMismatch` — verifies error on inconsistent copies
- [x] `TestDB_Monitor_IgnoresWALModTimeChange` — verifies mtime-only changes don't trigger spurious syncs
- [x] Existing `TestDB_Monitor_CheapChangeDetection` still passes
- [x] Existing `TestDB_Monitor_DetectsSaltChangeAfterRestart` still passes
- [x] Full test suite passes with `-race`: `go test -race ./...`

### DIFF
--- a/db.go
+++ b/db.go
@@ -95,6 +95,14 @@ type DB struct {
 	// trigger a feedback loop because stale file size exceeds threshold.
 	lastSyncedWALOffset int64
 
+	// forceNextSnapshot forces the next verify() call to return
+	// snapshotting=true. Set by checkpoint() when the WAL restarts during
+	// checkpointing, which indicates external writes may have been
+	// auto-checkpointed to the DB while the read lock was released.
+	// Those pages would not be in the new WAL or any LTX file, so a full
+	// snapshot is needed to ensure complete coverage.
+	forceNextSnapshot bool
+
 	// last file info for each level
 	maxLTXFileInfos struct {
 		sync.Mutex
@@ -1167,6 +1175,19 @@ func (db *DB) verify(ctx context.Context) (info syncInfo, err error) {
 		return info, nil // first sync
 	}
 
+	if db.forceNextSnapshot {
+		db.forceNextSnapshot = false
+		hdr, err := readWALHeader(db.WALPath())
+		if err != nil {
+			return info, fmt.Errorf("read wal header for forced snapshot: %w", err)
+		}
+		info.offset = WALHeaderSize
+		info.salt1 = binary.BigEndian.Uint32(hdr[16:])
+		info.salt2 = binary.BigEndian.Uint32(hdr[20:])
+		info.reason = "WAL restarted during checkpoint"
+		return info, nil
+	}
+
 	// Determine last WAL offset we save from.
 	ltxPath := db.LTXPath(0, pos.TXID, pos.TXID)
 	ltxFile, err := os.Open(ltxPath)
@@ -1192,32 +1213,22 @@ func (db *DB) verify(ctx context.Context) (info syncInfo, err error) {
 	if fi, err := os.Stat(db.WALPath()); err != nil {
 		return info, fmt.Errorf("open wal file: %w", err)
 	} else if info.offset > fi.Size() {
-		// If we previously synced to the exact end of the WAL, this truncation
-		// is expected (normal checkpoint behavior). Reset position and continue
-		// incrementally rather than triggering a full snapshot. See issue #927.
-		if db.syncedToWALEnd {
-			db.syncedToWALEnd = false // Clear flag
+		db.syncedToWALEnd = false
 
-			// Read new WAL header to get current salt values
-			hdr, err := readWALHeader(db.WALPath())
-			if err != nil {
-				return info, fmt.Errorf("read wal header after expected truncation: %w", err)
-			}
-
-			info.offset = WALHeaderSize
-			info.salt1 = binary.BigEndian.Uint32(hdr[16:])
-			info.salt2 = binary.BigEndian.Uint32(hdr[20:])
-			info.snapshotting = false
-			info.reason = ""
-
-			db.Logger.Log(ctx, internal.LevelTrace, "wal truncated after sync to end (expected checkpoint)",
-				"new_salt1", info.salt1,
-				"new_salt2", info.salt2)
-
-			return info, nil
+		hdr, err := readWALHeader(db.WALPath())
+		if err != nil {
+			return info, fmt.Errorf("read wal header after wal truncation: %w", err)
 		}
 
-		info.reason = "wal truncated by another process"
+		info.offset = WALHeaderSize
+		info.salt1 = binary.BigEndian.Uint32(hdr[16:])
+		info.salt2 = binary.BigEndian.Uint32(hdr[20:])
+		info.reason = "wal truncated, snapshotting for complete page coverage"
+
+		db.Logger.Log(ctx, internal.LevelTrace, "wal truncated, forcing snapshot",
+			"new_salt1", info.salt1,
+			"new_salt2", info.salt2)
+
 		return info, nil
 	}
 
@@ -1272,8 +1283,6 @@ func (db *DB) verify(ctx context.Context) (info syncInfo, err error) {
 
 	slog.Debug("verify.2", "lastPageMatch", lastPageMatch)
 
-	// Salt has changed which could indicate a FULL checkpoint.
-	// If we have a last page match, then we can assume that the WAL has not been overwritten.
 	if !saltMatch {
 		db.Logger.Log(ctx, internal.LevelTrace, "wal restarted",
 			"salt1", salt1,
@@ -1281,14 +1290,7 @@ func (db *DB) verify(ctx context.Context) (info syncInfo, err error) {
 
 		info.offset = WALHeaderSize
 		info.salt1, info.salt2 = salt1, salt2
-
-		if detected, err := db.detectFullCheckpoint(ctx, [][2]uint32{{salt1, salt2}, {dec.Header().WALSalt1, dec.Header().WALSalt2}}); err != nil {
-			return info, fmt.Errorf("detect full checkpoint: %w", err)
-		} else if detected {
-			info.reason = "full or restart checkpoint detected, snapshotting"
-		} else {
-			info.snapshotting = false
-		}
+		info.reason = "wal restarted, snapshotting for complete page coverage"
 
 		return info, nil
 	}
@@ -1335,38 +1337,6 @@ func (db *DB) lastPageMatch(ctx context.Context, dec *ltx.Decoder, prevWALOffset
 		}
 		return true, nil // Page matches
 	}
-}
-
-// detectFullCheckpoint attempts to detect checks if a FULL or RESTART checkpoint
-// has occurred and we may have missed some frames.
-func (db *DB) detectFullCheckpoint(ctx context.Context, knownSalts [][2]uint32) (bool, error) {
-	walFile, err := os.Open(db.WALPath())
-	if err != nil {
-		return false, fmt.Errorf("open wal file: %w", err)
-	}
-	defer walFile.Close()
-
-	var lastKnownSalt [2]uint32
-	if len(knownSalts) > 0 {
-		lastKnownSalt = knownSalts[len(knownSalts)-1]
-	}
-
-	rd, err := NewWALReader(walFile, db.Logger)
-	if err != nil {
-		return false, fmt.Errorf("new wal reader: %w", err)
-	}
-	m, err := rd.FrameSaltsUntil(ctx, lastKnownSalt)
-	if err != nil {
-		return false, fmt.Errorf("frame salts until: %w", err)
-	}
-
-	// Remove known salts from the map.
-	for _, salt := range knownSalts {
-		delete(m, salt)
-	}
-
-	// If we have more than one unknown salt, then we have a FULL or RESTART checkpoint.
-	return len(m) >= 1, nil
 }
 
 type syncInfo struct {
@@ -1695,6 +1665,8 @@ func (db *DB) checkpoint(ctx context.Context, mode string) error {
 		db.syncedSinceCheckpoint = false
 		return nil
 	}
+
+	db.forceNextSnapshot = true
 
 	// Start a transaction. This will be promoted immediately after.
 	tx, err := db.db.BeginTx(ctx, nil)
--- a/db_internal_test.go
+++ b/db_internal_test.go
@@ -1426,6 +1426,107 @@ func TestDB_Issue994_RunawayDiskUsage(t *testing.T) {
 	}
 }
 
+// TestDB_Verify_ForceSnapshotAfterCheckpointWALRestart verifies that when
+// forceNextSnapshot is set (simulating a WAL restart during checkpoint),
+// verify() returns snapshotting=true with proper WAL header values.
+//
+// This tests the fix for a race condition where external writes could be
+// auto-checkpointed to the database while execCheckpoint() releases the
+// read lock. Those pages would be missing from the new WAL, so a full
+// snapshot is required.
+func TestDB_Verify_ForceSnapshotAfterCheckpointWALRestart(t *testing.T) {
+	dir := t.TempDir()
+	dbPath := filepath.Join(dir, "db")
+
+	db := NewDB(dbPath)
+	db.MonitorInterval = 0
+	db.Replica = NewReplica(db)
+	db.Replica.Client = &testReplicaClient{dir: t.TempDir()}
+	db.Replica.MonitorEnabled = false
+	if err := db.Open(); err != nil {
+		t.Fatal(err)
+	}
+	defer func() {
+		if err := db.Close(context.Background()); err != nil {
+			t.Fatal(err)
+		}
+	}()
+
+	sqldb, err := sql.Open("sqlite", dbPath)
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer sqldb.Close()
+
+	if _, err := sqldb.Exec(`PRAGMA journal_mode = wal;`); err != nil {
+		t.Fatal(err)
+	}
+	if _, err := sqldb.Exec(`CREATE TABLE t (id INT, data TEXT)`); err != nil {
+		t.Fatal(err)
+	}
+	if _, err := sqldb.Exec(`INSERT INTO t VALUES (1, 'initial')`); err != nil {
+		t.Fatal(err)
+	}
+
+	ctx := context.Background()
+
+	if err := db.Sync(ctx); err != nil {
+		t.Fatal(err)
+	}
+
+	// Without forceNextSnapshot, verify should return snapshotting=false
+	// (incremental sync from where the last LTX left off).
+	info, err := db.verify(ctx)
+	if err != nil {
+		t.Fatalf("verify without flag: %v", err)
+	}
+	if info.snapshotting {
+		t.Fatal("verify without flag: expected snapshotting=false")
+	}
+
+	// Read current WAL header to get expected salt values.
+	walHdr, err := readWALHeader(db.WALPath())
+	if err != nil {
+		t.Fatalf("read WAL header: %v", err)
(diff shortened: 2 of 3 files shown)


db.go
```diff
@@ -1570,28 +1576,43 @@ func (db *DB) writeLTXFromDB(ctx context.Context, enc *ltx.Encoder, walFile *os.
...
+		for pgno := prevCommit + 1; pgno <= commit; pgno++ {
+			if pgno == lockPgno {
+				continue
+			}
+			if _, ok := pageMap[pgno]; ok {
+				continue
+			}
+			pgnos = append(pgnos, pgno)
+		}
+	}
```
benbjohnson: What is this supposed to be doing? I don't see how the `commit` field is related to the SHM `mxFrame` detection.

db.go
```diff
@@ -2105,6 +2130,15 @@ func (db *DB) monitor() {
 			continue
 		}
 
+		shmMxFrame, err := readSHMMxFrameKey(db.shmPath())
+		if err != nil {
```
benbjohnson, line 2136: We should `DEBUG` log that we couldn't read the SHM mxFrame

litestream.go
```diff
@@ -133,6 +133,27 @@ func readWALHeader(filename string) ([]byte, error) {
...
+		return [4]byte{}, err
+	}
+	defer f.Close()
+
+	if _, err := f.ReadAt(a[:], 16); err != nil {
+		return [4]byte{}, err
+	}
+	if _, err := f.ReadAt(b[:], 64); err != nil {
+		return [4]byte{}, err
+	}
```
benbjohnson: I don't think you even need to both comparing the two versions. I think they're double written as a consistency check but just reading the first should be sufficient.

However, if you do read both, it'd be more efficient to just do one `Read()` for both headers (bytes 0 through 96) and extract the two `mxFrame` fields from there.

db.go
corylanou: Removed. This was an unrelated compaction fix carried over from PR #1078 — it doesn't belong in this PR. Reverted all the `writeLTXFromWAL` changes, the `commit` field on `syncInfo`, and the `info.commit` assignment.

db.go
corylanou: Added `db.Logger.Debug("failed to read shm mxFrame", "error", err)` before the fallback sync.

litestream.go
corylanou: Simplified to only read the first copy at offset 16. Removed the second ReadAt, the mismatch check, and the `TestReadSHMMxFrameKeyMismatch` test.

corylanou:
Added a separate commit for the zero-fill fix that I've now isolated from the mxFrame change.

**Why this is needed:** The integration tests (TestWALGrowth, TestConcurrentOperations, TestS3AccessPointLocalStack) all fail during restore with:

```
nonsequential page numbers in snapshot transaction: N,M
```

When SQLite grows a database (increases page count), the new blank pages may not appear in the WAL — SQLite writes them directly to the database file. Without including these pages in incremental LTX files, compaction/restore produces snapshots with gaps in page numbers.

This is a pre-existing bug, but the mxFrame change exposes it more frequently because we now correctly detect WAL changes that were previously missed (the core fix in this PR). The first push (which included this fix) passed CI; the second push (which removed it per review) failed all three of those tests with the same error.

The fix tracks the previous `commit` (database size in pages) and fills in zero-filled pages for any page numbers between the old and new commit that aren't present in the WAL's page map.

db.go
corylanou: Update: it turns out this fix IS needed. After removing it, CI failed on TestWALGrowth, TestConcurrentOperations, and TestS3AccessPointLocalStack — all with the same error during restore:

```
nonsequential page numbers in snapshot transaction
```

When SQLite grows a database, new blank pages may not go through the WAL. Without including zero-filled entries for those pages, compaction/restore produces snapshots with gaps. This is a pre-existing bug, but the mxFrame change exposes it more because we now correctly detect WAL changes that were previously missed.

I've re-added it as a separate commit (14940ea) with a clear explanation.

benbjohnson:
> When SQLite grows a database (increases page count), the new blank pages may not appear in the WAL — SQLite writes them directly to the database file. Without including these pages in incremental LTX files, compaction/restore produces snapshots with gaps in page numbers.

This one surprises me. It seems like we should have hit this bug a long time ago if the WAL can extend the database size without actually writing pages. Can you isolate that fix to its own PR first and add a test case of how that can be reproduced with SQLite? I'm feeling skeptical that it can actually happen. :)

RichardAtRiva:
This fix didn't make it into the last release (0.5.8). Can we expect that this will be in the next release after (0.5.9)?

Issue #1013, opened 2026-01-17
IPC Control Commands - Phased Implementation


corylanou:
## Overview

Runtime control commands for dynamically managing database replication without restarting the daemon. Enables:
- **Multi-tenant deployments**: Selectively enable/disable replication per database
- **Failover with external locks**: Replication state doesn't persist across restarts
- **Dynamic provisioning**: Start replicating new databases without daemon restart

### Phase 1: Core Start/Stop ✅
**PR #1010** (merged)

- Unix socket-based control server (`socket.path` config)
- `litestream start -socket <path> <db>` - Start replication
- `litestream stop -socket <path> <db>` - Stop replication
- Store methods: `EnableDB`, `DisableDB`, `FindDB`, `RegisterDB`
- `DB.IsOpen()` method for state tracking
- HTTP-over-Unix-socket protocol with JSON requests

### Phase 2: Status & Query Commands
- [ ] `litestream status -socket <path> <db>` - Query replication status for a database
- [ ] `litestream list -socket <path>` - List all managed databases
- [ ] `litestream info -socket <path>` - Show daemon info (version, pid, uptime)

### Phase 3: Sync Command
- [ ] `litestream sync -socket <path> <db>` - Force immediate sync
- [ ] `--wait` flag to block until sync completes
- [ ] `--timeout` flag for maximum wait time

### Future Considerations (deferred)
- Config persistence (`persist-to-config`) - Only implement if requested
- HTTP API on separate port (currently just Unix socket)
- `auto-replicate` config flag for directory watcher

## Related
- Supersedes #1002
- Original implementation attempt: #1008

PR #1089, opened 2026-02-09
feat: add POST /sync-replicate endpoint for on-demand sync replication

chrismccord:
Add a new control socket endpoint that triggers immediate DB sync + replica upload, returning only after the data is durably stored on the replica (e.g. S3). This enables synchronous replication from any language via a simple HTTP POST to the Unix socket (by way of a small sqlite extension)

The endpoint:
- Accepts {"path": "/path/to/db", "timeout": 30}
- Calls DB.Sync() to flush WAL frames into LTX files
- Calls Replica.Sync() to upload LTX files to the configured replica
- Returns {"status": "ok", "path": "...", "txid": N} on success
- Returns appropriate errors for unknown DBs, missing replicas, or sync failures

This is the server-side component needed by external VFS shims (like a SQLite loadable extension) that intercept WAL xSync to trigger replication before COMMIT completes — giving applications opt-in synchronous durability guarantees per-transaction.

Includes tests for success, not-found, and no-replica error paths.

example extension:
https://github.com/chrismccord/litestream/tree/cm-sidecar-sync/ext/sync

### DIFF
--- a/server.go
+++ b/server.go
@@ -299,6 +299,13 @@ func (s *Server) handleSyncReplicate(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	beforeDBPos, err := db.Pos()
+	if err != nil {
+		writeJSONError(w, http.StatusInternalServerError, fmt.Sprintf("failed to get database position: %v", err), nil)
+		return
+	}
+	beforeReplicaPos := db.Replica.Pos()
+
 	if err := db.Sync(ctx); err != nil {
 		writeJSONError(w, http.StatusInternalServerError, fmt.Sprintf("sync failed: %v", err), nil)
 		return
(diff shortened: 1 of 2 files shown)


server.go
```diff
@@ -262,6 +263,65 @@ func (s *Server) handleTXID(w http.ResponseWriter, r *http.Request) {
...
+
+	if err := db.Sync(ctx); err != nil {
+		writeJSONError(w, http.StatusInternalServerError, fmt.Sprintf("sync failed: %v", err), nil)
+		return
+	}
+
+	if err := db.Replica.Sync(ctx); err != nil {
+		writeJSONError(w, http.StatusInternalServerError, fmt.Sprintf("replica sync failed: %v", err), nil)
+		return
+	}
```
benbjohnson, line 317: It might be worth grabbing the database position before you perform the sync and verify that the transaction ID is incremented on both the database and the replica after those syncs are performed. These calls can return no error in the event of a no-op.

benbjohnson:
One note for anyone else reading this pull request in the future: this is not true synchronous replication, as we are not blocking other reads until the sync is performed. Therefore, other processes and connections could still see the data before it is officially synced to S3.

corylanou:
Closing this PR as the feature has been shipped via #1120 (`feat(ipc): add sync command for on-demand database replication`), which merged on 2026-02-13.

The implementation in #1120 covers the same use case — on-demand sync triggering via the IPC socket — but follows the established IPC architecture from #1013 more closely:

- `POST /sync` endpoint on the Unix socket server
- `litestream sync` CLI command with `-wait` and `-timeout` flags
- Fire-and-forget mode (returns `sync_triggered` immediately) and blocking mode (waits for WAL-to-LTX + remote upload to complete)

Thanks for the original proposal — it helped drive the design.

PR #1094, opened 2026-02-09
feat(retention): add skip-remote-deletion for cloud-managed retention

### LINKED ISSUE #604: Let cloud provider handle retention instead of Litestream ?
Most cloud providers provide a feature of file retention built in with their object storage product. One of the positive aspects would be to remove some permissions from the IAM credentials provided to LiteStream (DeleteObject and UpdateObject), which I believe would enhance security in case of a credentials leak.

corylanou:
## Summary

- Adds `skip-remote-deletion` config option that disables Litestream's remote file deletion during retention enforcement
- Allows cloud provider lifecycle policies (S3 lifecycle rules, GCS object lifecycle, etc.) to handle file retention instead of Litestream
- Enhances security by removing the need for `DeleteObject` IAM permissions — if credentials leak, an attacker cannot delete backups
- Local file cleanup continues normally regardless of this setting
- Logs a startup warning about idle databases that may lose backup coverage

## Configuration

```yaml
skip-remote-deletion: true
```

## Implementation

Guards all 5 `DeleteLTXFiles()` call sites across `compactor.go` (3) and `db.go` (2) with `SkipRemoteDeletion` check. Follows the existing `VerifyCompaction` propagation pattern: `Store` → `DB` → `Compactor`.

## Test plan

- [x] 3 compactor-level tests verify remote files preserved while local cleanup runs
- [x] 2 DB-level tests verify end-to-end behavior with real SQLite databases
- [x] 1 store-level test verifies propagation to all DBs
- [x] Full test suite passes with race detector (`go test -race ./...`)
- [x] Pre-commit hooks pass (imports, vet, staticcheck)

Closes #604

### DIFF
--- a/db.go
+++ b/db.go
@@ -205,6 +205,7 @@ func NewDB(path string) *DB {
 		MonitorInterval:      DefaultMonitorInterval,
 		BusyTimeout:          DefaultBusyTimeout,
 		L0Retention:          DefaultL0Retention,
+		RetentionEnabled:     true,
 		ShutdownSyncTimeout:  DefaultShutdownSyncTimeout,
 		ShutdownSyncInterval: DefaultShutdownSyncInterval,
 		Logger:               slog.With("db", filepath.Base(path)),


store.go
```diff
@@ -435,6 +442,16 @@ func (s *Store) SetVerifyCompaction(v bool) {
 	}
 }
 
+func (s *Store) SetSkipRemoteDeletion(v bool) {
```
benbjohnson: Can you change the `SkipRemoteDeletion` name to `RetentionEnabled` to match the config?

corylanou: Fixed: Renamed `SkipRemoteDeletion` to `RetentionEnabled` across all files. This required inverting the boolean logic since:
- `SkipRemoteDeletion=true` (skip deletion) → `RetentionEnabled=false`
- `SkipRemoteDeletion=false` (do deletion) → `RetentionEnabled=true`

The field now defaults to `true` in `NewStore()` to maintain the same default behavior (retention enforcement enabled). Updated all usages in `store.go`, `db.go`, `compactor.go`, `replicate.go`, and their tests.

corylanou:
@codex review

PR #1102, opened 2026-02-11
feat(restore): add -f follow mode for continuous restore

### LINKED ISSUE #1065: feat(restore) Continuous restore from file replica to restored database file?
## Feature Description

I would like have multiple copies of the same SQLite file.

I thought I could replicate safely to other SQLite files by restoring multiple times repeatedly.

Those destination restores files are read-only.

I would configure litestream to replicate to a file replica and then somehow continuously replay that litestream wal to multiple SQLite files.

I am currently using the SQLite Online Backup  C API

I am also investigating using sqlite_rsync in the SQLite rsync branch in  SQLite.

## Use Cases

Can keep an SQLite file in sync with another SQLite database on disk, single writer many readers.

The exact problem I am trying to solve is how how to scale writing and reading independently

I assumed it would be faster than copying the entire database every time but more incremental

## Attempted Solutions

Currently I am using the SQLite Online Backup API.

Stop all programs running SQLite and use a standard `cp` of the database file.

## Proposal

Optionally when you run restore you can run it in continuous mode when the source is a file replica

It would be a faster incremental backup.

Alternatively a script that runs in a loop and involves the restore command to catch data.

I understand that file watching on cloud buckets is difficult and per host. But I feel the syncing a single database to multiple files is useful for scalability.

corylanou:
## Summary

- Adds a `-f` flag to the `restore` command that continuously polls for and applies new LTX files after the initial restore, similar to `tail -f`
- Enables maintaining read-only database replicas that stay in sync with the source via incremental LTX file application
- Falls back to higher compaction levels when level 0 files have been compacted away, handles database truncation (VACUUM), and exits cleanly on Ctrl+C

Closes #1065

## Implementation Details

**`db.go`**: Added `Follow` and `FollowInterval` fields to `RestoreOptions` with `DefaultFollowInterval` (1s).

**`replica.go`**: Added validation (follow is incompatible with `-txid`/`-timestamp`), and four new methods:
- `follow()` - Main poll loop that opens the restored DB, reads page size, and ticks at the configured interval
- `applyNewLTXFiles()` - Polls level 0 for new files, handles gaps via `fillFollowGap()`
- `applyLTXFile()` - Applies a single LTX file's pages to the database (follows `Hydrator.ApplyLTX` pattern)
- `fillFollowGap()` - Searches compaction levels 1-8 to bridge gaps when L0 files are compacted

**`cmd/litestream/restore.go`**: Added `-f` flag, signal handling for clean Ctrl+C shutdown, usage docs and example.

## Usage

```bash
# Continuously restore (follow) a database from a replica
litestream restore -f -o /tmp/read-replica.db s3://mybucket/db
```

**Important**: The restored database should only be opened in read-only mode by consumers.

## Test plan

- [x] `TestReplica_Restore_Follow_IncompatibleFlags` - Verifies error when Follow + TXID or Follow + Timestamp
- [x] `TestReplica_Restore_Follow` - Integration test: creates source DB, syncs to file replica, restores with follow, inserts new data, verifies follow applies it, cancels cleanly
- [x] `TestReplica_Restore_Follow_ContextCancellation` - Verifies immediate cancel returns nil (clean shutdown)
- [x] All existing tests pass with race detector
- [x] Pre-commit hooks pass (go-imports, go-vet, staticcheck)

### DIFF
--- a/replica.go
+++ b/replica.go
@@ -759,14 +759,12 @@ func (r *Replica) applyNewLTXFiles(ctx context.Context, f *os.File, afterTXID lt
 			continue
 		}
 
-		newPageSize, err := r.applyLTXFile(ctx, f, info, pageSize)
-		if err != nil {
+		if err := r.applyLTXFile(ctx, f, info, pageSize); err != nil {
 			return closeLevel0(fmt.Errorf(
 				"apply ltx file (level=%d, min=%s, max=%s): %w",
 				info.Level, info.MinTXID, info.MaxTXID, err,
 			))
 		}
-		pageSize = newPageSize
 		currentTXID = info.MaxTXID
 	}
 
@@ -796,37 +794,32 @@ func (r *Replica) applyNewLTXFiles(ctx context.Context, f *os.File, afterTXID lt
 // (bytes 18-19) to indicate DELETE journal mode instead of WAL mode, and
 // randomize the schema change counter (bytes 24-27) to invalidate cached
 // schemas in other connections.
-func (r *Replica) applyLTXFile(ctx context.Context, f *os.File, info *ltx.FileInfo, pageSize uint32) (uint32, error) {
+func (r *Replica) applyLTXFile(ctx context.Context, f *os.File, info *ltx.FileInfo, pageSize uint32) error {
 	rc, err := r.Client.OpenLTXFile(ctx, info.Level, info.MinTXID, info.MaxTXID, 0, 0)
 	if err != nil {
-		return pageSize, fmt.Errorf("open ltx file: %w", err)
+		return fmt.Errorf("open ltx file: %w", err)
 	}
 	defer rc.Close()
 
 	dec := ltx.NewDecoder(rc)
 	if err := dec.DecodeHeader(); err != nil {
-		return pageSize, fmt.Errorf("decode header: %w", err)
+		return fmt.Errorf("decode header: %w", err)
 	}
 
 	hdr := dec.Header()
 
-	ps := pageSize
-	if hdr.PageSize > 0 {
-		ps = hdr.PageSize
-	}
-
 	if err := internal.LockFileExclusive(f); err != nil {
-		return pageSize, fmt.Errorf("acquire exclusive lock: %w", err)
+		return fmt.Errorf("acquire exclusive lock: %w", err)
 	}
 	defer internal.UnlockFile(f)
 
 	for {
 		var phdr ltx.PageHeader
-		data := make([]byte, ps)
+		data := make([]byte, pageSize)
 		if err := dec.DecodePage(&phdr, data); err == io.EOF {
 			break
 		} else if err != nil {
-			return pageSize, fmt.Errorf("decode page: %w", err)
+			return fmt.Errorf("decode page: %w", err)
 		}
 
 		if phdr.Pgno == 1 && len(data) >= 28 {
@@ -834,24 +827,24 @@ func (r *Replica) applyLTXFile(ctx context.Context, f *os.File, info *ltx.FileIn
 			_, _ = rand.Read(data[24:28])
 		}
 
-		off := int64(phdr.Pgno-1) * int64(ps)
+		off := int64(phdr.Pgno-1) * int64(pageSize)
 		if _, err := f.WriteAt(data, off); err != nil {
-			return pageSize, fmt.Errorf("write page %d: %w", phdr.Pgno, err)
+			return fmt.Errorf("write page %d: %w", phdr.Pgno, err)
 		}
 	}
 
 	if hdr.Commit > 0 {
-		newSize := int64(hdr.Commit) * int64(ps)
+		newSize := int64(hdr.Commit) * int64(pageSize)
 		if err := f.Truncate(newSize); err != nil {
-			return pageSize, fmt.Errorf("truncate: %w", err)
+			return fmt.Errorf("truncate: %w", err)
 		}
 	}
 
 	if err := dec.Close(); err != nil {
-		return pageSize, fmt.Errorf("close decoder: %w", err)
+		return fmt.Errorf("close decoder: %w", err)
 	}
 
-	return ps, f.Sync()
+	return f.Sync()
 }
 
 // fillFollowGap attempts to bridge a gap in level 0 files by searching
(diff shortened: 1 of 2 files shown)


corylanou:
Design note on approach vs alternatives:

We intentionally kept this change as a targeted hardening of the new follow path instead of rewriting restore selection logic wholesale.

Why this was needed:
- One-shot restore already existed, but follow mode introduces a moving target.
- Compaction can produce overlapping files (e.g. 100-200) that still cover a follower cursor (e.g. 150), so seeking only from cursor+1 in higher levels can miss valid data.
- We also needed follow polling to treat iterator/list failures as real errors (not "no updates").
- Incremental apply now closes the decoder so trailer/file checksum validation is enforced before treating the update as successful.

What this patch does:
- Keeps existing L0-first behavior.
- Adds explicit higher-level gap-fill behavior that can see overlapping compacted files.
- Falls back to higher levels when L0 is empty.
- Surfaces iterator errors/close failures.
- Verifies checksum via dec.Close() in apply path.

Possible alternatives if we want a different long-term shape:
1. Unify selection into a single "next covering file" algorithm across levels:
   - pick file where MinTXID <= cursor+1 and MaxTXID > cursor,
   - choose the candidate that advances furthest,
   - repeat until no progress.
   This can remove most special-case gap-fill logic.
2. Persist follower state as TXID+checksum (not TXID only) so restart/reconciliation can detect divergence earlier and resume more deterministically.
3. Add an explicit abstraction for follow planning (like CalcRestorePlan but incremental), so rules are easier to reason about and test independently.

replica.go
```diff
@@ -626,7 +631,261 @@ func (r *Replica) Restore(ctx context.Context, opt RestoreOptions) (err error) {
...
+		}
+		currentTXID = bridgedTXID
+	}
+
+	return currentTXID, nil
+}
+
+// applyLTXFile applies a single LTX file's pages to the database file.
+// This follows the same pattern as Hydrator.ApplyLTX (vfs.go:712-747).
+func (r *Replica) applyLTXFile(ctx context.Context, f *os.File, info *ltx.FileInfo, pageSize uint32) error {
```
benbjohnson: Since we're applying LTX files to live databases, we need to add in locking as well. The simplest approach is to copy what we're doing in the Litestream VFS and rewrite the [header in the first page in SQLite](https://www.sqlite.org/fileformat.html) so that it turns off WAL mode and uses the rollback journal instead. It's set on bytes 18 & 19. You'll need to update that every time you write a page 1 to disk from LTX.

Once that's being set, you'll need to acquire and release locks to get yourself a write lock in rollback journal mode. That will prevent other SQLite processes from acquiring readers & writers while you're updating the database. We could implement locks on the WAL but that gets more complicated and I don't think it's worth it yet.

benbjohnson: One other thing we should consider later is that writing to the database directly could corrupt it if Litestream dies. There are a couple of options since it's just a read replica:

1. Require that the database be deleted by the user before restarting "follow" so that we can start from a fresh snapshot.

2. Write a file like `my.db-txid` that saves the last successful TXID written so we can recover when a following restore is run again. This mainly helps for large databases.

corylanou: Done. Added locking and header rewriting:

1. **Exclusive file locking**: Added platform-specific locking functions (`internal/lock_unix.go` and `internal/lock_windows.go`) using `fcntl` byte-range locks at SQLite's PENDING_BYTE / SHARED range (Unix) and `LockFileEx` at the same offsets (Windows). The exclusive lock is acquired before writing pages and released after sync.

2. **WAL mode disable**: When writing page 1, bytes 18-19 are set to `0x01, 0x01` to indicate DELETE (rollback) journal mode instead of WAL mode. This matches the pattern in `vfs.go:759-763`.

3. **Schema change counter**: Bytes 24-27 are randomized on each page 1 write to invalidate cached schemas in other connections.

The crash recovery txid file is tracked as #1131 for a follow-up.

corylanou: Created #1131 to track the crash recovery txid file as a follow-up.

replica.go
```diff
@@ -787,58 +796,62 @@ func (r *Replica) applyNewLTXFiles(ctx context.Context, f *os.File, afterTXID lt
 // (bytes 18-19) to indicate DELETE journal mode instead of WAL mode, and
 // randomize the schema change counter (bytes 24-27) to invalidate cached
 // schemas in other connections.
-func (r *Replica) applyLTXFile(ctx context.Context, f *os.File, info *ltx.FileInfo, pageSize uint32) error {
+func (r *Replica) applyLTXFile(ctx context.Context, f *os.File, info *ltx.FileInfo, pageSize uint32) (uint32, error) {
 	rc, err := r.Client.OpenLTXFile(ctx, info.Level, info.MinTXID, info.MaxTXID, 0, 0)
 	if err != nil {
-		return fmt.Errorf("open ltx file: %w", err)
+		return pageSize, fmt.Errorf("open ltx file: %w", err)
```
benbjohnson: The page size of a database can't ever change.

corylanou: Good point - reverted the page size handling. The function now returns just `error` again and uses the passed-in `pageSize` directly.

Will do manual testing of the locking behavior next.

corylanou:
## Manual Locking Test Results

Tested the SQLite-compatible file locking (fcntl byte-range locks at PENDING_BYTE / SHARED range) with concurrent readers while follow mode is actively writing pages.

### Test 1: Concurrent Reads During Follow Mode
- Started `litestream replicate` → file replica
- Started `litestream restore -f` (follow mode) to continuously apply LTX files
- Ran 20 concurrent `sqlite3` reader queries (500ms apart) while inserting 15 rows to source
- **Result: 20/20 reads succeeded, 0 errors**
- Row counts progressed monotonically: 13 → 17 → 20 → 23 → 26 → 28
- No partial writes visible to readers

### Test 2: Lock Contention Stress Test
- 50 rapid inserts to source (50ms apart) generating frequent follower updates
- 50 rapid reads from restored DB in parallel (50ms apart) with `PRAGMA busy_timeout=10000`
- **Result: 50/50 reads succeeded, 0 busy retries, 0 errors**
- Source DB count: 78, Restored DB count: 78 (exact match after catch-up)
- No database corruption detected

### Summary
The exclusive lock acquired during page writes properly coordinates with SQLite's built-in locking. Readers using rollback journal mode (bytes 18-19 set to DELETE mode) correctly wait for or work around the exclusive lock. The schema change counter randomization (bytes 24-27) ensures readers see fresh data.

PR #1105, opened 2026-02-12
fix(vfs): allow pooled opens in write mode

### LINKED ISSUE #1019: VFS expectedTXID not updated after successful sync causes false conflict errors

corylanou:
## Summary

Fixes #1019. After discussion with @benbjohnson, this keeps **single-writer enforcement** while avoiding the pooled-connection regression from global open-time gating.

The regression came from rejecting every second `OpenMainDB` in write mode. In practice, `database/sql` can open additional handles for concurrent reads, so this caused intermittent `SQLITE_BUSY` / `database is locked` failures in normal read/write workloads.

The final approach enforces contention at the **write lock boundary** instead of at open:
- allow multiple `OpenMainDB` connections in write mode
- enforce single writer when a connection attempts `RESERVED` lock

### Changes

- **`VFS` / `VFSFile` write gating**
- remove global rejection of additional `OpenMainDB` calls in write mode
- enforce single active writer on `LockReserved` by returning `sqlite3vfs.BusyError` when another writer is active
- release writer slot when lock drops below `LockReserved`

- **Write buffer handling for pooled opens**
- assign per-open write buffer paths (`writeSeq`) so multiple open handles do not collide on one local buffer file

- **Tests (`vfs_write_test.go`)**
- replace second-open rejection expectations with pooled-open success expectations
- add lock-level contention coverage to verify second writer still gets `BusyError`
- keep regression coverage aligned with concurrent read/write behavior (`TestVFS_ReadWhileWriting`)

### Concurrency Safety

- writer ownership is checked/assigned under mutex at `RESERVED` lock acquisition
- non-writer pooled connections are no longer rejected at open time
- writer slot is released when transaction lock level drops below `RESERVED`
- single-writer semantics are preserved where SQLite expects contention to occur

## Test plan

- [x] `TestVFS_OpenAllowsMultipleConnectionsWhenWriteEnabled`
- [x] `TestVFS_ConcurrentOpenAllSucceed`
- [x] `TestVFS_WriteLockBlocksConcurrentWriters`
- [x] `TestVFS_ReadWhileWriting`
- [x] `go test -tags vfs . -count=1`

## Notes

This PR carries forward the original #1019 investigation context and updates only the final technical solution to the lock-scoped writer gate.

### DIFF
--- a/docs/VFS.md
+++ b/docs/VFS.md
@@ -283,12 +283,13 @@ export LITESTREAM_SYNC_INTERVAL="1s"
 2. Dirty pages are tracked in memory
 3. Periodically (or on close), dirty pages are packaged into an LTX file
 4. The LTX file is uploaded to remote storage
-5. Conflict detection prevents overwrites if the remote has newer transactions
+5. If the remote has advanced unexpectedly, `ErrConflict` is returned as a safety check
 
 ### Write Mode Considerations
 
-- **Single writer**: Only one writer should be active at a time
-- **Conflict detection**: If another writer has updated the remote, `ErrConflict` is returned
+- **Connection pooling**: Multiple connections can be opened in write mode (for example, by `database/sql`)
+- **Single writer**: Write contention is enforced at lock acquisition. If another connection already holds write intent, SQLite returns `SQLITE_BUSY`
+- **Conflict detection**: If the remote has advanced unexpectedly, `ErrConflict` is returned
 - **Buffer durability**: The local buffer file provides crash recovery for uncommitted writes
 - **Sync interval**: Balance between durability (shorter) and performance (longer)
 - **New databases**: Write mode can create new databases from scratch if no LTX files exist


corylanou:
Closing in favor of a clean reimplementation on branch `fix/1019-vfs-pooled-write-coordination`. This PR was built on a commit from PR #1022 that Ben rejected ("This is the wrong approach"). The new branch takes a simpler approach: VFS-level `lastSyncedTXID` propagation + RESERVED lock gating, without the rejected `vfsWriteState` reference counting.

PR #1126, opened 2026-02-16
feat(docker): add hardened scratch-based image variant

### LINKED ISSUE #1124: Hardened Image for litestream
## Feature Description
Would it be possible to provide an alternative hardened OCI image for litestream (e.g: build on strach or distroless)?
This would help reduce the attack surface.

## Use Cases

When working on a regulated environment, you must perform regular/autonomous CVEs scans.
During those scans alerts can popup for CVEs can pop due to vuln in the litestream binary other can pop due to vuln in debian:bookworm-slim's lib which could be avoid by using a minimal image as noted below.
More over those image comes only with the strict necessary which increase even further the security (e.g: no shell, no curl/wget...)

## Attempted Solutions

One could rebuild it's own image by adding the binary into.
This poses extra (yet minimal) toil and this does not benefit to the opensource community

## Proposal

Adjust the dockerfile to use strach.

```
FROM golang:1.24 AS builder

# Install build dependencies for VFS extension
RUN apt-get update && apt-get install -y gcc libc6-dev && rm -rf /var/lib/apt/lists/*

WORKDIR /src/litestream
COPY . .

ARG LITESTREAM_VERSION=latest

# Build litestream binary
RUN --mount=type=cache,target=/root/.cache/go-build \
	--mount=type=cache,target=/go/pkg \
	go build -ldflags "-s -w -X 'main.Version=${LITESTREAM_VERSION}' -extldflags '-static'" -tags osusergo,netgo,sqlite_omit_load_extension -o /usr/local/bin/litestream ./cmd/litestream

# Build VFS loadable extension
RUN --mount=type=cache,target=/root/.cache/go-build \
	--mount=type=cache,target=/go/pkg \
	mkdir -p dist && \
	CGO_ENABLED=1 go build \
	-tags "vfs,SQLITE3VFS_LOADABLE_EXT" \
	-buildmode=c-archive \
	-o dist/litestream-vfs.a ./cmd/litestream-vfs && \
	mv dist/litestream-vfs.h src/litestream-vfs.h && \
	gcc -DSQLITE3VFS_LOADABLE_EXT -g -fPIC -shared \
	-o dist/litestream-vfs.so \
	src/litestream-vfs.c \
	dist/litestream-vfs.a \
	-lpthread -ldl -lm

FROM alpine:latest as certs
RUN apk --update add ca-certificates

FROM scratch

COPY --from=certs /etc/ssl/certs/ca-certificates.crt /etc/ssl/certs/ca-certificates.crt
COPY --from=builder /usr/local/bin/litestream /usr/local/bin/litestream
COPY --from=builder /src/litestream/dist/litestream-vfs.so /usr/local/lib/litestream-vfs.so

ENTRYPOINT ["/usr/local/bin/litestream"]
CMD []
```

corylanou:
## Summary

Closes #1124

- Adds a `scratch`-based Docker image target for environments requiring minimal attack surface (no shell, no extra libs, no CVE surface from base image)
- Publishes both Debian (default) and scratch variants from the same workflow with `-scratch` tag suffixes (e.g., `0.5.9-scratch`)
- Existing users are unaffected — default tags (`latest`, `0.5.9`, `0.5`) continue to produce the Debian image

### Additional fixes
- Pin Alpine to `3.21` for reproducible cert extraction
- Add `flavor: latest=false` on scratch metadata to prevent overwriting the Debian `:latest` tag
- Use `release.tag_name` instead of `release.name` for `LITESTREAM_VERSION` (avoids spaces breaking ldflags)
- Omit VFS `.so` from scratch image (requires glibc dynamic loader, non-functional in scratch)
- Keep `default` (Debian) as last Dockerfile stage so plain `docker build .` still works

### Tag strategy

| Variant | Tag Examples |
|---------|-------------|
| Default (Debian) | `latest`, `0.5.9`, `0.5` |
| Hardened (Scratch) | `latest-scratch`, `0.5.9-scratch`, `0.5-scratch` |

## Test plan

- [ ] Build both targets locally: `docker build --target default -t litestream:default .` and `docker build --target scratch -t litestream:scratch .`
- [ ] Verify scratch image runs: `docker run --rm litestream:scratch version`
- [ ] Verify plain `docker build .` still produces the Debian image
- [ ] Compare image sizes: `docker images litestream`
- [ ] Verify scratch image has minimal layers: `docker inspect litestream:scratch | jq '.[0].RootFS.Layers | length'`

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -31,10 +31,8 @@ RUN --mount=type=cache,target=/root/.cache/go-build \
 # --- Hardened image (Scratch) ---
 FROM alpine:3.21 AS certs
 RUN apk --update add ca-certificates && \
-	echo "root:x:0:0:root:/root:/sbin/nologin" > /etc/minimal-passwd && \
-	echo "nonroot:x:65532:65532:nonroot:/home/nonroot:/sbin/nologin" >> /etc/minimal-passwd && \
-	echo "root:x:0:" > /etc/minimal-group && \
-	echo "nonroot:x:65532:" >> /etc/minimal-group
+	echo "nonroot:x:65532:65532:nonroot:/home/nonroot:/sbin/nologin" > /etc/minimal-passwd && \
+	echo "nonroot:x:65532:" > /etc/minimal-group
 
 FROM scratch AS hardened
 COPY --from=certs /etc/ssl/certs/ca-certificates.crt /etc/ssl/certs/ca-certificates.crt


Dockerfile
```diff
@@ -28,7 +28,22 @@ RUN --mount=type=cache,target=/root/.cache/go-build \
 	dist/litestream-vfs.a \
 	-lpthread -ldl -lm
 
-FROM debian:bookworm-slim
+# --- Hardened image (Scratch) ---
+FROM alpine:3.21 AS certs
+RUN apk --update add ca-certificates && \
+	echo "root:x:0:0:root:/root:/sbin/nologin" > /etc/minimal-passwd && \
```
wilfriedroset: Even with the use of scratch, we are aiming for a tightly secured deployment. Should litestream run as an unprivileged user?

Dockerfile
```diff
@@ -28,7 +28,22 @@ RUN --mount=type=cache,target=/root/.cache/go-build \
...
+	echo "root:x:0:" > /etc/minimal-group
+
+FROM scratch AS hardened
+COPY --from=certs /etc/ssl/certs/ca-certificates.crt /etc/ssl/certs/ca-certificates.crt
+COPY --from=certs /etc/minimal-passwd /etc/passwd
+COPY --from=certs /etc/minimal-group /etc/group
+COPY --from=builder /usr/local/bin/litestream /usr/local/bin/litestream
+ENTRYPOINT ["/usr/local/bin/litestream"]
```
wilfriedroset, line 43: Similarly to the comment above, should we add an instruction USER to switch to an unprivileged user (e.g: nobody) ?

Dockerfile
corylanou: Good point! Added `USER nonroot:nonroot` (UID 65532) to the hardened image. This follows the distroless convention and aligns with security best practices for scratch-based containers.

The `/etc/passwd` and `/etc/group` files now include both root and nonroot entries to support the USER instruction.

Dockerfile
corylanou: For the default (Debian) image, I opted not to add USER for backward compatibility reasons. The default image is used by existing deployments that may mount SQLite databases as root. Changing to non-root would break volume permissions for those users.

The hardened image (scratch) targets greenfield deployments in regulated environments where non-root is expected. Users choosing `-scratch` tags are explicitly opting into stricter security, so breaking changes are acceptable there.

If there's consensus to also harden the default image, it could be done in a separate PR with proper migration guidance.

Dockerfile
```diff
@@ -28,7 +28,25 @@ RUN --mount=type=cache,target=/root/.cache/go-build \
 	dist/litestream-vfs.a \
 	-lpthread -ldl -lm
 
-FROM debian:bookworm-slim
+# --- Hardened image (Scratch) ---
+FROM alpine:3.21 AS certs
+RUN apk --update add ca-certificates && \
+	echo "root:x:0:0:root:/root:/sbin/nologin" > /etc/minimal-passwd && \
+	echo "nonroot:x:65532:65532:nonroot:/home/nonroot:/sbin/nologin" >> /etc/minimal-passwd && \
+	echo "root:x:0:" > /etc/minimal-group && \
```
wilfriedroset: Do we need to keep the root group ?

Dockerfile
```diff
@@ -28,7 +28,25 @@ RUN --mount=type=cache,target=/root/.cache/go-build \
 	dist/litestream-vfs.a \
 	-lpthread -ldl -lm
 
-FROM debian:bookworm-slim
+# --- Hardened image (Scratch) ---
+FROM alpine:3.21 AS certs
+RUN apk --update add ca-certificates && \
+	echo "root:x:0:0:root:/root:/sbin/nologin" > /etc/minimal-passwd && \
```
wilfriedroset: Do we need to keep the user root ?

corylanou: Good catch\! Removed root user entry. Since the container runs entirely as nonroot (UID 65532), root is unnecessary. This further reduces attack surface by not having a root identity defined at all.

Dockerfile
corylanou: Removed root group as well. The hardened image now only defines the nonroot user/group, aligning with minimal attack surface principles.

wilfriedroset:
What is the drawback of omiting VFS .so?

@corylanou is there a way to automate the test the resulting scratch image so that we can catch the issues before they reach users?

corylanou:
@wilfriedroset Great questions!

**VFS .so drawback:** The VFS extension is a loadable SQLite extension that lets an existing SQLite process load Litestream's replication at runtime (via `load_extension()`). It's omitted from the scratch image because it's built with CGO and links against glibc — scratch has no dynamic linker, so it would fail to load. Users of the scratch image can still use the standard `litestream replicate` sidecar/wrapper pattern, which is the most common usage. Minor trade-off in practice.

**Automated testing:** Great idea — created #1130 to add CI smoke tests for the scratch image (build verification, binary execution, non-root user check, CA certs, etc.). We'll tackle that as a follow-up effort so we can get this PR merged and maintain momentum.

PR #1127, opened 2026-02-16
test(metrics): add coverage for all Prometheus metrics

corylanou:
## Summary
- Fix `TestDB_Sync_ErrorMetrics` which always skipped — it injected errors via `errorReplicaClient.WriteLTXFile`, but `db.Sync()` never calls `WriteLTXFile` (that happens in `replica.Sync()`)
- Add tests for all previously untested Prometheus metrics:
  - `litestream_checkpoint_error_count` — trigger via closed db connection
  - `litestream_l0_retention_files_total` — verify gauges set during retention enforcement
  - `litestream_replica_operation_errors_total` — verify error counter increments by label
- Enhance `testReplicaClient` to actually list and open stored LTX files (previously returned empty iterators), enabling tests that need file iteration
- Remove unused `errorReplicaClient` type

Discovered while reviewing #1123 — the named-return metrics pattern is correct per the Go spec, but the test suite wasn't actually validating any error metrics.

## Test plan
- [ ] `go test -run "TestDB_Sync_ErrorMetrics|TestDB_Checkpoint_ErrorMetrics|TestDB_L0RetentionMetrics" -race -v ./` — all pass (sync test previously always skipped)
- [ ] `go test -run "TestReplicaOperationErrorMetrics" -race -v ./internal/` — passes
- [ ] `go test -race ./...` — full suite passes


internal/internal_test.go
```diff
@@ -49,3 +49,23 @@ func TestReplicaOperationMetrics(t *testing.T) {
...
+
+		OperationErrorCounterVec.WithLabelValues(testType, "DELETE", code).Inc()
+
+		after := testutil.ToFloat64(
+			OperationErrorCounterVec.WithLabelValues(testType, "DELETE", code))
+		if after != baseline+1 {
+			t.Fatalf("litestream_replica_operation_errors_total[%s]=%v, want %v", code, after, baseline+1)
+		}
+	}
+}
```
benbjohnson: This test seems like it's just testing Prometheus internals and not Litestream.

corylanou: Agreed — deleted that file entirely. The `testReplicaClient` mock now instruments operation metrics (PUT, GET, LIST, DELETE) the same way real clients do, and a new `TestDB_ReplicaSync_OperationMetrics` test verifies them through the full `Replica.Sync()` → `uploadLTXFile()` → `WriteLTXFile()` path.

PR #1132, opened 2026-02-19
test(ci): add smoke tests for hardened scratch Docker image

### LINKED ISSUE #1130: test: add CI smoke tests for hardened scratch Docker image
## Summary

Add automated CI smoke tests for the hardened scratch-based Docker image to catch issues before they reach users.

Ref: PR #1126 (comment: https://github.com/benbjohnson/litestream/pull/1126#issuecomment-3925785300)

## Motivation

The scratch image omits the VFS `.so` extension and runs as a non-root user, which are intentional constraints. However, there's currently no automated verification that the scratch image works correctly after building. CI smoke tests would catch regressions early.

## Proposed Tests

- Build both `default` and `hardened` targets in CI
- Run `docker run --rm litestream:scratch version` to verify the binary executes
- Verify the container runs as non-root (UID 65532)
- Verify no shell exists in the scratch image
- Verify CA certificates are present (TLS connections work)
- Consider using Google's `container-structure-test` for declarative checks

## Acceptance Criteria

- [ ] CI job builds the scratch image and runs smoke tests
- [ ] Tests fail if the binary doesn't execute in the scratch environment
- [ ] Tests verify non-root user and minimal filesystem

corylanou:
## Summary

- Replaces the `docker-build-test` job with `docker-smoke-test` in the CI commit workflow
- Builds **both** Docker targets: `default` (Debian) and `hardened` (scratch)
- Adds smoke tests for the hardened image: version command, non-root user (`nonroot:nonroot`), no shell, and CA certificates present
- Uses only standard `docker` commands (no external test tools) to keep CI dependencies minimal

Fixes #1130

## Test Plan

- [ ] CI `docker-smoke-test` job passes on this PR
- [ ] Default image builds and `version` command succeeds
- [ ] Hardened image builds and `version` command succeeds
- [ ] Hardened image user check validates `nonroot:nonroot`
- [ ] Hardened image confirms no `/bin/sh` shell exists
- [ ] Hardened image confirms CA certificates are present

### DIFF
--- a/.github/workflows/commit.yml
+++ b/.github/workflows/commit.yml
@@ -119,6 +119,14 @@ jobs:
             echo "FAIL: Expected user 'nonroot:nonroot', got '$user'"
             exit 1
           fi
+          docker create --name uid-check litestream:hardened version
+          uid=$(docker export uid-check | tar -xf - --to-stdout etc/passwd | grep '^nonroot:' | cut -d: -f3)
+          docker rm uid-check
+          echo "nonroot UID: $uid"
+          if [ "$uid" != "65532" ]; then
+            echo "FAIL: Expected UID 65532, got '$uid'"
+            exit 1
+          fi
 
       - name: "Hardened: verify no shell exists"
         run: |


.github/workflows/commit.yml
```diff
@@ -80,18 +80,68 @@ jobs:
...
           push: false
           load: true
-          tags: litestream:test
+          tags: litestream:hardened
+
+      - name: "Default: verify version command"
+        run: docker run --rm litestream:default version
+
+      - name: "Hardened: verify version command"
+        run: docker run --rm litestream:hardened version
```
wilfriedroset, line 112: Is this enough to make sure litestream will work as expected?

corylanou: The \`version\` command verifies the statically-linked binary actually executes in the scratch environment (catches missing libraries or broken builds). But it's not the only test — see the subsequent steps that also verify:

1. **Non-root user (UID 65532)** — confirms the container runs as `nonroot:nonroot` and the passwd file maps to UID 65532
2. **No shell exists** — confirms the scratch image has no `/bin/sh` (minimal attack surface)
3. **CA certificates present** — confirms TLS connections will work

Together these catch the most common scratch image regressions. For full functional testing (replication to S3/GCS/etc.), we rely on the existing integration tests which run against the Go binary directly.

PR #1136, opened 2026-02-20
feat(restore): add crash recovery txid file for follow mode

### LINKED ISSUE #1131: feat(restore): add crash recovery txid file for follow mode
## Summary

When using `restore -f` (follow mode), writing to the database directly could corrupt it if Litestream dies mid-write. A recovery mechanism is needed so follow mode can resume without requiring a full re-restore.

## Context

From @benbjohnson's review on #1102:

> One other thing we should consider later is that writing to the database directly could corrupt it if Litestream dies. There are a couple of options since it's just a read replica:
>
> 1. Require that the database be deleted by the user before restarting "follow" so that we can start from a fresh snapshot.
>
> 2. Write a file like `my.db-txid` that saves the last successful TXID written so we can recover when a following restore is run again. This mainly helps for large databases.

## Proposal

Write a sidecar file (e.g., `my.db-txid`) that records the last successfully applied TXID after each LTX file is fully written and synced. On restart, `restore -f` can read this file and resume from that point instead of requiring a full restore from snapshot.

This is especially important for large databases where a full re-restore would be expensive.

## References

- PR #1102 (follow mode implementation)
- Issue #1065 (original feature request)

corylanou:
## Description

Adds a sidecar `.txid` file for `restore -f` (follow mode) that enables crash recovery. The file records the last successfully applied TXID after each batch of LTX files using an atomic temp-file + fsync + rename pattern. On restart, follow mode reads this file and resumes from the saved TXID instead of requiring a full re-restore.

Key changes:
- `WriteTXIDFile` / `ReadTXIDFile` helpers with atomic write semantics
- `Restore()` checks for existing DB + `.txid` file to resume follow mode
- `follow()` writes TXID file after each successful batch apply
- Initial TXID file written after first restore before entering follow loop
- Error when DB exists but no `.txid` file (tells user to delete and re-restore)
- Stale TXID validation: detects when saved TXID is behind the earliest snapshot (retention pruned history) and returns a clear error instead of silently stalling
- `follow()` only advances in-memory `lastTXID` when `WriteTXIDFile` succeeds, keeping the durable checkpoint consistent

## Motivation and Context

When using `restore -f` (follow mode), Litestream writes pages directly to the database file. If the process dies mid-write, the database could be corrupted. Previously the only recovery option was to delete the database and re-restore from scratch — expensive for large databases.

Fixes #1131

## How Has This Been Tested?

- `go build ./...` — compiles cleanly
- `go test -race ./...` — all tests pass (including new and existing)

New tests added:
- `TestWriteTXIDFile` — correct hex format, atomic overwrite
- `TestReadTXIDFile` — missing file returns 0, valid file, malformed file errors, empty file errors
- `TestReplica_Restore_Follow_WriteTXIDFile` — verifies `.txid` created after initial restore and updated after follow applies
- `TestReplica_Restore_Follow_CrashRecovery` — full crash recovery: start follow, stop, add data, restart, verify resume
- `TestReplica_Restore_Follow_NoTXIDFile` — verifies error when DB exists but no `.txid` file
- `TestReplica_Restore_Follow_StaleTXID` — verifies error when saved TXID is behind pruned replica history

## Types of changes

## Checklist

### DIFF
--- a/replica.go
+++ b/replica.go
@@ -547,27 +547,30 @@ func (r *Replica) Restore(ctx context.Context, opt RestoreOptions) (err error) {
 			// and we can't catch up incrementally.
 			snapshotItr, itrErr := r.Client.LTXFiles(ctx, SnapshotLevel, 0, false)
 			if itrErr != nil {
-				// Can't validate saved TXID; delete DB and txid file, then do full restore.
-				r.Logger().Warn("cannot validate saved TXID, performing full restore", "err", itrErr)
-				if err := os.Remove(opt.OutputPath); err != nil {
-					return fmt.Errorf("remove database for re-restore: %w", err)
-				}
-				if err := os.Remove(TXIDPath(opt.OutputPath)); err != nil && !os.IsNotExist(err) {
-					return fmt.Errorf("remove txid file for re-restore: %w", err)
-				}
-			} else {
-				if snapshotItr.Next() {
-					info := snapshotItr.Item()
-					if info.MinTXID > txid {
-						_ = snapshotItr.Close()
-						return fmt.Errorf("cannot resume follow mode: saved TXID %s is behind the earliest snapshot (min TXID %s); replica history has been pruned — delete %s and %s-txid to re-restore", txid, info.MinTXID, opt.OutputPath, opt.OutputPath)
-					}
-				}
-				_ = snapshotItr.Close()
-
-				r.Logger().Info("resuming follow mode from crash recovery", "txid", txid, "output", opt.OutputPath)
-				return r.follow(ctx, opt.OutputPath, txid, opt.FollowInterval)
+				return fmt.Errorf("cannot validate saved TXID for crash recovery: %w", itrErr)
 			}
+
+			var latestSnapshot *ltx.FileInfo
+			for snapshotItr.Next() {
+				latestSnapshot = snapshotItr.Item()
+			}
+			if err := snapshotItr.Err(); err != nil {
+				_ = snapshotItr.Close()
+				return fmt.Errorf("iterate snapshots for crash recovery validation: %w", err)
+			}
+			_ = snapshotItr.Close()
+
+			if latestSnapshot != nil {
+				if latestSnapshot.MinTXID > txid {
+					return fmt.Errorf("cannot resume follow mode: saved TXID %s is behind the earliest snapshot (min TXID %s); replica history has been pruned -- delete %s and %s-txid to re-restore", txid, latestSnapshot.MinTXID, opt.OutputPath, opt.OutputPath)
+				}
+				if txid > latestSnapshot.MaxTXID {
+					return fmt.Errorf("cannot resume follow mode: saved TXID %s is ahead of latest snapshot (max TXID %s); delete %s and %s-txid to re-restore", txid, latestSnapshot.MaxTXID, opt.OutputPath, opt.OutputPath)
+				}
+			}
+
+			r.Logger().Info("resuming follow mode from crash recovery", "txid", txid, "output", opt.OutputPath)
+			return r.follow(ctx, opt.OutputPath, txid, opt.FollowInterval)
 		}
 	}
 
@@ -1545,7 +1548,15 @@ func WriteTXIDFile(outputPath string, txid ltx.TXID) error {
 		return fmt.Errorf("rename txid file: %w", err)
 	}
 
-	return nil
+	dir, err := os.Open(filepath.Dir(txidPath))
+	if err != nil {
+		return fmt.Errorf("open txid dir for sync: %w", err)
+	}
+	if err := dir.Sync(); err != nil {
+		_ = dir.Close()
+		return fmt.Errorf("sync txid dir: %w", err)
+	}
+	return dir.Close()
 }
 
 // ReadTXIDFile reads the TXID from a sidecar file at <outputPath>-txid.


replica.go
```diff
@@ -530,6 +531,37 @@ func (r *Replica) Restore(ctx context.Context, opt RestoreOptions) (err error) {
...
+	// In follow mode, if the database already exists, attempt crash recovery
+	// by reading the last applied TXID from the sidecar file.
+	if opt.Follow {
+		if _, statErr := os.Stat(opt.OutputPath); statErr == nil {
+			txid, readErr := ReadTXIDFile(opt.OutputPath)
+			if readErr != nil {
+				return fmt.Errorf("read txid file for crash recovery: %w", readErr)
+			}
+			if txid == 0 {
+				return fmt.Errorf("cannot resume follow mode: database exists at %s but no .txid file found; delete the database to re-restore", opt.OutputPath)
```
benbjohnson: Can you move the path to the end? If the path is long then it can be difficult to read. It's also hard to pattern match on it if users wanted to do that. e.g.

```go
return fmt.Errorf("cannot resume follow mode: database exists but no .txid file found; delete the database to re-restore: %s", opt.OutputPath)
```

replica.go
```diff
@@ -530,6 +531,37 @@ func (r *Replica) Restore(ctx context.Context, opt RestoreOptions) (err error) {
...
+			if itrErr == nil {
+				if snapshotItr.Next() {
+					info := snapshotItr.Item()
+					if info.MinTXID > txid {
+						_ = snapshotItr.Close()
+						return fmt.Errorf("cannot resume follow mode: saved TXID %s is behind the earliest snapshot (min TXID %s); replica history has been pruned — delete %s and %s.txid to re-restore", txid, info.MinTXID, opt.OutputPath, opt.OutputPath)
+					}
+				}
+				_ = snapshotItr.Close()
+			}
```
benbjohnson: This doesn't need to be an error. We should just fetch a full snapshot and write it to the file.

replica.go
```diff
@@ -646,6 +678,9 @@ func (r *Replica) Restore(ctx context.Context, opt RestoreOptions) (err error) {
 		rdrs = nil
 
 		maxTXID := infos[len(infos)-1].MaxTXID
+		if err := WriteTXIDFile(opt.OutputPath, maxTXID); err != nil {
+			r.Logger().Error("follow: failed to write initial txid file", "err", err)
+		}
```
benbjohnson, line 695: This should actually return an error rather than just log.

replica.go
```diff
@@ -702,6 +737,10 @@ func (r *Replica) follow(ctx context.Context, outputPath string, lastTXID ltx.TX
 				continue
 			}
 			if newTXID > lastTXID {
+				if err := WriteTXIDFile(outputPath, newTXID); err != nil {
+					r.Logger().Error("follow: failed to write txid file, not advancing checkpoint", "err", err)
+					continue
+				}
```
benbjohnson, line 754: This should return an error.

replica.go
```diff
@@ -1463,6 +1502,62 @@ func restoreCandidateBetter(curr, next *ltx.FileInfo) bool {
...
+// WriteTXIDFile atomically writes a TXID to a sidecar file at <outputPath>.txid.
+// Uses temp-file + fsync + rename for crash safety.
+func WriteTXIDFile(outputPath string, txid ltx.TXID) error {
+	txidPath := outputPath + ".txid"
+	tmpPath := txidPath + ".tmp"
+
+	f, err := os.Create(tmpPath)
+	if err != nil {
+		return fmt.Errorf("create txid temp file: %w", err)
+	}
```
benbjohnson, line 1531: I'd prefer the `f.Close()` and `os.Remove()` be deferred her rather than performed in the blocks below. It's cleaner and it also ensures they still happen even if there's a panic that occurs.

replica.go
```diff
@@ -1463,6 +1502,62 @@ func restoreCandidateBetter(curr, next *ltx.FileInfo) bool {
 	return next.CreatedAt.Before(curr.CreatedAt)
 }
 
+// WriteTXIDFile atomically writes a TXID to a sidecar file at <outputPath>.txid.
+// Uses temp-file + fsync + rename for crash safety.
+func WriteTXIDFile(outputPath string, txid ltx.TXID) error {
+	txidPath := outputPath + ".txid"
```
benbjohnson: Can you change the suffix to be `-txid` rather than `.txid`? It's more in line with how SQLite names associated files (e.g. `-wal` and `-shm`)

replica.go
corylanou: Fixed: Moved path to end of error message for better readability and pattern matching.

replica.go
corylanou: Fixed: Changed behavior to delete DB and txid file, then fall through to full snapshot restore when iterator fails. Added log warning to explain the recovery action.

replica.go
corylanou: Fixed: Now returns error instead of logging.

replica.go
corylanou: Fixed: Now returns error instead of logging.

replica.go
corylanou: Fixed: Added `defer f.Close()` and `defer os.Remove(tmpPath)` after file creation for cleaner cleanup.

replica.go
corylanou: Fixed: Changed suffix from `.txid` to `-txid` to match SQLite naming conventions. Also added a `TXIDPath()` helper function for consistent path generation.

PR #1157, opened 2026-02-23
feat(vfs): thread-safe config registry for per-database VFS configuration

### LINKED ISSUE #1150: feat(vfs): support thread-safe configuration via URI parameters and PRAGMAs (no setenv required)
## Problem Statement

The VFS loadable extension currently requires all configuration via environment variables (`LITESTREAM_REPLICA_URL`, `LITESTREAM_WRITE_ENABLED`, etc.) which must be set **before** VFS registration. On Linux, `setenv()` is not thread-safe — concurrent `getenv()` calls (from DNS resolution, glibc internals, other libraries) can read freed memory and crash. This makes Litestream VFS unusable as a library dependency when config isn't known at process start.

References:
- https://www.evanjones.ca/setenv-is-not-thread-safe.html
- Go's `os.Setenv` wraps C `setenv()` and inherits the same unsafety
- Python's `os.environ` mutations are similarly unsafe

Related: #1134

### Approach 1: URI Query Parameters (per-connection config at open time)

SQLite already supports URI filenames and passes the full URI to `xOpen()`. The `psanford/sqlite3vfs` bridge passes this `name` string through to `VFS.Open()`. When `OpenURI` flag is set, the name contains the full URI.

**Usage would look like:**
```
file:mydb.db?vfs=litestream&replica_url=s3://bucket/path&write_enabled=true&sync_interval=1s
```

**Implementation:** Parse URI parameters in `vfs.openMainDB()` (vfs.go ~line 135) using `net/url.Parse()`. Apply parsed values to the `VFSFile` instance, falling back to VFS-level defaults.

**Supported parameters:**
- `replica_url` — the replica backend URL (most critical — currently requires env var)
- `write_enabled` — enable write support
- `sync_interval` — write sync frequency
- `buffer_path` — write buffer persistence path
- `hydration_enabled` — enable background hydration
- `hydration_path` — hydration file path
- `poll_interval` — LTX polling frequency
- `cache_size` — page cache size

**Challenge:** `replica_url` creates a `ReplicaClient` — currently done once at VFS registration. Per-connection clients would need lifecycle management (close on VFSFile close). This is the most complex part.

### Approach 2: Extend PRAGMA Interface (runtime config after open)

The existing `FileControl()` mechanism (vfs.go:2271-2378) already supports read-write PRAGMAs like `litestream_write_enabled`. Extend this to cover more configuration:

```sql
PRAGMA litestream_poll_interval = '5s';
PRAGMA litestream_cache_size = 20971520;
PRAGMA litestream_hydration_enabled = 'true';
```

**Limitation:** PRAGMAs work per-connection and only after the DB is open. They can't configure `replica_url` since the ReplicaClient must exist before the first read. PRAGMAs are best for tuning parameters, not initial bootstrap config.

## Implementation Steps

1. **Verification spike** — Confirm URI params flow through `psanford/sqlite3vfs` to `VFS.Open()` name parameter
2. **Add URI parsing** in `openMainDB()` — extract and apply non-replica-url params (poll_interval, cache_size, etc.)
3. **Add ReplicaClient-per-connection support** — Allow `replica_url` in URI to create a per-file client, with proper cleanup on close
4. **Extend FileControl PRAGMAs** — Add read-write support for poll_interval, cache_size, hydration settings
5. **Update loadable extension** — Make env vars optional when URI params are provided; env vars become defaults
6. **Documentation** — Update examples showing URI-based and PRAGMA-based configuration

## Key Files

- `cmd/litestream-vfs/main.go` — VFS registration, env var reading (lines 42-112)
- `vfs.go` — VFS struct, Open(), openMainDB(), FileControl()
- `vfs_test.go` — Existing VFS tests
- `cmd/litestream-vfs/main_test.go` — Integration tests

## Community Feedback Welcome

We'd love input from anyone using (or wanting to use) the VFS extension as a library:

- Do URI parameters + PRAGMAs cover your use cases?
- Are there scenarios we haven't considered?
- Would you prefer a different API surface (e.g., a Go API for programmatic registration with config)?
- For non-Go embedders (Python, etc.), does the URI approach work with your SQLite bindings?

Please comment here with your thoughts so we can get this right before implementation.

### LINKED ISSUE #1231: Ability to set the poll interval for VFS when loaded as an extension
## Feature Description

We plan to use the VFS as a loaded SQLite extension and need to be able to adjust the poll interval (and possibly other parameters) in order to make it cost-feasible.

## Use Cases

Use the VFS as a loaded SQLite extension, polling at a non-default interval (e.g. longer than 1 second)
* Ideally the interval would also be dynamically adjustable so that we can turn it up and slow it down while keeping the cached state.

## Attempted Solutions

No workarounds to my knowledge.

## Proposal

I believe this PR: https://github.com/benbjohnson/litestream/pull/1157 would help with our needs given that it accepts a poll_interval as a URI parameter of the db spec.

Ideally the interval would be dynamically adjustable (e.g. via a PRAGMA, which was in an earlier version of the PR), but we could live without out it (by disconnecting and reconnecting with a different poll interval).

Thank you!

corylanou:
## Dependency status

**Resolved 2026-08-29.** Upstream `psanford/sqlite3vfs` shipped URI parameter support as the `URIOpener` interface at `v0.0.0-20260519004904-f9180fa2acc9` (2026-05-19), and closed psanford/sqlite3vfs#18 as superseded rather than rejected.

This PR now depends on that upstream version with **no `replace` directive** — `go.mod` is byte-identical to `main`'s. No code changes were needed: `vfs.go` already implements `OpenURI(name string, params map[string]string, flags sqlite3vfs.OpenFlag)`, which matches the upstream `URIOpener` signature exactly.

The earlier merge blocker on this PR no longer applies.

## Description

Add thread-safe VFS configuration for per-database runtime settings and URI parameters. This removes the need for loadable-extension users to rely on process-wide `setenv()` after startup, and lets callers configure options such as `replica_url` and `poll_interval` directly in the SQLite database URI.

**Key changes:**
- **Config registry** (`vfs_config.go`): `SetVFSConfig`/`GetVFSConfig`/`DeleteVFSConfig` keyed by database name, protected by `sync.RWMutex` with defensive copies on set/get to prevent concurrent mutation.
- **URI configuration**: supported VFS settings can be parsed from SQLite URI parameters and merged over registry configuration per connection.
- **Upstream dependency**: uses `github.com/psanford/sqlite3vfs` at the same version as `main`, no `replace` directive and no vendored copy. The `internal/sqlite3vfs` package that earlier revisions carried has been removed.
- **Unified parser**: `GoLitestreamConfigure()` and URI config now share the same `VFSConfig.Set()` validation path.
- **Per-connection clients**: `openMainDB()` creates a per-connection `ReplicaClient` when config specifies `replica_url`; clients are cleaned up on `VFSFile.Close()` and on `Open()` failure.
- **Optional env vars**: `LitestreamVFSRegister()` no longer requires `LITESTREAM_REPLICA_URL`, enabling per-database config via registry or URI parameters.
- **No PRAGMA support**: PRAGMA-based configuration was removed per review feedback.

## Motivation and Context

`setenv()` is not thread-safe on Linux. Library users who don't know VFS config values at process startup cannot safely call `os.Setenv` later. This PR adds registry and URI-based configuration so callers can configure each database connection without mutating process-wide environment variables.

Fixes #1150
Fixes #1231

## How Has This Been Tested?

- `go test -tags vfs -count=1 .`
- `go test -tags vfs -count=1 ./cmd/litestream-vfs`
- `go vet -tags vfs . ./cmd/litestream-vfs`
- `go build -tags "SQLITE3VFS_LOADABLE_EXT vfs" ./cmd/litestream-vfs`
- Commit hooks: `go-imports-repo`, `go-vet-repo-mod`, `go-staticcheck-repo-mod`
- After the 2026-08-29 upstream sync: `go build ./...` and `go build -tags vfs ./...`
- After the 2026-08-29 upstream sync: `go test -race -count=1 ./...` — all 10 packages green

## Types of changes

## Checklist
- [x] I have tested my changes


benbjohnson (changes requested):
I'm okay with setting parameters in the URL but I think the PRAGMA part is overkill. These should all be set at initialization time and PRAGMAs can be used anytime.

corylanou:
Updated for re-review:

- Removed PRAGMA-based configuration per review feedback.
- Added URI parameter support so options like `poll_interval` can be set from the SQLite database URI.
- Added an internal `sqlite3vfs` fork solely to expose SQLite filename/URI metadata to the Go VFS layer.
- Updated the PR body to note that this also fixes #1231.

Latest commit is `d63169d`; CI is green.

benbjohnson (approved):

go.mod
```diff
@@ -138,3 +138,5 @@ require (
 	google.golang.org/protobuf v1.36.10 // indirect
 	gopkg.in/yaml.v3 v3.0.1 // indirect
 )
+
+replace github.com/psanford/sqlite3vfs => github.com/corylanou/sqlite3vfs v0.1.0
```
ahirner: Can be removed after https://github.com/psanford/sqlite3vfs/pull/18  ?

darkgnotic:
Now that https://github.com/psanford/sqlite3vfs/commit/f9180fa2acc9fc03941664f53464cf0b1c9f74ac is in, is there still a merge blocker? We are hoping to be able to use this soon. Thank you!!  🙏

corylanou:
Dependency situation resolved.

Upstream landed URI parameter support itself as the `URIOpener` interface at `v0.0.0-20260519004904-f9180fa2acc9`, timestamped 2026-05-19 00:49:04Z — about a minute before psanford/sqlite3vfs#18 was closed at 00:50:12Z. The close was "superseded", not "rejected", and the shape shipped as `URIOpener` rather than the proposed `FilenameOpener`.

`main` has depended on real upstream with no `replace` since then; this branch was simply stale. It has now been merged with `main`, the `corylanou/sqlite3vfs` fork pin and `replace` directive dropped, and the require pinned to `main`'s upstream version. `go.mod` is byte-identical to `main`'s.

No source changes were required — `vfs.go` already implemented the upstream `OpenURI` signature.

Verified: builds clean with and without `-tags vfs`; `go test -race` green across all 10 packages.

PR #1167, opened 2026-02-25
fix(restore): tighten follow crash-recovery txid validation

dangra:
## Summary
- fix follow crash-recovery validation to compare against the earliest available snapshot (while still checking ahead-of-latest separately)
- keep strict validation when no snapshots are retained by scanning non-snapshot LTX ranges and rejecting stale/ahead/gap states
- add regression coverage for crash-recovery edge branches, including ahead snapshot, ahead non-snapshot, uncovered saved TXID, and no-files-available cases
- harden non-crash restore behavior by cleaning up `*.tmp` output files on failure and adding coverage for output-exists and canceled-context paths
- refactor `Restore()` control flow into explicit `full` / `resume` / `v3` mode routing with a single `r.follow()` call site

## Testing
- `go test ./...`

### DIFF
--- a/replica.go
+++ b/replica.go
@@ -550,8 +550,15 @@ func (r *Replica) Restore(ctx context.Context, opt RestoreOptions) (err error) {
 				return fmt.Errorf("cannot validate saved TXID for crash recovery: %w", itrErr)
 			}
 
-			var latestSnapshot *ltx.FileInfo
+			// Track the earliest snapshot (to detect pruned history) separately
+			// from the latest snapshot (to detect a TXID that is too far ahead).
+			// The iterator is sorted ascending by MinTXID, so the first item is
+			// the oldest available snapshot.
+			var earliestSnapshot, latestSnapshot *ltx.FileInfo
 			for snapshotItr.Next() {
+				if earliestSnapshot == nil {
+					earliestSnapshot = snapshotItr.Item()
+				}
 				latestSnapshot = snapshotItr.Item()
 			}
 			if err := snapshotItr.Err(); err != nil {
@@ -560,12 +567,65 @@ func (r *Replica) Restore(ctx context.Context, opt RestoreOptions) (err error) {
 			}
 			_ = snapshotItr.Close()
 
-			if latestSnapshot != nil {
-				if latestSnapshot.MinTXID > txid {
-					return fmt.Errorf("cannot resume follow mode: saved TXID %s is behind the earliest snapshot (min TXID %s); replica history has been pruned -- delete %s and %s-txid to re-restore", txid, latestSnapshot.MinTXID, opt.OutputPath, opt.OutputPath)
+			if earliestSnapshot != nil {
+				if earliestSnapshot.MinTXID > txid {
+					return fmt.Errorf("cannot resume follow mode: saved TXID %s is behind the earliest snapshot (min TXID %s); replica history has been pruned -- delete %s and %s-txid to re-restore", txid, earliestSnapshot.MinTXID, opt.OutputPath, opt.OutputPath)
 				}
 				if txid > latestSnapshot.MaxTXID {
 					return fmt.Errorf("cannot resume follow mode: saved TXID %s is ahead of latest snapshot (max TXID %s); delete %s and %s-txid to re-restore", txid, latestSnapshot.MaxTXID, opt.OutputPath, opt.OutputPath)
+				}
+			} else {
+				// Some replicas may have no retained snapshots. Fall back to checking
+				// all non-snapshot levels so we can still detect stale/ahead TXIDs
+				// instead of silently following from an unreachable position.
+				var earliestFile, latestFile *ltx.FileInfo
+				var txidCovered, resumeTXIDCovered bool
+				resumeTXID := txid
+				if txid != ^ltx.TXID(0) {
+					resumeTXID = txid + 1
+				}
+				for level := 0; level < SnapshotLevel; level++ {
+					itr, itrErr := r.Client.LTXFiles(ctx, level, 0, false)
+					if itrErr != nil {
+						return fmt.Errorf("cannot validate saved TXID for crash recovery: %w", itrErr)
+					}
+
+					for itr.Next() {
+						item := *itr.Item()
+						if earliestFile == nil || item.MinTXID < earliestFile.MinTXID {
+							earliestFile = &item
+						}
+						if latestFile == nil || item.MaxTXID > latestFile.MaxTXID {
+							latestFile = &item
+						}
+						if item.MinTXID <= txid && txid <= item.MaxTXID {
+							txidCovered = true
+						}
+						if item.MinTXID <= resumeTXID && resumeTXID <= item.MaxTXID {
+							resumeTXIDCovered = true
+						}
+					}
+					if err := itr.Err(); err != nil {
+						_ = itr.Close()
+						return fmt.Errorf("iterate ltx files for crash recovery validation: %w", err)
+					}
+					_ = itr.Close()
+				}
+
+				if earliestFile == nil {
+					return fmt.Errorf("cannot resume follow mode: no snapshots or LTX files available to validate saved TXID %s; delete %s and %s-txid to re-restore", txid, opt.OutputPath, opt.OutputPath)
+				}
+				if earliestFile.MinTXID > txid {
+					return fmt.Errorf("cannot resume follow mode: saved TXID %s is behind earliest available LTX file (min TXID %s); replica history has been pruned -- delete %s and %s-txid to re-restore", txid, earliestFile.MinTXID, opt.OutputPath, opt.OutputPath)
+				}
+				if txid > latestFile.MaxTXID {
+					return fmt.Errorf("cannot resume follow mode: saved TXID %s is ahead of latest available LTX file (max TXID %s); delete %s and %s-txid to re-restore", txid, latestFile.MaxTXID, opt.OutputPath, opt.OutputPath)
+				}
+				if !txidCovered {
+					return fmt.Errorf("cannot resume follow mode: saved TXID %s is not covered by any available LTX file range; replica history has gaps -- delete %s and %s-txid to re-restore", txid, opt.OutputPath, opt.OutputPath)
+				}
+				if latestFile.MaxTXID > txid && !resumeTXIDCovered {
+					return fmt.Errorf("cannot resume follow mode: next TXID %s is not covered by any available LTX file range; replica history has gaps -- delete %s and %s-txid to re-restore", resumeTXID, opt.OutputPath, opt.OutputPath)
 				}
 			}
 
--- a/replica.go
+++ b/replica.go
@@ -610,6 +610,7 @@ func (r *Replica) Restore(ctx context.Context, opt RestoreOptions) (err error) {
 
 	// Output to temp file & atomically rename.
 	tmpOutputPath := opt.OutputPath + ".tmp"
+	defer func() { _ = os.Remove(tmpOutputPath) }()
 	r.Logger().Debug("compacting into database", "path", tmpOutputPath, "n", len(rdrs))
 
 	f, err := os.Create(tmpOutputPath)
--- a/replica.go
+++ b/replica.go
@@ -535,102 +535,9 @@ func (r *Replica) Restore(ctx context.Context, opt RestoreOptions) (err error) {
 	// by reading the last applied TXID from the sidecar file.
 	if opt.Follow {
 		if _, statErr := os.Stat(opt.OutputPath); statErr == nil {
-			txid, readErr := ReadTXIDFile(opt.OutputPath)
-			if readErr != nil {
-				return fmt.Errorf("read txid file for crash recovery: %w", readErr)
-			}
-			if txid == 0 {
-				return fmt.Errorf("cannot resume follow mode: database exists but no -txid file found; delete the database to re-restore: %s", opt.OutputPath)
-			}
-			// Validate saved TXID is still reachable. If the earliest snapshot
-			// starts after our saved TXID, retention has pruned the history
-			// and we can't catch up incrementally.
-			snapshotItr, itrErr := r.Client.LTXFiles(ctx, SnapshotLevel, 0, false)
-			if itrErr != nil {
-				return fmt.Errorf("cannot validate saved TXID for crash recovery: %w", itrErr)
-			}
-
-			// Track the earliest snapshot (to detect pruned history) separately
-			// from the latest snapshot (to detect a TXID that is too far ahead).
-			// The iterator is sorted ascending by MinTXID, so the first item is
-			// the oldest available snapshot.
-			var earliestSnapshot, latestSnapshot *ltx.FileInfo
-			for snapshotItr.Next() {
-				if earliestSnapshot == nil {
-					earliestSnapshot = snapshotItr.Item()
-				}
-				latestSnapshot = snapshotItr.Item()
-			}
-			if err := snapshotItr.Err(); err != nil {
-				_ = snapshotItr.Close()
-				return fmt.Errorf("iterate snapshots for crash recovery validation: %w", err)
-			}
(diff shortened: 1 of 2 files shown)


corylanou (commented):
test

corylanou (commented):
test

corylanou (commented):
test

corylanou (commented):
test

cmd/litestream/restore.go
```diff
@@ -41,26 +45,78 @@ func (c *RestoreCommand) Run(ctx context.Context, args []string) (err error) {
...
+				case <-cancelCtx.Done():
+				}
+			}()
+
+			opt.OnRestored = func() error {
+				execArgs, err := shellwords.Parse(*execFlag)
+				if err != nil {
+					return fmt.Errorf("cannot parse exec command: %w", err)
+				}
+				cmd = exec.Command(execArgs[0], execArgs[1:]...)
```
corylanou: **[P1] Child process not bound to context**

This uses `exec.Command(...)` but the existing `replicate -exec` uses `exec.CommandContext(ctx, ...)` (see `replicate.go:375`). Without context binding, if `Restore()` exits due to a follow-loop error (e.g. network/replica failure), litestream returns but the child process keeps running orphaned.

Suggestion:
```go
cmd = exec.CommandContext(cancelCtx, execArgs[0], execArgs[1:]...)
```

cmd/litestream/restore.go
```diff
@@ -94,6 +150,8 @@ func (c *RestoreCommand) Run(ctx context.Context, args []string) (err error) {
 			return nil
 		}
 		return fmt.Errorf("no matching backup files available")
+	} else if errors.Is(err, context.Canceled) {
```
corylanou: **[P1] Child exit status lost when follow returns nil**

This only checks `execErr` when `r.Restore()` returns `context.Canceled`, but the follow loop returns `nil` when its context is canceled (see `replica.go` ~line 751: `case <-ctx.Done(): return nil`). So if the child exits non-zero and cancels context, follow returns `nil`, and litestream exits 0 instead of propagating the child failure.

You likely need:
```go
} else if err == nil && execErr != nil {
	return execErr
}
```
or restructure so `execErr` is always checked after `Restore()` returns.

cmd/litestream/restore.go
```diff
@@ -41,26 +45,78 @@ func (c *RestoreCommand) Run(ctx context.Context, args []string) (err error) {
...
+				select {
+				case sig := <-ch:
+					if cmd != nil && cmd.Process != nil {
+						slog.Info("sending signal to exec process", "signal", sig)
+						_ = cmd.Process.Signal(sig)
+						// cancel() is called by the cmd.Wait() goroutine when the child exits.
+					} else {
+						cancel() // signal arrived before child started
+					}
+				case <-cancelCtx.Done():
```
corylanou: **[P2] Signal goroutine handles only one signal + data race on `cmd`**

Two issues in this block (lines 65-76):

1. **Single signal**: The `select` consumes one signal then exits. If the child traps/ignores the first SIGINT, a second Ctrl+C cannot stop litestream. Consider looping to forward multiple signals, or sending SIGKILL on a second signal.

2. **Data race**: `cmd` is read here (line 68, signal goroutine) and written in `OnRestored` (line 84, restore goroutine) without synchronization. If Ctrl+C arrives during process startup, this is a race detectable by `-race`. Consider a mutex or channel handoff.

cmd/litestream/restore.go
```diff
@@ -41,26 +45,78 @@ func (c *RestoreCommand) Run(ctx context.Context, args []string) (err error) {
...
+				}
+			}()
+
+			opt.OnRestored = func() error {
+				execArgs, err := shellwords.Parse(*execFlag)
```
corylanou: **[P2] Whitespace-only `-exec` causes panic**

After `shellwords.Parse(*execFlag)`, line 84 indexes `execArgs[0]` without checking slice length. `-exec " "` parses to an empty slice, causing an index-out-of-range panic.

Suggestion:
```go
if len(execArgs) == 0 {
	return fmt.Errorf("exec command is empty")
}
```

replica.go
```diff
@@ -570,6 +570,11 @@ func (r *Replica) Restore(ctx context.Context, opt RestoreOptions) (err error) {
 			}
 
 			r.Logger().Info("resuming follow mode from crash recovery", "txid", txid, "output", opt.OutputPath)
+			if opt.OnRestored != nil {
+				if err := opt.OnRestored(); err != nil {
```
corylanou: **[Observation]** The `OnRestored` callback placement here (crash-recovery path) and on line 703 (normal path) looks correct — both fire before entering `r.follow()`. Good coverage of both entry points as described in the PR.

corylanou (commented):
## Code Review (3-pass, 6 unique findings)

Reviewed the `-exec` flag implementation against the PR description and existing `replicate -exec` behavior. Found several lifecycle/concurrency issues that should be addressed before merge.

**P1** = functional regression or data loss risk | **P2** = edge case or robustness issue

---

### Note: Finding outside this PR's diff

**[P1] Crash-recovery TXID validated against wrong snapshot** (`replica.go:555`)

The crash-recovery guard (from a prior commit on this branch) iterates all snapshot files but only keeps the last one (`latestSnapshot`). Since the iterator is sorted by `MinTXID` ascending (see `NewFileInfoSliceIterator` in the ltx package), `latestSnapshot` ends up being the **newest** snapshot, not the earliest. The check `latestSnapshot.MinTXID > txid` is meant to detect if history was pruned past the saved TXID, but comparing against the newest snapshot means a valid TXID within an older retained snapshot will be incorrectly rejected. Consider tracking both `earliestSnapshot` and `latestSnapshot`.

cmd/litestream/restore.go
dangra: Fixed in 7f339c18f3127ebe9f3d176168281d642cf8575e: switched to `exec.CommandContext(cancelCtx, ...)`. The child is now killed automatically when `cancelCtx` is cancelled, which happens via `defer cancel()` on any return path from `Run()`.

cmd/litestream/restore.go
dangra: Fixed in 7f339c18f3127ebe9f3d176168281d642cf8575e: `follow()` returns `nil` (not `context.Canceled`) on `ctx.Done`, so the previous check was dead code. Simplified to: return `execErr` whenever `Restore` returns `nil` or `context.Canceled`; treat any other non-nil error as a genuine failure.

cmd/litestream/restore.go
dangra: Both issues fixed in 7f339c18f3127ebe9f3d176168281d642cf8575e: (1) signal goroutine now loops so subsequent signals are forwarded; (2) replaced `var cmd *exec.Cmd` with `atomic.Pointer[exec.Cmd]`, stored after `Start()` so the signal goroutine either sees `nil` (cancel directly) or a fully running process.

cmd/litestream/restore.go
dangra: Fixed in 7f339c18f3127ebe9f3d176168281d642cf8575e: added `if len(execArgs) == 0 { return fmt.Errorf("exec command is empty") }` after `shellwords.Parse`.

dangra:
All findings addressed in 7f339c18f3127ebe9f3d176168281d642cf8575e. Summary:

| Finding | Fix |
|---|---|
| P1: child orphaned on follow error | `exec.CommandContext(cancelCtx, ...)` — `defer cancel()` kills child on any return path |
| P1: `execErr` never checked | `follow()` returns `nil` on `ctx.Done`, not `context.Canceled`; now return `execErr` whenever `Restore` returns `nil` or `context.Canceled` |
| P2: data race on `cmd` | `atomic.Pointer[exec.Cmd]`, stored after `Start()` |
| P2: single signal forwarded | Signal goroutine loops to forward multiple signals |
| P2: whitespace `-exec` panics | `len(execArgs) == 0` guard before indexing |
| P1 (out-of-diff): prune check uses wrong snapshot | Track `earliestSnapshot` and `latestSnapshot` separately; use `earliestSnapshot.MinTXID` for the pruning check, `latestSnapshot.MaxTXID` for the ahead check |

Full test suite (`go test -race ./...`) passes after all fixes.

cmd/litestream/restore.go
```diff
@@ -88,16 +165,19 @@ func (c *RestoreCommand) Run(ctx context.Context, args []string) (err error) {
...
 	if err := r.Restore(ctx, opt); errors.Is(err, litestream.ErrTxNotAvailable) {
 		if *ifReplicaExists {
 			slog.Info("no matching backups found")
 			return nil
 		}
 		return fmt.Errorf("no matching backup files available")
-	} else if err != nil {
+	} else if err != nil && !errors.Is(err, context.Canceled) {
 		return err
 	}
```
corylanou, line 99: **[P2] Context cancellation swallowed for non-exec restores**

The `!errors.Is(err, context.Canceled)` guard suppresses cancellation errors even when `-exec` is not used. If a caller sets a timeout or cancels the context, `Run()` returns `nil` instead of the cancellation error, falsely reporting a successful restore that was actually interrupted mid-operation.

`execErr` is always `nil` in the non-exec path, so `return execErr` becomes `return nil`.

Suggestion: only suppress `context.Canceled` when `-exec` is active:
```go
} else if errors.Is(err, context.Canceled) && *execFlag != "" {
	return execErr
} else if err != nil {
	return err
}
```

replica.go
```diff
@@ -560,16 +567,21 @@ func (r *Replica) Restore(ctx context.Context, opt RestoreOptions) (err error) {
 			}
 			_ = snapshotItr.Close()
```
corylanou: **[P2] Crash recovery accepted without snapshots to validate against**

When `SnapshotLevel` is empty (no snapshots exist), `earliestSnapshot` is `nil` and this entire validation block is skipped. If incremental history has been pruned but no snapshots remain, follow mode resumes from a stale TXID and polls forever without applying updates or surfacing an error — leaving the replica silently stale.

Consider explicitly rejecting crash recovery when no snapshots are available:
```go
if earliestSnapshot == nil {
	return fmt.Errorf("cannot resume follow mode: no snapshots available to validate saved TXID %s; delete %s and %s-txid to re-restore", txid, opt.OutputPath, opt.OutputPath)
}
```

corylanou (commented):
## Follow-up Review (3-pass, post-fix)

All 6 prior findings are confirmed fixed. Tests pass with `-race`. Nice work on the `atomic.Pointer` approach and the regression test for the snapshot bug.

Found 2 new issues below.

replica.go
dangra: Addressed in eb9016a. Crash-recovery validation now handles missing snapshots by scanning non-snapshot LTX levels and rejecting stale/ahead saved TXIDs, with regression coverage in `TestReplica_Restore_Follow_StaleTXID_NoSnapshots`.

cmd/litestream/restore.go
dangra: Fixed in eb9016a. `RestoreCommand.Run()` now only suppresses `context.Canceled` when `-exec` is active, and otherwise returns cancellation to the caller.

replica.go
```diff
@@ -560,13 +567,66 @@ func (r *Replica) Restore(ctx context.Context, opt RestoreOptions) (err error) {
...
 
-			if latestSnapshot != nil {
-				if latestSnapshot.MinTXID > txid {
-					return fmt.Errorf("cannot resume follow mode: saved TXID %s is behind the earliest snapshot (min TXID %s); replica history has been pruned -- delete %s and %s-txid to re-restore", txid, latestSnapshot.MinTXID, opt.OutputPath, opt.OutputPath)
+			if earliestSnapshot != nil {
+				if earliestSnapshot.MinTXID > txid {
```
benbjohnson: I'm not sure this makes sense. `MinTXID` should always be `1` for snapshots.

dangra:
Following up on the no-snapshot crash-recovery discussion: the current behavior is intentionally strict.

We require the saved TXID itself to remain covered by retained history, not only `txid+1`. If saved TXID coverage is lost, we fail fast and require re-restore instead of attempting to continue from an inferred position.

This is deliberate for crash-recovery safety and is now covered by explicit tests in this PR:
- `TestReplica_Restore_Follow_UncoveredSavedTXID_NoSnapshots`
- `TestReplica_Restore_Follow_AheadTXID_NoSnapshots`
- `TestReplica_Restore_Follow_NoSnapshotsAndNoLTXFiles`
- `TestReplica_Restore_Follow_StaleTXID_NoSnapshotsWithGap`

corylanou (approved):
Looks good. The two issues from the prior review are both fixed correctly:

1. Crash-recovery validation now compares against the earliest snapshot for pruning detection, not the newest
2. When no snapshots exist, falls through to non-snapshot LTX range validation instead of silently accepting a stale TXID

The gap detection logic (txid+1 reachability) is a nice addition — prevents resuming from an invalid position where incremental files have been pruned.

Helper method extraction is clean and doesn't change behavior. Test coverage looks solid with regression tests for the key edge cases (ahead snapshot, ahead non-snapshot, uncovered saved TXID, no files available).

All tests pass with `-race`.

replica.go
```diff
@@ -665,35 +697,113 @@ func (r *Replica) Restore(ctx context.Context, opt RestoreOptions) (err error) {
...
-	// Enter follow mode if enabled, continuously applying new LTX files.
-	if opt.Follow {
-		for _, rd := range rdrs {
-			if closer, ok := rd.(io.Closer); ok {
-				_ = closer.Close()
-			}
+	return infos[len(infos)-1].MaxTXID, nil
+}
+
+func (r *Replica) validateCrashRecoveryTXID(ctx context.Context, txid ltx.TXID, outputPath string) error {
```
benbjohnson, line 718: This can all be simplified a lot IMO:

1. Incrementally restore from L1 starting from the first LTX file that includes `txid+1`
2. Incrementally restore from L0 where L1 left off
3. If `txid+1` doesn't exist in L1, perform a full restore

We shouldn't get in an error state just because `txid+1` was been dropped due to snapshot retention enforcement.

benbjohnson (changes requested):

replica.go
dangra: Makes sense 🤦🏼‍♂️

replica.go
dangra: Ok, a few questions here:

1. Should it check for gaps in L1 or L0 files or that is not expected at all? 
  "gaps" in the sense of ltx files not assembling into a contiguous range from txid+1 to max txid.

2.  are you asking to ignore levels past L1? if it can't restore from L1 or L0, do full snapshot restore (L9). right? 

3. Can it assume the data is valid and if there is a ltx file that includes txid+1, then it will be resumable from that point? no need to check nor validate further whether it reaches max txid.

benbjohnson: > Should it check for gaps in L1 or L0 files or that is not expected at all? "gaps" in the sense of ltx files not assembling into a contiguous range from txid+1 to max txid.

Gaps shouldn't occur at all. If there are gaps then there's a compaction issue. When compaction occurs, it always looks at the max of the lower level and then compacts between the MaxTXID of the lower level and the MaxTXID of the current level.

> are you asking to ignore levels past L1? if it can't restore from L1 or L0, do full snapshot restore (L9). right?

Basically yes. L0 files are short lived and usually get deleted after 5 minutes or so. L1-L9 files all have the same lifetime so if a TXID doesn't exist in L1 then it won't exist in anything higher.

The only reason the TXID wouldn't exist in L1+ is if the retention period has passed so that's when you'd fall back on a full restore. A full restore builds a full restore plan across all levels by using the `CalcRestorePlan()` function.

> Can it assume the data is valid and if there is a ltx file that includes txid+1, then it will be resumable from that point? no need to check nor validate further whether it reaches max txid.

Yes, as long as `txid+1` exists in L1 then you should be able to apply all the files from that point to the last file in L1. You can then apply all the files in L0 that are after L1's MaxTXID so you can pick up any transactions that were created since the last L1 compaction.

PR #1170, opened 2026-02-26
feat(restore): add post-restore integrity validation

corylanou:
## Description

Adds optional integrity checking after database restores to catch corruption early, regardless of source.

### Changes

- **`IntegrityCheckMode` type** (`None`/`Quick`/`Full`): Controls whether and how `PRAGMA integrity_check` or `PRAGMA quick_check` runs after restore.
- **`checkIntegrity()` function**: Runs the selected integrity check on the restored database. On failure, cleans up the corrupt file and returns an error.
- **Integrated into `Restore()` and `RestoreV3()`**: Both code paths now support post-restore validation.
- **`EnsureExists()` defaults to `IntegrityCheckQuick`**: K8s init containers calling `EnsureExists()` catch corrupt restores before the application starts.
- **`-integrity-check` CLI flag**: Users can opt into validation via `litestream restore -integrity-check quick|full`.

### Why

Even with the root cause fix for #1164 (PR #1166), integrity checking provides defense in depth against corruption from any source — disk errors, buggy storage backends, network corruption during transfer, etc. It's cheap insurance: `quick_check` is fast and catches most structural issues.

Related to #1164

## How Has This Been Tested?

- `go test -race -count=1 ./...` — all tests pass
- New tests in `replica_internal_test.go`:
  - `TestCheckIntegrity_Quick_ValidDB` — valid DB returns no error
  - `TestCheckIntegrity_Full_ValidDB` — full mode on valid DB returns no error
  - `TestCheckIntegrity_None_Skips` — `IntegrityCheckNone` returns nil without opening DB
  - `TestCheckIntegrity_CorruptDB` — corrupted DB returns integrity check error

## Types of changes

## Checklist

### DIFF
--- a/replica.go
+++ b/replica.go
@@ -529,6 +529,8 @@ func (r *Replica) Restore(ctx context.Context, opt RestoreOptions) (err error) {
 		return fmt.Errorf("cannot use follow mode with -txid")
 	} else if opt.Follow && !opt.Timestamp.IsZero() {
 		return fmt.Errorf("cannot use follow mode with -timestamp")
+	} else if opt.IntegrityCheck != IntegrityCheckNone && opt.IntegrityCheck != IntegrityCheckQuick && opt.IntegrityCheck != IntegrityCheckFull {
+		return fmt.Errorf("unsupported integrity check mode: %d", opt.IntegrityCheck)
 	}
 
 	// In follow mode, if the database already exists, attempt crash recovery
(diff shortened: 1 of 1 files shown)


replica.go
```diff
@@ -1177,6 +1197,59 @@ func checkpointV3(dbPath string) error {
...
+		if err := rows.Scan(&result); err != nil {
+			return fmt.Errorf("scan integrity check result: %w", err)
+		}
+		if result != "ok" {
+			results = append(results, result)
+		}
+	}
+	if err := rows.Err(); err != nil {
+		return fmt.Errorf("iterate integrity check results: %w", err)
+	}
```
benbjohnson: This should always be a single row. You can just use QueryRowContext() instead of `QueryContext()` and then scanning.

corylanou: Fixed: Simplified to use `QueryRowContext()` with a single `Scan()` call. Since the integrity check returns a single result row, this is cleaner than iterating through rows.

PR #1173, opened 2026-02-27
feat(s3): add manifest file to reduce LIST operations

### LINKED ISSUE #1172: Reduce S3 LIST operations by introducing manifest files for LTX replication levels
## Feature Description

Reduce the number of S3 LIST operations by introducing manifest files that describe the expected LTX files for each replication level (e.g., L0 and L1). Readers such as `litestream restore` could fetch the manifest instead of repeatedly listing the bucket.

## Use Cases

Litestream currently polls the replica frequently (default every second), which results in a large number of S3 LIST operations to discover new LTX files.

In one of my applications, I observed approximately:

- ~100k LIST operations  
- ~4k GET operations  

Since LIST operations are more expensive and scale poorly, this significantly increases storage API costs.

The goal is to reduce LIST operations and replace them with fewer, more efficient GET requests.

## Attempted Solutions

Currently using Litestream with default polling configuration. The only way to reduce LIST operations is to increase the polling interval, which increases replication latency and restore detection time.

No clean workaround exists to maintain low latency while minimizing LIST costs.

## Proposal

The writer could periodically upload a manifest file (for example, JSON) containing the expected set of LTX files for each level (L0/L1).

Readers would:

- Fetch the manifest via GET
- Determine which LTX files exist
- Avoid issuing frequent LIST requests

This would significantly reduce S3 API costs and improve scalability, especially for frequently polling replicas.

corylanou:
## Description

Introduces a `manifest.json` file that S3 writers maintain after each mutation. Readers (restore/follow mode) fetch this single file via GET instead of calling `ListObjectsV2` (LIST) every polling cycle. This replaces expensive LIST operations with cheap GET operations.

**New files:**
- `s3/manifest.go` — Manifest/ManifestEntry types, sorted insertion, removal, level filtering, and a `manifestIterator` implementing `ltx.FileIterator`
- `s3/manifest_test.go` — 9 unit tests covering add/remove/filter/JSON round-trip/iterator

**Modified files:**
- `s3/replica_client.go` — Added `ManifestEnabled` (reader) and `ManifestWriteEnabled` (writer) flags; manifest I/O methods; hooks in `WriteLTXFile`, `DeleteLTXFiles`, `DeleteAll`, and `LTXFiles`
- `cmd/litestream/main.go` — Enables `ManifestWriteEnabled` for all S3 clients created via config
- `cmd/litestream/restore.go` — Enables `ManifestEnabled` for restore/follow mode S3 clients

## Motivation and Context

A user observed ~100k LIST operations vs ~4k GET operations when Litestream polls S3 every second. LIST costs 5-10x more than GET on S3. This change reduces S3 costs by replacing per-second LIST calls with a single GET of a small manifest file.

Fixes #1172

## How Has This Been Tested?

- `go build ./...` — compiles cleanly
- `go test -race -count=1 ./...` — all tests pass (including 9 new manifest tests)
- `pre-commit run --all-files` — all hooks pass (imports, vet, staticcheck)

**Backward compatibility verified by design:**
- Old writer + new reader → no manifest.json exists → reader falls back to LIST
- New writer + old reader → manifest.json created but ignored (doesn't match LTX filename pattern)
- Corrupt manifest → JSON decode fails → reader falls back to LIST
- Missing manifest → 404 from S3 → reader falls back to LIST

## Types of changes

## Checklist

### DIFF
--- a/cmd/litestream/restore.go
+++ b/cmd/litestream/restore.go
@@ -16,7 +16,6 @@ import (
 
 	"github.com/benbjohnson/litestream"
 	"github.com/benbjohnson/litestream/internal"
-	"github.com/benbjohnson/litestream/s3"
 )
 
 // RestoreCommand represents a command to restore a database from a backup.
@@ -321,9 +320,7 @@ func (c *RestoreCommand) loadFromURL(ctx context.Context, replicaURL string, ifD
 	if err != nil {
 		return nil, err
 	}
-	if s3Client, ok := r.Client.(*s3.ReplicaClient); ok {
-		s3Client.ManifestEnabled = true
-	}
+	r.Client.SetManifestEnabled(true)
 	_, err = r.CalcRestoreTarget(ctx, *opt)
 	return r, err
 }
@@ -359,9 +356,7 @@ func (c *RestoreCommand) loadFromConfig(_ context.Context, dbPath, configPath st
 		return nil, errSkipDBExists
 	}
 
-	if s3Client, ok := db.Replica.Client.(*s3.ReplicaClient); ok {
-		s3Client.ManifestEnabled = true
-	}
+	db.Replica.Client.SetManifestEnabled(true)
 	return db.Replica, nil
 }
 
--- a/cmd/litestream/restore.go
+++ b/cmd/litestream/restore.go
@@ -296,6 +296,16 @@ func (c *RestoreCommand) prepareOutputPath(path string, force bool) error {
 	return nil
 }
 
+type manifestEnabler interface {
+	SetManifestEnabled(bool)
+}
+
+func enableManifest(client litestream.ReplicaClient) {
+	if client, ok := client.(manifestEnabler); ok {
+		client.SetManifestEnabled(true)
+	}
+}
+
 // loadFromURL creates a replica & updates the restore options from a replica URL.
 func (c *RestoreCommand) loadFromURL(ctx context.Context, replicaURL string, ifDBNotExists bool, opt *litestream.RestoreOptions) (*litestream.Replica, error) {
 	if opt.OutputPath == "" {
@@ -320,7 +330,7 @@ func (c *RestoreCommand) loadFromURL(ctx context.Context, replicaURL string, ifD
 	if err != nil {
 		return nil, err
 	}
-	r.Client.SetManifestEnabled(true)
+	enableManifest(r.Client)
 	_, err = r.CalcRestoreTarget(ctx, *opt)
 	return r, err
 }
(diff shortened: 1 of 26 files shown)


cmd/litestream/restore.go
```diff
@@ -121,6 +122,9 @@ func (c *RestoreCommand) loadFromURL(ctx context.Context, replicaURL string, ifD
 	if err != nil {
 		return nil, err
 	}
+	if s3Client, ok := r.Client.(*s3.ReplicaClient); ok {
+		s3Client.ManifestEnabled = true
+	}
```
benbjohnson: This seems like something we should make available to every type of client. It'd also be good to not have significantly different code paths depending on the client.

benbjohnson: This seems like something we should make available to every type of client. It'd also be good to not have significantly different code paths depending on the client.

benbjohnson: Also, this should be opt-in. I'm not sure that the manifest is a net positive for all use cases and could be extra `PUT` costs for someone not running another node in follow mode.

s3/replica_client.go
```diff
@@ -1140,9 +1160,139 @@ func (c *ReplicaClient) DeleteAll(ctx context.Context) error {
...
+		m = NewManifest()
+	}
+	c.manifest = m
 	return nil
 }
 
+func (c *ReplicaClient) updateManifestAfterWrite(ctx context.Context, info *ltx.FileInfo) {
+	if !c.ManifestWriteEnabled {
+		return
+	}
```
benbjohnson: If `ManifestWriteEnabled` is false then we should ensure the manifest does not exist after the first mutation (write or delete). Otherwise, if the user has manifests enabled and then disables them then the manifest will continue to exist even when the list of files has changed.

We should only attempt to delete once on the first mutation though. We don't want to have a `DELETE` call on every mutation.

benbjohnson (changes requested):

cmd/litestream/restore.go
corylanou: Fixed in 244bf7c:

1. **Interface abstraction**: Added `SetManifestEnabled(bool)` to the `ReplicaClient` interface. All backends implement it (S3 sets the field, others are no-ops). This removes the S3-specific type assertions from restore.go.

2. **Opt-in config**: Manifest writing is now disabled by default. Users must explicitly enable it via `manifest: true` in YAML config. This avoids extra PUT costs for users not running follow-mode replicas.

s3/replica_client.go
corylanou: Fixed in 244bf7c:

Added `cleanupStaleManifest()` that deletes manifest.json on the first mutation when `ManifestWriteEnabled=false`. Key details:

- Uses `manifestCleanupAttempted` flag to ensure only one DELETE attempt per client lifetime
- Uses `ManifestConfigured` flag to only run cleanup for clients created from config (avoids cleanup attempts in tests and programmatic clients that never had manifests)
- Cleanup is called from both `updateManifestAfterWrite()` and `updateManifestAfterDelete()` when manifest writing is disabled
- Errors are DEBUG-logged and non-fatal (consistent with other manifest operations)

PR #1184, opened 2026-03-04
ci(pr-metrics): replace collapsible sections with plain headers

benbjohnson:
## Summary
- Replace `<details>`/`<summary>` HTML tags with `###` headers so all PR metrics sections (Binary Size, Dependency Changes, govulncheck Output, Build Info, History) are always visible
- Add backward compatibility for parsing history from both old collapsible and new plain comment formats

## Test plan
- [ ] Verify workflow YAML is valid
- [ ] Open a test PR to confirm the comment renders with all sections expanded
- [ ] Verify history parsing works on PRs that have existing comments with `<details>` tags


benbjohnson:
@corylanou Heads up that I removed the collapsible sections of the PR Build Metrics because I got annoyed from clicking them open everytime. :)

PR #1192, opened 2026-03-10
perf(db): cache Pos() result to avoid repeated disk reads

benbjohnson:
## Summary

- Cache the replication position from the latest L0 LTX file using `atomic.Pointer[ltx.Pos]`, eliminating repeated disk reads on every `Pos()` call
- Invalidate the cache in all code paths that modify L0 files: deletion, retention enforcement, sync errors, replica fetch, and local state reset
- On successful sync, update the cache directly from the encoder rather than requiring a subsequent disk read

## Benchmarks

**TestIdleCPUProfile with 500 idle databases, 10-second CPU profile:**

| Metric | `main` (baseline) | `cache-db-pos` | Change |
|---|---|---|---|
| **Total CPU samples** | **8.89s (87.80% load)** | **2.86s (28.23% load)** | **-68% CPU** |
| `DB.Pos()` cumulative | 2.58s (29.02%) | 0s (not visible) | eliminated |
| GC (`gcDrain`) | 4.69s (52.76%) | 0.76s (26.57%) | -84% |
| `syscall.syscall` | 3.25s (36.56%) | 1.64s (57.34% of smaller total) | -50% absolute |

The GC improvement is a cascading effect: `Pos()` was allocating LTX decoders, file handles, and buffers on every monitor tick. With the cache, those allocations vanish.

## Test plan

- [x] Existing tests pass (`go test -race ./...`)
- [x] Tests that write L0 files directly call `invalidatePosCache()` to reflect the new caching behavior
- [x] CPU profile benchmark confirms `Pos()` eliminated from idle hot path

### DIFF
--- a/db.go
+++ b/db.go
@@ -16,7 +16,6 @@ import (
 	"strconv"
 	"strings"
 	"sync"
-	"sync/atomic"
 	"time"
 
 	"github.com/prometheus/client_golang/prometheus"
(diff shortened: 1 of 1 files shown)


benbjohnson:
@codex review

Issue #1210, opened 2026-03-25
perf: reduce idle CPU, memory, and cloud API costs


corylanou:
## Problem

Litestream consumes unnecessary CPU, memory, and cloud storage API calls even when monitored databases have zero write activity. Multiple users have reported this across different versions and backends:

- #992 — 0.7% CPU when idle (single DB)
- #955 — 100-200% CPU with idle databases
- #1051 — Directory watcher with 400 DBs crashes (CPU >100%, memory >1.5GB)
- #1171 — 480 PUTs/hour despite zero writes and 1h sync-interval
- #1175 — Runaway replication with no changes
- #1197 — Full snapshots every few minutes on idle DB

The root causes are structural: Litestream uses fixed-interval polling loops that perform significant I/O work unconditionally, and the notification mechanism always fires even when no changes occurred.

### 1. Notify channel fires unconditionally (`db.go:1048-1052`)

The `if changed` guard is **commented out**. Every 1-second `Sync()` call closes and recreates the notify channel, waking ALL replica monitors regardless of whether anything changed:

```go
// Notify replicas of WAL changes.
// if changed {
close(db.notify)
db.notify = make(chan struct{})
// }
```

This means even on a completely idle database, the replica monitor wakes every second and calls `Replica.Sync()`, which opens LTX files, calculates positions, and potentially uploads to cloud storage. This is the primary cause of #1171 (480 PUTs/hour) and #1175 (runaway replication).

### 2. No early-exit in `Sync()` for unchanged WAL (`db.go:994`)

`DB.Sync()` performs the following work on every tick, even when the database is completely idle:
- Acquires mutex (`db.mu.Lock()`)
- Calls `db.init()` which does `os.Stat()` on the db path
- Calls `db.ensureWALExists()` which does `os.Stat()` on the WAL file
- Calls `db.verify()` which opens and decodes the latest LTX file, reads the WAL header, and potentially reads WAL frames
- Calls `db.sync()` which opens the WAL file, creates a `WALReader`, and calls `PageMap()`
- Calls `os.Stat()` on the database file again for metrics

There is no cheap "has the WAL file changed since last sync?" check at the top. A simple comparison of the WAL file's mtime and size against cached values would skip all of this work in the common idle case.

### 3. Multiple 1-second polling loops per database

Each database creates at minimum 2 goroutines with 1-second tickers:
- `db.monitor()` — 1s ticker (`db.go:2257`)
[...]

PR #1222, opened 2026-03-31
refactor(db,replica): extract sync/checkpoint/monitor subsystem

corylanou:
## Description

Refactor the sync/checkpoint/monitoring subsystem per #1221, and fix a pre-existing page loss bug uncovered during soak testing.

### Refactoring (Phases 1, 3, 4, 6 of #1221)

**Phase 1 — SyncResult return type (`db.go`):** `verifyAndSync()` and `sync()` return a `syncResult` struct instead of mutating `db.lastSyncedWALOffset`, `db.syncedToWALEnd`, `db.pos`, and `db.maxLTXFileInfos` as side effects. New `applySyncResult()` centralizes all state application in one visible call site.

**Phase 3 — Checkpoint phase extraction (`db.go`):** Extract `checkpoint()` into named phase methods (`checkpointPreSync`, `checkpointExec`, `checkpointPostSync`), making the checkpoint protocol explicit.

**Phase 4 — ReplicaScheduler (`replica.go`):** Extract `syncScheduler` type from `Replica.monitor()`, separating scheduling/backoff from sync correctness and error recovery. Extract `handleSyncError()` for LTX error detection, auto-recovery, and rate-limited logging. `monitor()` shrinks from 114 lines to a 30-line coordination loop.

**Phase 6 — Restore separation (`replica.go` → `restore.go`):** Move ~1200 lines of restore-related code to its own file. `replica.go` shrinks from ~1775 to ~590 lines.

### Bug Fix: Page loss under sustained writes

Soak testing at 500 writes/sec revealed a pre-existing bug where pages existed in the database but were never captured in any L0 file, causing restore failures with "nonsequential page numbers" errors. The root cause:

During `execCheckpoint()`, litestream releases its read transaction (`db.rtx`) to allow the SQLite checkpoint to run. In this window, the **application's** SQLite auto-checkpoint (`wal_autocheckpoint`, default 1000 frames) can independently move WAL frames to the DB and trigger a WAL restart. Pages moved during this uncontrolled window were never captured in any L0 file because the post-checkpoint sync used flag-based heuristics (`syncedToWALEnd`) that had become stale.

**Fix:** The post-checkpoint sync now always creates a snapshot — reading ALL pages from the DB under the write lock. This is correct by construction: no flags, no detection heuristics, no edge cases. The snapshot captures every page regardless of what happened during the checkpoint window.

**Supporting fixes:**
- **TRUNCATE → PASSIVE downgrade:** When the pre-sync can't reach WAL end (sustained writes), downgrade TRUNCATE to PASSIVE. PASSIVE copies frames without truncating, so uncaptured frames remain in the WAL for the next sync.
- **Snapshot on PrevFrameMismatch:** When the WAL changes between `verify()` and `sync()`, force a snapshot to capture pages from old WAL frames.
- **Clear `syncedToWALEnd` before checkpoint:** Prevents the stale flag from suppressing necessary snapshots after the checkpoint window.
- **L0 retention TXID range coverage:** L0 retention now checks actual L1 TXID ranges instead of a single `maxL1TXID` threshold, preventing deletion of L0 files in gaps between L1 files.
- **Retention gating:** `EnforceSnapshotRetention` only skips on actual compaction failures, not on `ErrNoCompaction`/`ErrCompactionTooEarly`.
- **Snapshot page count validation:** `writeLTXFromDB()` verifies all pages 1..commit are written, catching page gaps at creation time.

### Phases deferred to follow-up PRs

- **Phase 2 (SyncState/executor split):** Lock scope narrowing needs careful concurrency analysis.
- **Phase 5 (VFSFile decomposition):** 40+ fields, 30+ methods crossing multiple field groups.

## Motivation and Context

Since January 2026, recurring bugs cluster around the same code paths in `db.go` and `replica.go`. The #1210 idle optimization required 7 follow-up commits because the coupling between change detection, notification, checkpoint, and verification was implicit. This refactor makes state transitions explicit and concerns separated. The page loss bug was pre-existing and reproduced 100% under sustained 500 writes/sec — it was uncovered by the soak testing done to validate the refactoring.

Fixes #1221

## How Has This Been Tested?

- `go test -race -count=1 ./...` — all packages pass
- Codex code review completed — 2 findings addressed (removed debug validation from restore hot path, fixed retention gating for quiet databases)
- Soak tests pass across all three load profiles (2 consecutive clean runs):
  - **low-volume** (10 writes/sec): restoration OK, integrity OK, 0 violations
  - **burst-volume** (500 writes/sec bursts): restoration OK, integrity OK, 0 violations
  - **high-volume** (500 writes/sec sustained): restoration OK, integrity OK, 0 missing pages, 0 violations

## Types of changes

## Checklist

### DIFF
--- a/compactor.go
+++ b/compactor.go
@@ -350,18 +350,26 @@ func (c *Compactor) EnforceL0Retention(ctx context.Context, retention time.Durat
 	if err != nil {
 		return fmt.Errorf("fetch l1 files: %w", err)
 	}
-	var maxL1TXID ltx.TXID
+	type txidRange struct{ min, max ltx.TXID }
+	var l1Ranges []txidRange
 	for itr.Next() {
 		info := itr.Item()
-		if info.MaxTXID > maxL1TXID {
-			maxL1TXID = info.MaxTXID
-		}
+		l1Ranges = append(l1Ranges, txidRange{info.MinTXID, info.MaxTXID})
 	}
 	if err := itr.Close(); err != nil {
 		return fmt.Errorf("close l1 iterator: %w", err)
 	}
-	if maxL1TXID == 0 {
+	if len(l1Ranges) == 0 {
 		return nil
+	}
+
+	coveredByL1 := func(txid ltx.TXID) bool {
+		for _, r := range l1Ranges {
+			if txid >= r.min && txid <= r.max {
+				return true
+			}
+		}
+		return false
 	}
 
 	threshold := time.Now().Add(-retention)
@@ -390,7 +398,7 @@ func (c *Compactor) EnforceL0Retention(ctx context.Context, retention time.Durat
 			break
 		}
 
-		if info.MaxTXID <= maxL1TXID {
+		if coveredByL1(info.MaxTXID) {
 			deleted = append(deleted, info)
 		}
 	}
@@ -420,7 +428,7 @@ func (c *Compactor) EnforceL0Retention(ctx context.Context, retention time.Durat
 		}
 	}
 
-	c.logger.Info("l0 retention enforced", "deleted_count", len(deleted), "max_l1_txid", maxL1TXID)
+	c.logger.Info("l0 retention enforced", "deleted_count", len(deleted), "l1_ranges", len(l1Ranges))
 
 	return nil
 }
--- a/db.go
+++ b/db.go
@@ -2458,15 +2458,47 @@ func (db *DB) checkpointWithExecutor(ctx context.Context, mode string, exec *syn
 	defer db.chkMu.Unlock()
 	exec.checkpointAttempted = true
 
+	before, err := db.checkpointStart(ctx, mode, exec)
+	if err != nil {
+		return false, err
+	}
+	defer func() {
+		if before.barrierTx != nil {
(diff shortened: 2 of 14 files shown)


corylanou:
**Role: reference / soak proof — not a merge target.**

This draft proved the sync/checkpoint/monitor extraction direction and, under soak, surfaced the page-loss bug. Per the "really, really small increments" constraint on #1221, it stays open as reference/proof rather than merging. The real work ships as small slices.

Current navigation hub for what's actually in review: https://gist.github.com/corylanou/f2709d9868cad1943c63096bf978e881

corylanou:
Closing this POC. The extraction it prototyped landed through the executor cluster instead: #1289 (reverted by #1313, re-landed as #1314) and #1290. `syncExecutor` lives on `main` at `db.go:222` with `newSyncExecutor` at `db.go:1930`.

This branch is 17 commits behind with conflicts across `db.go`, `replica.go`, `store.go`, and `compactor.go`, so there is nothing worth salvaging by rebasing it against the shipped implementation.

#1221 stays open — the executor half is done, but the checkpoint/monitoring extraction it proposes is not fully covered by what merged.

PR #1224, opened 2026-04-05
refactor(db): return sync results explicitly

corylanou:
## Summary
- change `verifyAndSync()` to return a `syncResult` instead of relying on `sync()` to apply sync state as a side effect
- apply sync results explicitly in `Sync()` and `checkpoint()`
- add focused unit coverage for `applySyncResult()`

## Why
This is the first increment from #1221, following Ben's comment to start with `SyncResult` and keep the refactor very small.

The existing proof-of-concept work in #1222 and branch `issue-1221-proposal-incremental-extraction-of-sync-checkpoint-monitorin` was useful to validate the overall direction, but it is too broad to merge safely or review comfortably in one pass. The plan here is to keep that branch as reference material only, and land the extraction as small, reviewable steps instead.

## Scope
This PR intentionally stops at the `SyncResult` extraction. It does not include the wider checkpoint restructuring, sync-state split, restore extraction, or soak-test expansion from the proof-of-concept branch.

## Testing
- `go test ./...`
- `go test -race ./...`
- `go test . -run '^(TestApplySyncResult|TestDB_Sync_UpdatesMetrics|TestDB_Sync_InitErrorMetrics)$' -count=1`
- `go test . -run '^(TestDB_CheckpointDoesNotTriggerSnapshot|TestDB_CheckpointPageGapWithConcurrentWrites|TestDB_Sync_CompactionValidAfterGrowthAndCheckpoint)$' -count=1`
- `go test -count=1 -v -tags='integration,soak' -run '^(TestLTXBehavior|TestLTXBehavior_NoExcessiveSnapshots)$' -test.short -timeout=30m ./tests/integration`

### DIFF
--- a/db.go
+++ b/db.go
@@ -1019,17 +1019,18 @@ func (db *DB) Sync(ctx context.Context) (err error) {
 		return fmt.Errorf("ensure wal exists: %w", err)
 	}
 
-	origWALSize, newWALSize, synced, err := db.verifyAndSync(ctx, false)
+	result, err := db.verifyAndSync(ctx, false)
 	if err != nil {
 		return err
 	}
+	db.applySyncResult(result)
 
 	// Track that data was synced for time-based checkpoint decisions.
-	if synced {
+	if result.synced {
 		db.syncedSinceCheckpoint = true
 	}
 
-	if err := db.checkpointIfNeeded(ctx, origWALSize, newWALSize); err != nil {
+	if err := db.checkpointIfNeeded(ctx, result.origWALSize, result.newWALSize); err != nil {
 		return fmt.Errorf("checkpoint: %w", err)
 	}
 
@@ -1044,7 +1045,7 @@ func (db *DB) Sync(ctx context.Context) (err error) {
 	if fi, err := os.Stat(db.path); err == nil {
 		db.dbSizeGauge.Set(float64(fi.Size()))
 	}
-	db.walSizeGauge.Set(float64(newWALSize))
+	db.walSizeGauge.Set(float64(result.newWALSize))
 
 	// Notify replicas of WAL changes.
 	// if changed {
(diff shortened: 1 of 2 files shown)


corylanou:
Follow-up after comparing this phase-1 branch against `origin/main` and trying to answer whether the `SyncResult` extraction made behavior worse.

What I added on this branch:
- `7272b64` adds a deterministic seam test for the phase-1 refactor: `TestVerifyAndSync_DelaysStateMutationUntilApply`
- That test verifies `verifyAndSync()` returns updated state without mutating the DB object until `applySyncResult()` is called
- This is the part of the change I trust most, because it directly exercises the refactor seam instead of depending on soak timing

Commands run on this branch:
- `go test -count=1 ./... -run '^(TestApplySyncResult|TestVerifyAndSync_DelaysStateMutationUntilApply)$'`
- `go test -count=1 -v -tags='integration,soak' -run '^TestComprehensiveSoak$' -test.short -timeout=1h ./tests/integration`
- `SOAK_AUTO_PURGE=yes go test -count=1 -v -tags='integration,soak,docker' -run '^TestMinIOSoak$' -test.short -timeout=30m ./tests/integration`
- `go test -count=1 -v -tags='integration,soak' -run '^(TestLTXBehavior|TestLTXBehavior_NoExcessiveSnapshots)$' -test.short -timeout=30m ./tests/integration`

Focused comparison runs against `origin/main`:
- `TestLTXBehavior_NoExcessiveSnapshots` rerun 3x on this branch: `0/3` logged `nonsequential page numbers`
- `TestLTXBehavior_NoExcessiveSnapshots` rerun 3x on `origin/main`: `1/3` logged `nonsequential page numbers`
- `TestLTXBehavior/high-volume` rerun 3x on this branch: `1/3` logged `nonsequential page numbers`
- `TestLTXBehavior/high-volume` rerun 3x on `origin/main`: `0/3` logged `nonsequential page numbers`

Interpretation:
- I do not have evidence that phase 1 clearly regresses behavior relative to `main`
- I also do not think the system is fully clean: the intermittent `nonsequential page numbers in snapshot transaction` compaction error exists on both code lines
- Because that error reproduces on `main` too, I did not turn it into a new hard failing soak assertion in this PR. That would fail baseline and would not tell us whether phase 1 itself is acceptable
- Instead, I added the deterministic state-seam test for the actual refactor so we have coverage that `verifyAndSync()` no longer mutates DB state implicitly

My read at this point:
- If the goal of this PR is only Ben's requested first increment, I think the code change is still reviewable on its own merits
- If the intermittent compaction issue needs to block even small refactors, it should probably be addressed as a separate bug first or in parallel, because it is not unique to this branch

I wanted this written down here so the merge decision can be based on actual branch-vs-main comparison instead of a single flaky soak run.

benbjohnson:
### Code review

No issues found. Checked for bugs and CLAUDE.md compliance.

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

benbjohnson (approved):

PR #1228, opened 2026-04-07
refactor(db): extract syncState struct for sync-tracking fields

corylanou:
## Description

Extract the three mutable sync-tracking fields (`syncedSinceCheckpoint`, `syncedToWALEnd`, `lastSyncedWALOffset`) from the `DB` struct into a new `syncState` struct. Thread `*syncState` as a parameter through `verifyAndSync()`, `verify()`, `sync()`, `checkpointIfNeeded()`, `checkpoint()`, and `applySyncResult()`.

This is Phase 2 of the incremental extraction plan from #1221, building on the `syncResult` return type introduced in PR #1224 (Phase 1).

## Motivation and Context

Decoupling these fields from `DB` and passing them explicitly prepares for Phase 3 (checkpoint decomposition), where checkpoint phases will operate on sync state without requiring direct `DB` field access. The extraction also makes the data flow through sync/checkpoint methods explicit rather than implicit through struct field mutation.

Refs #1221

## How Has This Been Tested?

- `go build ./...`
- `go test ./... -count=1` — all pass
- `go test -race ./... -count=1` — all pass
- `pre-commit run --all-files` — all pass
- Existing `TestApplySyncResult` and `TestVerifyAndSync_DelaysStateMutationUntilApply` tests updated and passing

## Types of changes

## Checklist

### DIFF
--- a/cmd/litestream/main.go
+++ b/cmd/litestream/main.go
@@ -1468,6 +1468,10 @@ func NewS3ReplicaClientFromConfig(c *ReplicaConfig, _ *litestream.Replica) (_ *s
 	if isCloudflareR2 {
 		client.Concurrency = s3.DefaultR2Concurrency
 	}
+	if isTigris {
+		client.Concurrency = s3.DefaultTigrisConcurrency
+		client.PartSize = s3.DefaultTigrisPartSize
+	}
 
 	// Apply upload configuration if specified.
 	if c.PartSize != nil {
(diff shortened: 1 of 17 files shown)


corylanou:
Superseded by #1265, which reframes this work as a soak-driven sync/checkpoint stability fix instead of the original Phase 2 syncState refactor.

PR #1229, opened 2026-04-08
fix(restore): fix v3 restore

darkgnotic:
## Description

Fix logic that restores from wal segments to follow the logic in litestream v0.3.x restores:
* append all segments for a given index into a single file
* apply each index to the db in order ([ref](https://github.com/benbjohnson/litestream/blob/ad0a123938631cb081e090f15d01f10af72884e8/replica.go#L1203))

The previous restore-from-v3 logic was using a single tmp wal file, overwriting the previous one with each segment, and only applying the last one.

## Motivation and Context

Fixes v3 restore from wal files

## How Has This Been Tested?

Tested on multiple previously failing generations.

## Types of changes

## Checklist

### DIFF
--- a/replica.go
+++ b/replica.go
@@ -1177,7 +1177,7 @@ func (r *Replica) applyWALSegmentsV3(ctx context.Context, client ReplicaClientV3
 		}
 	}()
 
-	applyLastWalSegment := func() error {
+	applyLastWalFile := func() error {
 		if f == nil {
 			return nil
 		} else if err = f.Close(); err != nil {
@@ -1191,11 +1191,11 @@ func (r *Replica) applyWALSegmentsV3(ctx context.Context, client ReplicaClientV3
 	// to the db.
 	for _, seg := range segments {
 		if seg.Offset == 0 {
-			// Apply the last WAL segment, if any
-			if err = applyLastWalSegment(); err != nil {
+			// Apply the last WAL file, if any
+			if err = applyLastWalFile(); err != nil {
 				return err
 			}
-			// Open a new WAL segment
+			// Open a new WAL file
 			if f, err = os.OpenFile(walPath, os.O_CREATE|os.O_WRONLY, 0644); err != nil {
 				return err
 			}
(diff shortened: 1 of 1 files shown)


replica.go
```diff
@@ -1168,40 +1168,58 @@ func (r *Replica) applyWALSegmentsV3(ctx context.Context, client ReplicaClientV3
...
+		}
+		f = nil
+		return checkpointV3(dbPath)
+	}
+
+	// Reconstruct each wal file, appending non-zero offsets, and apply them
+	// to the db.
 	for _, seg := range segments {
-		if err := r.writeWALSegmentV3(ctx, client, generation, seg, walPath); err != nil {
+		if seg.Offset == 0 {
```
benbjohnson, line 1193: We may want to add a check that the first segment starts at offset zero, otherwise `writeWALSegmentV3()` will panic.

replica.go
```diff
@@ -1168,40 +1168,58 @@ func (r *Replica) applyWALSegmentsV3(ctx context.Context, client ReplicaClientV3
...
+			// Apply the last WAL file, if any
+			if err = applyLastWalFile(); err != nil {
+				return err
+			}
+			// Open a new WAL file
+			if f, err = os.OpenFile(walPath, os.O_CREATE|os.O_WRONLY, 0644); err != nil {
+				return err
+			}
```
benbjohnson, line 1201: I'd probably add an `os.O_TRUNC` as well but it should be a moot point as the next WAL will invalidate the salt.

replica.go
```diff
@@ -1168,40 +1168,58 @@ func (r *Replica) applyWALSegmentsV3(ctx context.Context, client ReplicaClientV3
...
-	// Seek to offset and write.
-	if _, err := f.Seek(seg.Offset, io.SeekStart); err != nil {
-		return err
-	}
-	if _, err := io.Copy(f, rc); err != nil {
-		return err
-	}
-	return f.Sync()
+	_, err = io.Copy(f, rc)
+	return err
```
benbjohnson, line 1222: It might be good to validate that the segments are contiguous otherwise you could have data loss if there's a gap in segments.

benbjohnson (approved):
I left some comments for some fixes but nothing critical enough to block merging. We can do the fixes in a separate PR.

darkgnotic:
> I left some comments for some fixes but nothing critical enough to block merging. We can do the fixes in a separate PR.

Thank you @benbjohnson!

I've submitted a PR with your suggested improvements here:  https://github.com/benbjohnson/litestream/pull/1230

PR #1230, opened 2026-04-14
fix(restore): check for wal segment gaps in v3 restore

darkgnotic:
## Description

Adds checks for missing segments during v3 restore

## Motivation and Context

Early detection of corruption when there are missing wal segment files when restoring from a v3 backup

Addresses feedback in https://github.com/benbjohnson/litestream/pull/1229

## How Has This Been Tested?

Deleted wal segments from local backup folder and verified that the restore produces the expected errors.

## Types of changes

## Checklist

### DIFF
--- a/replica.go
+++ b/replica.go
@@ -1170,6 +1170,8 @@ func (r *Replica) applyWALSegmentsV3(ctx context.Context, client ReplicaClientV3
 	walPath := dbPath + "-wal"
 	var f *os.File
 	var err error
+	index := 0
+	offset := int64(0)
 
 	defer func() {
 		if f != nil {
@@ -1184,7 +1186,11 @@ func (r *Replica) applyWALSegmentsV3(ctx context.Context, client ReplicaClientV3
 			return err
 		}
 		f = nil
-		return checkpointV3(dbPath)
+		if err = checkpointV3(dbPath); err != nil {
+			return err
+		}
+		r.Logger().Info(fmt.Sprintf("applied WAL index %d (%d bytes)", index, offset))
+		return nil
 	}
 
 	// Reconstruct each wal file, appending non-zero offsets, and apply them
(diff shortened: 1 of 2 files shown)


corylanou:
Thanks so much for jumping on this — the v3 restore work in #1229 plus this gap-detection follow-up is really appreciated, and the safety net you're adding here is exactly the right instinct. A few things I'd like to sort out before merge:

1. **Tests.** This needs coverage before it lands. `TestReplica_RestoreV3` in `replica_test.go` already has helpers (`createV3Backup`, `v3WALSegmentData`) that make it pretty straightforward. We need cases for: (a) missing-index gap error, (b) missing-offset gap error, (c) a successful restore across multiple WAL indices (which the existing suite doesn't actually exercise — that's part of why the new code paths are uncovered). Happy to add these myself if that's easier on your end — just let me know.

2. **Walk me through the `index = 0` starting state.** When segments begin at index 0 (the normal snapshot+WAL case), the check `seg.Index != index+1` at line 1204 looks like it would reject a valid first segment, since `index+1 == 1` but `seg.Index == 0`. Am I reading this right? A multi-index test would shake this out immediately.

3. **Undisclosed change: `O_TRUNC`.** Adding `O_TRUNC` to the WAL file open at line 1210 is a real fix (without it, a longer prior index's WAL would leak tail bytes into the next file), but it's not mentioned in the PR description. Could you call it out, and ideally have a test that proves it's correct?

4. **Minor polish:**
   - Error message style is inconsistent between the two new errors — pick one (Go convention is lowercase, no trailing period).
   - The new `Info` log fires once per WAL index per restore — consider `Debug` or a single summary line at the end. Also worth using structured fields (`"index", index, "bytes", offset`) to match the existing `Debug` call right below it.

Again, really grateful for this — happy to pair on the tests if that helps.

darkgnotic:
> Thanks so much for jumping on this — the v3 restore work in #1229 plus this gap-detection follow-up is really appreciated, and the safety net you're adding here is exactly the right instinct. A few things I'd like to sort out before merge:

Thank you for taking a look at this Cory!

> 
> 1. **Tests.** This needs coverage before it lands. `TestReplica_RestoreV3` in `replica_test.go` already has helpers (`createV3Backup`, `v3WALSegmentData`) that make it pretty straightforward. We need cases for: (a) missing-index gap error, (b) missing-offset gap error, (c) a successful restore across multiple WAL indices (which the existing suite doesn't actually exercise — that's part of why the new code paths are uncovered). Happy to add these myself if that's easier on your end — just let me know.

I strongly agree and would gladly accept your help here!  🙏 

> 2. **Walk me through the `index = 0` starting state.** When segments begin at index 0 (the normal snapshot+WAL case), the check `seg.Index != index+1` at line 1204 looks like it would reject a valid first segment, since `index+1 == 1` but `seg.Index == 0`. Am I reading this right? A multi-index test would shake this out immediately.

Hmmm ... yes, I understand the problem here. In the examples I worked off of, the snapshot and wal always start at index 1. 

However, I now see that when retention kicks in, a new snapshot + wal are created at a larger index. 

So perhaps the correct logic would be to pass in the `snapshot.Index` passed in to the preceding call to `downloadSnapshotV3()`, and also pass that into `applyWALSegmentsV3()`, with the expectation of that being the index of the first WAL segment.

Is that correct?


> 3. **Undisclosed change: `O_TRUNC`.** Adding `O_TRUNC` to the WAL file open at line 1210 is a real fix (without it, a longer prior index's WAL would leak tail bytes into the next file), but it's not mentioned in the PR description. Could you call it out, and ideally have a test that proves it's correct?
> 4. **Minor polish:**
>    
>    * Error message style is inconsistent between the two new errors — pick one (Go convention is lowercase, no trailing period).
>    * The new `Info` log fires once per WAL index per restore — consider `Debug` or a single summary line at the end. Also worth using structured fields (`"index", index, "bytes", offset`) to match the existing `Debug` call right below it.
> 
> Again, really grateful for this — happy to pair on the tests if that helps.

Likewise, thank you for taking a look. I would very much appreciate your help in pairing on the tests. I am happy and eager to learn.  🙏

corylanou:
Pushed `9145e8e` to `rocicorp/v3-restore-checks`.

What changed:
- `applyWALSegmentsV3` now starts expected WAL index checks from `snapshot.Index`, fixing non-zero snapshot restore cases.
- Added v3 restore coverage for multiple WAL indices, missing WAL index, missing WAL offset, non-zero snapshot index, and large-to-small WAL index restore.
- Normalized the new gap errors and changed the per-index apply log to structured Debug.
- Fixed the v3 WAL test helper to write filenames in the format the v3 file client actually parses.

Local verification passed:
- `go build ./...`
- `go test -race -run TestReplica_RestoreV3 -v ./...`
- `go vet ./...`
- `gofmt -l replica.go replica_test.go`

GitHub currently reports no checks for this PR branch (`statusCheckRollup` is empty).

darkgnotic:
Thank you @corylanou!

I've gathered a few more samples from our application, including ones with non-initial snapshot generations, and verified restore works correctly in the complete case, and that missing wal segments are detected when possible.  🙏

PR #1244, opened 2026-04-24
test(db): retry SQLITE_BUSY in checkpoint page-gap test

corylanou:
## Description
Replace the `db.Checkpoint` call in `TestDB_CheckpointPageGapWithConcurrentWrites` with a single attempt that `t.Skip`s on `SQLITE_BUSY` instead of failing.

## Motivation and Context
The test runs `db.Checkpoint(ctx, CheckpointModeTruncate)` concurrently with a writer goroutine. `wal_checkpoint(TRUNCATE)` can legitimately return `SQLITE_BUSY` while a writer is actively committing — on a slow CI runner this caused the test to fail with `database is locked (5) (SQLITE_BUSY)` even though no actual page loss occurred.

An earlier revision of this PR retried `db.Checkpoint` on BUSY. Code review flagged that `db.Checkpoint` is not just the raw `wal_checkpoint` pragma — it also runs an internal `verifyAndSync` before each checkpoint, which uploads L0 files. Retrying therefore widens page coverage across attempts and could mask the very TOCTOU page-gap bug this test is designed to detect. Skipping on BUSY is the honest response: no checkpoint occurred, so the invariant cannot be validated this iteration.

Observed flake: [run 24842765150](https://github.com/benbjohnson/litestream/actions/runs/24842765150) — `TestDB_CheckpointPageGapWithConcurrentWrites` failed at `db_internal_test.go:1902` with `SQLITE_BUSY` on a post-merge main build, after the same code passed in the PR's own CI.

## How Has This Been Tested?
- `go test -run TestDB_CheckpointPageGapWithConcurrentWrites -count=20 -race ./.` — all 20 iterations pass.
- `go vet ./...` clean.
- Only `SQLITE_BUSY` / `database is locked` triggers `t.Skip`; any other error still fails the test.

## Types of changes

## Checklist

### DIFF
--- a/db_internal_test.go
+++ b/db_internal_test.go
@@ -1896,29 +1896,21 @@ func TestDB_CheckpointPageGapWithConcurrentWrites(t *testing.T) {
 	case err := <-writerDone:
 		t.Fatalf("writer exited before starting: %v", err)
 	}
-	// Retry SQLITE_BUSY on the checkpoint just as the writer does — wal_checkpoint(TRUNCATE)
-	// can legitimately return BUSY while a writer is actively committing on a slow runner,
-	// and a BUSY return means no checkpoint happened so there is nothing to verify.
-	checkpointDeadline := time.Now().Add(30 * time.Second)
-	var checkpointErr error
-	for {
-		checkpointErr = db.Checkpoint(ctx, CheckpointModeTruncate)
-		if checkpointErr == nil {
-			break
-		}
-		if !strings.Contains(checkpointErr.Error(), "database is locked") &&
-			!strings.Contains(checkpointErr.Error(), "SQLITE_BUSY") {
-			break
-		}
-		if time.Now().After(checkpointDeadline) {
-			break
-		}
-		time.Sleep(time.Millisecond)
-	}
-	if checkpointErr != nil {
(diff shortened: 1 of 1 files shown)


corylanou:
Closing this in favour of #1410.

Measurement, not redundancy, is the reason. With a page gap injected (`delete(allPages, 42)`) and the test run 100 times:

- `main` today (retry on BUSY, from #1292): **100 fail / 100** — catches the regression every run.
- This PR (single attempt, `t.Skip` on BUSY): **14 fail, 86 skip / 100**.

BUSY fires on roughly 86% of runs, so the `t.Skip` path would leave the assertion unexecuted on most of them and turn a working regression guard into a green check that means almost nothing.

The concern raised here — that retrying performs extra pre-checkpoint sync and uploads additional L0 files (`db.go:2402`, `db.go:2478`, `db.go:2520`), widening page coverage enough to mask a gap — did not reproduce. `main` caught the injected gap on every run. Worth noting the injection removes a page from the expected set, so it proves the assertion is live under retry but does not fully exercise the widening hypothesis; #1410 records that caveat.

The real finding is the 86% BUSY rate itself. #1410 tracks removing the contention so the assertion runs every time, rather than tolerating BUSY through either retries or skips.

Issue #1232, opened 2026-04-22
CLI: agent-friendly usability improvements (tracking issue)


corylanou:
## Summary

The Litestream CLI is mostly agent-friendly (non-interactive, flag-driven, some structured output), but a recent review surfaced several gaps that make it harder for AI agents (and scripts) to use it reliably. This tracks each improvement so they can be landed as small, independent PRs.

Inspired by [this thread on building CLIs for agents](https://x.com/ericzakariasson/status/2036762680401223946) — the guidance there maps almost one-to-one onto the items below.

## Goals

- Consistent structured output (JSON everywhere read-only)
- Safe defaults on destructive paths (`-dry-run`)
- Self-documenting `-help` on every subcommand (examples included)
- Predictable success payloads an agent can chain into the next command
- Cleaner error messages that suggest the corrected invocation

Non-goals: breaking existing invocations, rewriting the command tree, adding interactive TUI.

## Task list

Each checkbox is scoped to a single small PR. They can be taken in any order unless a dependency is noted.

### Safety

- [ ] **Add `-dry-run` to `restore`** — print the plan (source, target path, txid range, files to fetch) and exit 0 without writing.
- [ ] **Add `-dry-run` to `reset`** — list LTX files that would be removed from the metadata directory, don't remove them.
- [ ] **Add `-dry-run` to `unregister`** — show what would be torn down (DB path, replicas, final sync behavior).
- [ ] **Guard `restore` overwrites** — when the output path exists and differs from an empty file, require `-force` (keep current behavior under `-force`). Applies only to interactive/default runs; `-if-db-not-exists` paths unchanged.

### Structured output

- [ ] **Add `-json` to `status`** — emit one object per database with `database`, `status`, `local_txid`, `wal_size`.
- [ ] **Add `-json` to `databases`** — emit the configured database list as JSON.
- [ ] **Add `-json` to `ltx`** — emit LTX file metadata (level, min/max txid, size, timestamp) as JSON array.
- [ ] **Add `-json` to `restore`** — emit a final summary object (`db_path`, `replica`, `txid`, `duration_ms`, `integrity_check`) on success.
- [ ] **Document `-json` output schemas** in `docs/` so schemas are stable for downstream consumers.

### Success payloads (chainability)

- [ ] **`register` prints confirmation on success** — `db_path`, `replica`, `socket`. JSON under `-json`.
[...]

PR #1254, opened 2026-04-24
feat(cli): guard restore overwrites

corylanou:
## Summary
- add -force to litestream restore for intentional overwrite of an existing output database
- require -force for non-empty output paths and existing SQLite sidecar files
- allow empty placeholder output files by removing them before the restore; keep follow mode and -if-db-not-exists behavior unchanged

## Testing
- go test -run 'TestRestoreCommand_RunRequiresForceForExistingOutput|TestRestoreCommand_RunForceOverwritesExistingOutput|TestRestoreCommand_RunAllowsEmptyOutput|TestRestoreCommand_RunJSONOutput|TestRestoreCommand_RunDryRunJSONOutput' ./cmd/litestream
- go test ./...
- pre-commit run --all-files

Related to #1232
Stacked on #1253

### DIFF
--- a/cmd/litestream/restore.go
+++ b/cmd/litestream/restore.go
@@ -34,6 +34,7 @@ func (c *RestoreCommand) Run(ctx context.Context, args []string) (err error) {
 	ifReplicaExists := fs.Bool("if-replica-exists", false, "")
 	timestampStr := fs.String("timestamp", "", "timestamp")
 	dryRun := fs.Bool("dry-run", false, "print restore plan without writing")
+	force := fs.Bool("force", false, "overwrite existing output database")
 	jsonOutput := fs.Bool("json", false, "output raw JSON")
 	fs.BoolVar(&opt.Follow, "f", false, "follow mode")
 	fs.DurationVar(&opt.FollowInterval, "follow-interval", opt.FollowInterval, "polling interval for follow mode")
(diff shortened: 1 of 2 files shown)


cmd/litestream/restore.go
```diff
@@ -255,6 +262,37 @@ func (c *RestoreCommand) restoreTXID(ctx context.Context, r *litestream.Replica,
...
+		} else if err != nil && !os.IsNotExist(err) {
+			return fmt.Errorf("cannot access SQLite sidecar path: %w", err)
+		}
+	}
+
+	for _, removePath := range []string{path, path + "-wal", path + "-shm"} {
+		if err := os.Remove(removePath); err != nil && !os.IsNotExist(err) {
+			return fmt.Errorf("remove existing output path: %w", err)
+		}
+	}
```
benbjohnson: It's probably worth adding a `-journal` suffix on here just in case

benbjohnson (approved):

cmd/litestream/restore.go
corylanou: Good call — added `-journal` to both the sidecar check and the cleanup loop in 850689d. Also extended the existing -force overwrite test to cover the journal file and added a new test that asserts the guard refuses when a -journal exists without -force.

Issue #1273, opened 2026-05-15
Split soak-stabilization PR into reviewable fixes


corylanou:
## Summary

PR #1265 has become a validated integration branch for the soak-stabilization work, but it is too large to review and merge directly. Split it into focused, reviewable PRs from `origin/main`.

## Context

- Integration PR: #1265
- Integration branch: `fix-soak-sync-checkpoint-stability`
- Validated head: `51dda8e9301a324a27bad9ac27c332d46c7f790f`
- Live soak check on 2026-05-15 showed PR-1265 workers running for roughly 27 hours with no active failure clusters and zero recent failed verifications in the incident windows.

## Reviewable Work

- [x] #1274
- [x] #1275
- [x] #1276
- [x] #1277
- [x] #1278
- [x] #1279
- [x] #1280
- [x] #1281

## Working Rules

- Start each child PR from `origin/main`, not from the full integration branch.
- Keep each PR focused on one production concern with its own tests.
- Reference #1265 and the validated head SHA in each child PR as soak evidence.
- Do not replay the investigation history one commit at a time; fold intermediate experiments and rollbacks into coherent final changes or drop them.
- Keep #1265 as the reference branch until the split PRs have landed or been intentionally superseded.

## Related

- Supersedes direct review of #1265 as a merge target.
- Related to #1221 and #1228.

corylanou:
## 2026-05-18 soak triage note

Live control-plane diagnosis around 2026-05-18T13:53Z shows one current PR-1265 degraded worker: `worker-pr-1265-gharchive-mixed`. The fleet headline also includes old dormant main/PR-1228 failures, so future triage should compare by `source`, `git_sha`, and `litestream_sha` before treating the aggregate headline as current PR-1265 signal.

Current PR-1265 failure evidence:

- Worker: `worker-pr-1265-gharchive-mixed`, profile `gharchive-mixed`, source `pr-1265`, Litestream `51dda8e9301a324a27bad9ac27c332d46c7f790f`, status `degraded`.
- Latest verification failed from 2026-05-18T13:31:42Z to 2026-05-18T13:36:47Z after ~304s with `sync database: replica sync: context deadline exceeded`.
- Runtime snapshot was healthy and disk was not constrained: ~52.5GB total, ~48.1GB free, ~8.3% used, DB ~2.3GB, WAL ~1.86GB.
[...]

PR #1265, opened 2026-05-06
integration: soak-stabilization reference branch

corylanou:
## Status

This PR is now a draft integration/reference branch, not the final reviewable merge PR.

The branch captures the full soak-stabilization stack that was validated together. The work is being split into smaller PRs tracked by #1273 so each production change can be reviewed and merged independently.

## Current Soak Evidence

- Validated head: `51dda8e9301a324a27bad9ac27c332d46c7f790f`
- Live control-plane check on 2026-05-15 showed PR-1265 workers running for roughly 27 hours.
- No active failure clusters were reported by `/api/diagnosis`.
- PR-1265 incident windows showed zero recent failed verifications across the checked workers.
- CI for this PR head is green.

## Split Tracking

Parent issue: #1273

Focused child issues:

- #1274
- #1275
- #1276
- #1277
- #1278
- #1279
- #1280
- #1281

## How To Use This PR

Use this PR as reference material and soak evidence only. New implementation PRs should start from `origin/main`, not from this branch, and should re-author only the scoped change needed for the corresponding issue.

Do not replay the investigation history one commit at a time. Intermediate experiments and rollbacks should be folded into coherent final changes or dropped.

## Original Problem Summary

The soak system exposed several recurring failure families while exercising the prior stabilization work:

- Foreground `/sync` calls could time out behind DB sync/checkpoint executor work.
- Restore/integrity failures appeared around WAL/checkpoint boundary handling and page-growth coverage.
- S3-compatible storage showed request-canceled list failures and Tigris upload-default issues.
- A failed restore could leave `/data/test.db.restored.tmp`, consuming worker disk and causing subsequent checkpoint/sync failures.

## Local Validation Run On This Branch

```bash
go test ./... -run 'TestReplica_Restore_RemovesTempFileOnFailure|TestReplica_Restore_InvalidFileSize' -count=1
go test ./... -count=1
pre-commit run --all-files
```

Additional targeted validation previously run on this branch:

```bash
go test . -run 'TestDB_Snapshot(ExcludesUnsyncedWALFrames|SkipsExistingTXID)?$' -count=1
go test -race . -run TestDB_SnapshotExcludesUnsyncedWALFrames -count=1
```

PR #1286, opened 2026-05-18
fix(compactor): close pipe on upload failure

### LINKED ISSUE #1277: Fix compactor upload failure pipe cleanup
## Summary

A compactor upload failure path can leave the producer side blocked on a pipe. Split this provider-agnostic compactor cleanup into its own small PR.

## Scope

- Close/cleanup the pipe correctly when upload fails.
- Keep the change limited to compactor behavior and tests.

## Starting point

Use PR #1265 commit `8e2d1865cd4c69a3077e7ba3902db2601c76eee7` as the reference, re-applied from `origin/main`.

## Validation

- Focused compactor regression test.
- `go test ./... -run Compactor -count=1` if practical.

## Related

- Parent: #1273
- Integration soak reference: #1265

corylanou:
## Description
Close the compaction pipe reader when `WriteLTXFile` returns an upload/write error, then wait for the compactor goroutine to exit before returning. This prevents the producer side from remaining blocked on a pipe after an early upload failure.

## Motivation and Context
The soak split found a provider-agnostic compactor cleanup path that can leak a blocked producer goroutine when the destination upload fails before consuming the compacted LTX stream. This is extracted from #1265 as an isolated compactor fix.

Fixes #1277
Related: #1273, #1265

## How Has This Been Tested?
- `go test . -run TestCompactor_CompactClosesPipeOnWriteError -count=1`
- `go test ./... -run Compactor -count=1`
- `git diff --check`
- `go test ./... -count=1`

## Types of changes

## Checklist

### DIFF
--- a/compactor.go
+++ b/compactor.go
@@ -154,29 +154,20 @@ func (c *Compactor) Compact(ctx context.Context, dstLevel int) (*ltx.FileInfo, e
 	}
 
 	pr, pw := io.Pipe()
-	compactErrCh := make(chan error, 1)
 	go func() {
 		comp, err := ltx.NewCompactor(pw, rdrs)
 		if err != nil {
-			err = fmt.Errorf("new ltx compactor: %w", err)
-			_ = pw.CloseWithError(err)
-			compactErrCh <- err
+			_ = pw.CloseWithError(fmt.Errorf("new ltx compactor: %w", err))
 			return
 		}
 		comp.HeaderFlags = ltx.HeaderFlagNoChecksum
-		err = comp.Compact(ctx)
-		_ = pw.CloseWithError(err)
-		compactErrCh <- err
+		_ = pw.CloseWithError(comp.Compact(ctx))
 	}()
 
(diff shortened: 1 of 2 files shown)


compactor.go
```diff
@@ -154,20 +154,30 @@ func (c *Compactor) Compact(ctx context.Context, dstLevel int) (*ltx.FileInfo, e
...
 		comp.HeaderFlags = ltx.HeaderFlagNoChecksum
-		_ = pw.CloseWithError(comp.Compact(ctx))
+		err = comp.Compact(ctx)
+		_ = pw.CloseWithError(err)
+		compactErrCh <- err
 	}()
 
 	info, err := c.client.WriteLTXFile(ctx, dstLevel, minTXID, maxTXID, pr)
+	_ = pr.CloseWithError(err)
+	compactErr := <-compactErrCh
```
benbjohnson: Is the fix to close the `pr` side? The `err` passed to `pw.CloseWithError()` should come through when reading from the `pr` side, right? Do we really need the extra channel?

corylanou: you are correct.  fixed.

benbjohnson (approved):

Issue #1280, opened 2026-05-15
Stabilize DB sync/checkpoint executor under soak load


corylanou:
## Summary

The original soak failures included foreground sync calls timing out behind checkpoint/sync executor work. Re-author the DB sync/checkpoint executor changes into one or more reviewable PRs.

## Scope

- Make sync/checkpoint state publication explicit and safe.
- Bound sync executor work so monitor work cannot starve foreground sync.
- Keep locking/context behavior reviewable.
- Split into stacked PRs if the first implementation is still too large.

## Suggested PR split

1. Mechanical sync state extraction if it still reduces the final diff.
2. Executor lifecycle/fairness and bounded work.
3. Any remaining starvation fixes.

## Starting point

Use the relevant DB sync/executor commits from PR #1265 as reference material, but do not replay the investigation history blindly. Fold rollback/intermediate semaphore commits into the final coherent implementation or drop them.

## Validation

- Focused DB sync/checkpoint tests.
- Race test for the highest-risk concurrency path.
- Reference the clean #1265 soak result.

## Related

- Parent: #1273
- Integration soak reference: #1265

corylanou:
2026-05-18 soak triage update: `worker-pr-1265-gharchive-mixed` is the current relevant degraded PR-1265 worker. Latest failure ran from 2026-05-18T13:31:42Z to 13:36:47Z and failed sync after ~304s with `sync database: replica sync: context deadline exceeded`.

Evidence points at this issue first: disk was healthy (~48GB free), Litestream process/socket were responding, `/sync-status` showed a large WAL before sync (~1.86GB) and a much smaller WAL after timeout (~610KB), and goroutines showed `DB.sync`/`Fsync` plus concurrent compaction upload work. This looks like foreground verification sync interacting poorly with background sync/replica/compaction work or timeout boundaries under GH Archive mixed load.

corylanou:
Fly re-auth follow-up: `worker-pr-1265-gharchive-mixed` had a later OOM restart at 2026-05-18T09:31:35Z (`exit_code=1, oom_killed=true`), but the first sustained failures began earlier at 2026-05-17T19:33Z with `db sync: wait for db sync executor: context deadline exceeded`.

The sequence matters for this issue: this appears to start as sync executor starvation/timeout under GH Archive mixed load, then degrades into replica-sync/S3 upload deadline failures, with an isolated OOM restart later in the run. Other PR-1265 machines did not show restart/OOM events in Fly machine status.

[...]

PR #1287, opened 2026-05-18
feat(db): expose sync diagnostics

### LINKED ISSUE #1279: Add sync diagnostics for soak and control-plane triage
## Summary

The soak incident was much easier to diagnose with sync phase/status information. Add production-safe diagnostics separately from correctness fixes so it can be reviewed on observability merits.

## Scope

- Expose DB sync status diagnostics used by the soak control plane.
- Include server/API tests.
- Keep this limited to observability and avoid behavior changes where possible.

## Starting point

Use PR #1265 commit `0ce028650245c3ea40c3b96d81710085b276f4a3` as the reference, re-applied from `origin/main`.

## Validation

- Sync-status/server tests.
- `go test ./... -run SyncStatus -count=1` if practical.

## Related

- Parent: #1273
- Integration soak reference: #1265

corylanou:
## Summary
Expose the latest database sync/checkpoint diagnostic state through the existing Unix control socket at `GET /debug/sync-status`.

## Problem
The PR #1265 soak incident required bounded failure snapshots to identify where sync was stuck. The useful signal was the current DB operation, phase, WAL size, snapshotting state, and last synced WAL offset. Without a runtime endpoint, the control plane has to infer this from repeated failures or ask for ad-hoc worker access.

## Solution
Add a protected diagnostic snapshot on `DB` and update it around existing sync/checkpoint phases. Expose all database diagnostics through the control socket, with optional `path` filtering.

## Scope
**In scope:**
- Sync/checkpoint phase diagnostics on `DB`
- `GET /debug/sync-status` control socket endpoint
- DB/server tests for the diagnostic response

**Not in scope:**
- Sync executor fairness or starvation fixes
- Memory/OOM mitigation
- Any change to sync, checkpoint, compaction, or restore behavior

## Test Plan
```bash
go test ./... -run 'SyncDiagnostic|SyncStatus|HandleSyncStatus' -count=1
go test ./... -count=1
git diff --check
```

Commit hooks also ran `goimports`, `go vet`, and `staticcheck`.

## Related
- Fixes #1279
- Parent: #1273
- Related to #1265
- Root-cause soak follow-up remains in #1280

### DIFF
--- a/db.go
+++ b/db.go
@@ -198,11 +198,45 @@ type DB struct {
 	Logger *slog.Logger
 }
 
+type syncDiagnosticOperation string
+
+const (
+	syncDiagnosticOperationSync       syncDiagnosticOperation = "sync"
+	syncDiagnosticOperationCheckpoint syncDiagnosticOperation = "checkpoint"
+)
+
+type syncDiagnosticPhase string
+
+const (
+	syncDiagnosticPhaseStarting                       syncDiagnosticPhase = "starting"
+	syncDiagnosticPhaseEnsureWAL                      syncDiagnosticPhase = "ensure_wal"
+	syncDiagnosticPhaseVerifyAndSync                  syncDiagnosticPhase = "verify_and_sync"
+	syncDiagnosticPhaseCheckpointIfNeeded             syncDiagnosticPhase = "checkpoint_if_needed"
+	syncDiagnosticPhaseUpdateMetrics                  syncDiagnosticPhase = "update_metrics"
+	syncDiagnosticPhaseVerify                         syncDiagnosticPhase = "verify"
+	syncDiagnosticPhaseSyncOpenLTX                    syncDiagnosticPhase = "sync_open_ltx"
+	syncDiagnosticPhaseSyncPageMap                    syncDiagnosticPhase = "sync_page_map"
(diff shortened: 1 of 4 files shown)


db.go
```diff
@@ -276,6 +311,114 @@ func (db *DB) Path() string {
...
+	return func(s *syncDiagnosticState) { s.snapshotting = v }
+}
+
+func syncDiagnosticCheckpointMode(mode string) syncDiagnosticOption {
+	return func(s *syncDiagnosticState) { s.checkpointMode = mode }
+}
+
+func syncDiagnosticReason(reason string) syncDiagnosticOption {
+	return func(s *syncDiagnosticState) { s.reason = reason }
+}
```
benbjohnson: These helper functions seem like overkill. Can we just pass a `func(*syncDiagnosticState)` to `setSyncDiagnosticPhase()` and have each one update whichever fields they want?

corylanou: Addressed in 40e1044: removed the diagnostic option helpers, updated phase changes to use inline `func(*syncDiagnosticState)` callbacks, and added internal typed constants for diagnostic operation/phase values while keeping the public diagnostic JSON fields as strings. CI is green.

benbjohnson (approved):

PR #1311, opened 2026-06-13
fix(compaction): resume remote source reads

### LINKED ISSUE #1308: Compaction reads source LTX files over raw S3 streams; idle-disconnect mid-compact fails with unexpected EOF
> ## ⚠️ Implementation contract — read before writing any code
>
> This issue is part of a **stacked PR series**. Do NOT base work on `main`.
>
> - **Canonical map (read first):** https://gist.github.com/corylanou/f2709d9868cad1943c63096bf978e881 — review order, scope of every PR, current blockers.
> - **Current stack (bottom-up):** #1289 → #1290 → #1291 → #1292 → #1166 → #1302 → #1305 → #1306.
> - **Base your branch on:** `fix-restore-throttle-retry` (head of #1306, the current stack top). This issue is intended as **order 9** — it lands BELOW the #1307 fix, so do this one first.
> - **Open the PR with `--base fix-restore-throttle-retry`**, never `--base main`. Title format matches the stack: `fix(...): ...`. Include the stack-map blockquote at the top of the PR body (copy it from #1306's description).
> - **After the PR opens:** the gist map must be updated (new row, mermaid node, review-order entry) and the soak fleet re-dispatched to the new stack top via litestream-soak's `scripts/start-pr-soak.sh <PR>`. If you don't have access to do those, say so in the PR rather than skipping silently.
> - **Scope guard:** one change site (`compactor.go:146` — wrap S3-sourced readers in `internal.NewResumableReader`, size from `info.Size`) plus a regression test with a stub opener that drops the stream mid-read and asserts compaction completes. The `LocalFileOpener` fast path is untouched. Do not refactor anything else.
> - **Validation:** test must fail on the unfixed code.

## Summary

`Compactor.compact` opens its source LTX files with a raw `client.OpenLTXFile` (compactor.go:146) and streams them through `ltx.NewCompactor` into a (potentially multi-minute) multipart upload. When the storage provider closes an idle source stream mid-compaction — the exact behavior `internal.ResumableReader` was added to the *restore* path for (#992 class) — the compaction fails:

```
level=ERROR msg="compaction failed" system=store db=test.db level=2
error="write ltx file: s3: upload to .../0002/0000000000000001-00000000000019a0.ltx:
upload multipart failed, ... cause: read multipart upload data failed,
read page header 0: unexpected EOF"
```

Observed in production soak testing on the current PR stack head (`fix-restore-throttle-retry`, d208afa9e153 — [stack map](https://gist.github.com/corylanou/f2709d9868cad1943c63096bf978e881)); compactor.go:146 is identical on `main`, so pre-existing rather than stack-introduced.

## Impact

Self-healing (the next compaction cycle retries) but each failure wastes a full multipart upload, delays level progression, and under sustained provider flakiness can starve higher levels — on the failing worker, level 0002 stayed empty while L1 accumulated 200+ files, which lengthens restore plans and restore times (compounding benbjohnson/litestream#1307).

## Suggested fix

Wrap the S3-sourced readers in `internal.NewResumableReader` (file size is available from `info.Size`), symmetrical with the restore path in `replica.go Restore()`. The `LocalFileOpener` fast path is unaffected. Worth noting: compaction sources are also subject to retention deletion mid-compact, so the NotExist-is-terminal semantics of the resumable reader remain correct here; the reader only needs to heal transport-level disconnects.

Happy to take this in the stack as a small follow-up PR.

Found by the litestream-soak production harness during the same incident window as #1307.

corylanou:

## Summary
Wrap remote compaction source readers in `internal.ResumableReader` so a provider-side idle disconnect can reopen from the last byte offset. The `LocalFileOpener` fast path remains unchanged.

## Problem
`Compactor.Compact` opened remote source LTX files as raw `OpenLTXFile` streams before sending them through `ltx.NewCompactor`. If the provider closed an idle source stream mid-compact, compaction failed even though the source object still existed and range reads could recover it.

The regression test demonstrates the failure by returning a source reader that cuts off after 16 bytes. On the unfixed code it fails with:

```text
write ltx file: extract timestamp from LTX header: EOF
```

## Solution
The remote fallback path now wraps the opened source stream with `internal.NewResumableReader`, passing `info.Size` from the listed LTX metadata. Premature EOF and transient read errors reopen through `OpenLTXFile` at the current offset.

## Scope
**In scope:**
- Remote compaction source reads
- Regression coverage for a mid-read disconnect and non-zero-offset reopen

**Not in scope:**
- Local compaction reads via `LocalFileOpener`
- Retry/backoff policy changes
- Storage backend behavior changes

Fixes #1308

## Test Plan
- `go test -run TestCompactor_CompactResumesRemoteSourceAfterDisconnect ./`
- `go test -run 'TestCompactor' ./`
- `go build ./...`
- `go test ./...`
- `golangci-lint run --new-from-rev=fix-restore-throttle-retry`
- `go test -race ./...`

`golangci-lint run` against the full stack still reports existing baseline errcheck/staticcheck findings outside this patch; the new-code check above is clean.

## Types of changes

## Checklist
- [x] My code follows the code style of this project (`gofmt`, repo import hook)
- [x] I have tested my changes (`go test ./...`, `go test -race ./...`)
- [ ] I have updated the documentation accordingly (not needed)

### DIFF
--- a/compactor.go
+++ b/compactor.go
@@ -10,6 +10,8 @@ import (
 
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/superfly/ltx"
+
+	"github.com/benbjohnson/litestream/internal"
 )
 
 // Compactor handles compaction and retention for LTX files.
@@ -147,7 +149,7 @@ func (c *Compactor) Compact(ctx context.Context, dstLevel int) (*ltx.FileInfo, e
 		if err != nil {
 			return nil, fmt.Errorf("open ltx file: %w", err)
 		}
-		rdrs = append(rdrs, f)
+		rdrs = append(rdrs, internal.NewResumableReader(ctx, c.client, info.Level, info.MinTXID, info.MaxTXID, info.Size, f, c.logger))
 	}
 	if len(rdrs) == 0 {
 		return nil, ErrNoCompaction
--- a/compactor_test.go
+++ b/compactor_test.go
@@ -119,6 +119,35 @@ func TestCompactor_CompactClosesPipeOnWriteError(t *testing.T) {
 	}
 }
 
+func TestCompactor_CompactResumesRemoteSourceAfterDisconnect(t *testing.T) {
+	client := newDisconnectingCompactionClient(t.TempDir(), 16)
+	compactor := litestream.NewCompactor(client, slog.Default())
+
+	createTestLTXFile(t, client, 0, 1, 1)
+
+	info, err := compactor.Compact(context.Background(), 1)
+	if err != nil {
+		t.Fatal(err)
+	}
+	if info.Level != 1 {
+		t.Errorf("Level=%d, want 1", info.Level)
+	}
+	if info.MinTXID != 1 || info.MaxTXID != 1 {
+		t.Errorf("TXID range=%d-%d, want 1-1", info.MinTXID, info.MaxTXID)
+	}
+
+	var resumed bool
+	for _, offset := range client.openOffsets[1:] {
+		if offset > 0 {
+			resumed = true
+			break
+		}
+	}
+	if !resumed {
+		t.Fatalf("OpenLTXFile offsets=%v, want reopen at non-zero offset", client.openOffsets)
+	}
+}
+
 func TestCompactor_MaxLTXFileInfo(t *testing.T) {
 	t.Run("WithFiles", func(t *testing.T) {
 		client := file.NewReplicaClient(t.TempDir())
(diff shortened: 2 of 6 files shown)


corylanou:
Soak dispatch note:

- `scripts/start-pr-soak.sh 1311` built and pushed the PR image successfully:
  `registry.fly.io/litestream-soak:sha-59d04a899442-pr-1311-ls-10df097f56b8`
- The final `notify-deployment-ready.sh` step did not run because this environment does not have `SOAK_ADMIN_BEARER_TOKEN` or `SOAK_BASIC_AUTH_USERNAME` / `SOAK_BASIC_AUTH_PASSWORD` set, and I did not find a local `.env` / `.envrc` to source.
- To finish the dispatch from an environment with soak admin credentials:

```bash
SOAK_ADMIN_BEARER_TOKEN=... /Users/corylanou/projects/corylanou/litestream-soak/scripts/notify-deployment-ready.sh \
  59d04a8994423ec54681edd339591606ee5af770 \
  pr-1311 \
  manual_pr_soak \
  registry.fly.io/litestream-soak:sha-59d04a899442-pr-1311-ls-10df097f56b8 \
  10df097f56b80c36d9001b9f6516d05096cea443
```

corylanou:
Merged to main as #1326 via extraction; the patch there is byte-identical to this branch's commit (4eb72ef), so nothing remains to review here.

PR #1312, opened 2026-06-13
fix(retention): retain prior snapshot txid

### LINKED ISSUE #1307: Restore fails mid-flight when compaction retention deletes LTX files from its plan
> ## ⚠️ Implementation contract — read before writing any code
>
> This issue is part of a **stacked PR series**. Do NOT base work on `main`.
>
> - **Canonical map (read first):** https://gist.github.com/corylanou/f2709d9868cad1943c63096bf978e881 — review order, scope of every PR, current blockers.
> - **Current stack (bottom-up):** #1289 → #1290 → #1291 → #1292 → #1166 → #1302 → #1305 → #1306.
> - **Base your branch on:** `issue-1308-compaction-reads-source-ltx-files-over-raw-s3-streams-idle-d` (head of #1311, the order-9 branch — #1308 has landed as the current stack top). This is the branch to fork from; do NOT use `fix-restore-throttle-retry` or `main`.
> - **Open the PR with `--base issue-1308-compaction-reads-source-ltx-files-over-raw-s3-streams-idle-d`**, never `--base main`. Title format matches the stack: `fix(...): ...`. Include the stack-map blockquote at the top of the PR body (copy it from #1306's description).
> - **After the PR opens:** the gist map must be updated (new row, mermaid node, review-order entry) and the soak fleet re-dispatched to the new stack top via litestream-soak's `scripts/start-pr-soak.sh <PR>`. If you don't have access to do those, say so in the PR rather than skipping silently.
> - **Sequencing note:** #1308 should land below this one (it shortens restores and narrows this race, giving cleaner validation data). Preferred fix direction here is **option 2** (retain one extra snapshot generation) per discussion; confirm in a comment before implementing if choosing differently.
> - **Validation:** regression test must fail on the unfixed code (run retention while a plan-holding reader is mid-file; assert the plan file survives / restore completes). The live reproducer is soak worker `worker-pr-1306-high-vol` (restore ~16min vs 10min snapshot interval).

## Summary

A restore that runs longer than the snapshot/retention cadence can fail with `reopen ltx file at offset N: file does not exist`: the restore computes its LTX plan once, decodes for many minutes over lazy/resumable S3 streams, and retention deletes plan files out from under it when a new snapshot raises the TXID watermark mid-restore. Restore is the operation that must not fail — a customer restoring a busy database hits this race in direct proportion to database size.

Observed in production soak testing on the current PR stack head (`fix-restore-throttle-retry`, d208afa9e153 — see the [stack map](https://gist.github.com/corylanou/f2709d9868cad1943c63096bf978e881)); the relevant code is identical on `main` (`CalcRestorePlan` one-shot, watermark retention), so this is pre-existing in the v0.5 LTX architecture rather than introduced by the stack.

## Production evidence (soak worker, 2026-06-12)

- `high-volume` profile, ord, 856MB db + 385MB WAL, `snapshot_interval=10m`, restore-validate durations grown to 15-19 minutes as the db grew.
- Failing check 16:00:31→16:16:41Z:
  ```
  Error: decode database: decode page 215837: read page header 10:
  reopen ltx file at offset 9420800: file does not exist
  ```
  The reader was 9.4MB into a file that existed at plan time — deletion-during-read, not a missing object at planning.
- Worker logs show compaction activity inside the restore window (16:14:01), and the post-failure prefix listing shows level 0002 empty while L1 holds 204 files — superseded files were retention-deleted during the restore.
- Identical failure shape on a second heavy worker the prior day (`reopen ltx file at offset…` on a level=2 file, 16-minute restore).
- The immediately preceding check (15-minute restore) passed: the race is probabilistic, window = restore duration vs snapshot/retention cadence. Once restore time exceeds the snapshot interval, every restore overlaps at least one retention cycle.

## Code path

- `replica.go` `Restore()`: `CalcRestorePlan(...)` once → one `internal.NewResumableReader` per plan file (lazy open) → ltx decode reads page-ordered across files for the full restore duration. No re-plan, no pinning.
- `internal/resumable_reader.go`: on reopen, `os.ErrNotExist` is (correctly) terminal — retrying a deleted object cannot succeed.
- `store.go` `EnforceSnapshotRetention()` → `db.EnforceSnapshotRetention` + `Compactor.EnforceRetentionByTXID`: after a new snapshot lands, every level deletes files with `MaxTXID < minSnapshotTXID`. No awareness of in-flight restores, no time-based grace on lower levels.

## Options (decision needed)

1. **Retention grace window**: don't delete superseded LTX files until N minutes after they fall below the watermark (N ≥ a generous max-restore-time, or configurable). Simplest; bounds the race without coordination.
2. **Keep one extra snapshot generation's history**: retain files reachable from the previous snapshot, not just the latest — any plan computed against an existing snapshot stays restorable until the next-next snapshot.
3. **Re-plan and resume on NotExist**: restore catches the NotExist, recomputes the plan, and resumes — robust but complex, since the ltx decoder reads page-ordered across all files and mid-decode substitution is non-trivial.

Option 1 or 2 look like the right cost/benefit; 3 could layer on later. Happy to take this in the stack as a follow-up PR once direction is agreed.

Found by the litestream-soak production harness (30-minute restore-verification cycles under sustained load).

corylanou:

## Description
Retains one extra snapshot generation for lower-level LTX retention by deriving the cascade watermark from the snapshot immediately before the first retained snapshot when one exists. This keeps LTX files already selected by an in-flight restore plan from being deleted as soon as the next snapshot raises the latest snapshot TXID.

The change is localized to `DB.EnforceSnapshotRetention()` and the existing store propagation path to lower-level `EnforceRetentionByTXID()` calls. The snapshot-level "always keep at least one snapshot" guard remains intact.

Stack gist and soak dispatch are intentionally not touched here; those are handled separately from this PR workflow.

## Motivation and Context
A long-running restore computes its LTX plan once, then streams those files lazily. If snapshot retention runs after a new snapshot appears, the old watermark can advance to the new snapshot TXID and delete lower-level files that the restore plan still references. That produces mid-restore failures such as `reopen ltx file ... file does not exist`.

This implements issue #1307 option 2: retain one extra snapshot generation's lower-level history, so a restore plan computed against the prior snapshot remains usable until another snapshot generation lands.

Fixes #1307

## How Has This Been Tested?
The regression test fails on the unfixed code:

```text
--- FAIL: TestStore_EnforceSnapshotRetention_RetainsInFlightRestorePlanFiles (0.03s)
    db_test.go:856: planned LTX file was deleted by retention: level=1 min=0000000000000006 max=000000000000000a: open ltx file ... no such file or directory
FAIL
FAIL    github.com/benbjohnson/litestream    1.253s
```

After the fix:

```text
go test -run 'Test(Store_EnforceSnapshotRetention_RetainsInFlightRestorePlanFiles|DB_EnforceRetention)$' -count=1 .
go build ./...
go test ./...
pre-commit run --all-files
```

All commands passed.

## Types of changes

## Checklist

### DIFF
--- a/db.go
+++ b/db.go
@@ -2867,9 +2867,11 @@ func (db *DB) EnforceSnapshotRetention(ctx context.Context, timestamp time.Time)
 	defer itr.Close()
 
 	var deleted []*ltx.FileInfo
+	var snapshots []*ltx.FileInfo
 	var lastInfo *ltx.FileInfo
 	for itr.Next() {
 		info := itr.Item()
+		snapshots = append(snapshots, info)
 		lastInfo = info
 
 		// If this snapshot is before the retention timestamp, mark it for deletion.
@@ -2877,17 +2879,23 @@ func (db *DB) EnforceSnapshotRetention(ctx context.Context, timestamp time.Time)
 			deleted = append(deleted, info)
 			continue
 		}
-
-		// Track the lowest snapshot TXID so we can enforce retention in lower levels.
-		// This is only tracked for snapshots not marked for deletion.
-		if minSnapshotTXID == 0 || info.MaxTXID < minSnapshotTXID {
-			minSnapshotTXID = info.MaxTXID
-		}
 	}
 
 	// If this is the snapshot level, we need to ensure that at least one snapshot exists.
 	if len(deleted) > 0 && deleted[len(deleted)-1] == lastInfo {
 		deleted = deleted[:len(deleted)-1]
+	}
+
+	for i, info := range snapshots {
+		if slices.Contains(deleted, info) {
+			continue
+		}
+		if i > 0 {
+			minSnapshotTXID = snapshots[i-1].MaxTXID
+		} else {
+			minSnapshotTXID = info.MaxTXID
+		}
+		break
 	}
 
 	// Remove files marked for deletion from remote storage (unless retention disabled).
--- a/db_test.go
+++ b/db_test.go
@@ -718,7 +718,7 @@ func TestDB_EnforceRetention(t *testing.T) {
 	retentionTime := time.Now().Add(-150 * time.Millisecond)
 	if minSnapshotTXID, err := db.EnforceSnapshotRetention(t.Context(), retentionTime); err != nil {
 		t.Fatal(err)
-	} else if got, want := minSnapshotTXID, ltx.TXID(4); got != want {
+	} else if got, want := minSnapshotTXID, ltx.TXID(3); got != want {
 		t.Fatalf("MinSnapshotTXID=%s, want %s", got, want)
 	}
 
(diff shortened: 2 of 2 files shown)


corylanou:
Merged to main as #1325 via extraction, which includes this fix plus an additional single-snapshot boundary fix (c672695) and expanded tests. Nothing unique remains here.

PR #1315, opened 2026-06-18
feat(config): add meta-dir for directory dbs

### LINKED ISSUE #1193: Feature Request: Support meta-dir for directory-based database configurations
# Feature Request: Support `meta-dir` for directory-based database configurations

When using directory-based database discovery (`dir:` + `pattern:` + `recursive: true`), there is no way to store Litestream's metadata/state files on a separate volume from the application data.

Today, `meta-path` exists for individual `path:`-based database entries, but it doesn't apply to directory configurations. Litestream always creates `-litestream` directories alongside the discovered database files.

## Use case

We run Litestream as a standalone ECS service that monitors SQLite databases on a shared EBS volume. The same volume is used by our application for reads and writes. Litestream's local state files (L0/L1 LTX files, WAL position tracking) are written to the same volume.

We experienced a bug in a prior Litestream version (v0.5.5) where phantom L0 files accumulated at ~209 MB/min, filling the shared disk and taking down both Litestream and our application. While this specific bug is fixed in v0.5.9, the underlying risk remains: any future bug or misconfiguration in Litestream's local state management could exhaust disk space on the application's data volume.

Since we use `dir:` with `recursive: true` to dynamically discover databases, we can't use `meta-path` (which requires knowing each database path ahead of time).

## Proposed solution

Add a `meta-dir` option at the directory-database level that redirects all metadata to a parallel directory tree on a separate path:

```yaml
dbs:
  - dir: /mnt/ebs
    pattern: "*.metadata"
    recursive: true
    meta-dir: /litestream-state
    replica:
      url: s3://my-bucket/backups
```

For a database discovered at `/mnt/ebs/tenant-a/data.metadata`, Litestream would store its metadata at `/litestream-state/tenant-a/data.metadata-litestream/` instead of `/mnt/ebs/tenant-a/data.metadata-litestream/`.

The relative path from `dir` would be preserved under `meta-dir`, maintaining the same parallel tree structure that already works for replica paths.

## Benefits

- Isolates Litestream state from application data, preventing disk-filling bugs from affecting the application
- Allows using ephemeral storage (container filesystem) for state, since it's reconstructable from S3
- No impact on existing behavior — `meta-dir` would be optional, defaulting to current behavior when omitted
- Consistent with how replica paths already mirror the directory structure

## Environment

- Litestream v0.5.9
- ECS on EC2, standalone task with shared EBS volume
- Using `dir:` + `pattern: "*.metadata"` + `recursive: true`

corylanou:
## Description
Add `meta-dir` to directory database configs. Discovered databases now place local Litestream state under the configured root while preserving the path relative to `dir` and appending the standard `-litestream` suffix.

Example behavior: `/mnt/ebs/tenant-a/data.metadata` with `dir: /mnt/ebs` and `meta-dir: /litestream-state` uses `/litestream-state/tenant-a/data.metadata-litestream`.

Scope:
- Add the `meta-dir` config field for directory-based database discovery
- Reject `meta-dir` on single `path` configs
- Reject configs that specify both `meta-path` and `meta-dir`
- Preserve existing `meta-path` behavior

## Motivation and Context
Directory-based discovery can monitor databases without knowing each path up front, but it previously had no dedicated directory-level state root. This allows local LTX state and WAL position metadata to live on a separate volume from application data.

Fixes #1193

## How Has This Been Tested?
- `go build ./...`
- `go test ./cmd/litestream`
- `go test ./...`
- `go test -race ./...`
- `golangci-lint run` fails on existing unrelated `errcheck` findings, including `cmd/litestream-test/load.go:82`, `gs/replica_client.go:163`, and `s3/leaser.go:215`

## Types of changes

## Checklist

### DIFF
--- a/cmd/litestream/main.go
+++ b/cmd/litestream/main.go
@@ -500,6 +500,12 @@ func (c *Config) Validate() error {
 		if db.Watch && db.Dir == "" {
 			return fmt.Errorf("database config #%d: 'watch' can only be enabled with a directory", idx+1)
 		}
+		if db.MetaDir != nil && db.Dir == "" {
+			return fmt.Errorf("database config #%d: 'meta-dir' can only be used with a directory", idx+1)
+		}
+		if db.MetaPath != nil && db.MetaDir != nil {
+			return fmt.Errorf("database config #%d: cannot specify both 'meta-path' and 'meta-dir'", idx+1)
+		}
 
 		// Use path or dir for identifying the config in error messages
 		dbIdentifier := db.Path
@@ -704,6 +710,7 @@ type DBConfig struct {
 	Watch              bool           `yaml:"watch"`     // Enable directory monitoring for changes
 	Snapshot           SnapshotConfig `yaml:"snapshot"`
 	MetaPath           *string        `yaml:"meta-path"`
+	MetaDir            *string        `yaml:"meta-dir"`
 	MonitorInterval    *time.Duration `yaml:"monitor-interval"`
 	CheckpointInterval *time.Duration `yaml:"checkpoint-interval"`
 	BusyTimeout        *time.Duration `yaml:"busy-timeout"`
@@ -718,6 +725,10 @@ type DBConfig struct {
 
 // NewDBFromConfig instantiates a DB based on a configuration.
 func NewDBFromConfig(dbc *DBConfig) (*litestream.DB, error) {
+	if dbc.MetaDir != nil {
+		return nil, fmt.Errorf("'meta-dir' can only be used with a directory")
+	}
+
 	configPath, err := expand(dbc.Path)
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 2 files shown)


cmd/litestream/main.go
```diff
@@ -847,13 +861,23 @@ func newDBFromDirectoryEntry(dbc *DBConfig, dirPath, dbPath string) (*litestream
...
 	// metadata path. Without this, all databases share the same meta-path and
 	// clobber each other's replication state.
-	if dbc.MetaPath != nil {
+	switch {
+	case dbc.MetaDir != nil:
+		baseMetaDir, err := expand(*dbc.MetaDir)
+		if err != nil {
+			return nil, fmt.Errorf("failed to expand meta dir for %s: %w", dbPath, err)
+		}
+		metaPathCopy := deriveMetaPathFromDirectoryRoot(baseMetaDir, relPath)
```
benbjohnson: It doesn't feel like we need a whole new function just for this functionality.

benbjohnson (approved):

cmd/litestream/main.go
corylanou: Fixed in e2b07a0 by inlining the meta-dir path derivation and removing the extra helper.

PR #1316, opened 2026-06-18
feat(cli): read config from stdin

### LINKED ISSUE #1205: Add ability to pass in configuration via STDIN
## Feature Description
First off hello and thanks for this great piece of tech! I brought up this feature idea a little while ago but never got a definitive yes/no on adding an additional flag so that configurations can be passed into Litestream via STDIN (https://github.com/benbjohnson/litestream/issues/356#issuecomment-1292855333).

I have already put together a proof-of-concept and have it working in my fork. All things considered it is not a very intrusive change and doesn't change a lot outside of `main.go`.

## Use Cases
I maintain the https://github.com/akoutmos/litestream library which allows you to easily run Litestream inside of your Elixir application. For most of the Object Storage backup destinations a configuration file is required and having the library create/cleanup temp files for configs feels a little messy. Being able to just pass in a dynamically generated config via STDIN definitely cleans up the integration.

## Attempted Solutions
Workaround for now is creating temp config files and then deleting them as part of app tear down. Not the cleanest experience as cleanup may not happen and then config files with credentials can linger.

## Proposal
The main change here would be to add an additional function in `main.go`:

```go
func ReadConfig(configPath string, fromStdin, expandEnv bool) (Config, error) {
	if fromStdin {
		if configPath != "" {
			return DefaultConfig(), fmt.Errorf("cannot specify both -config and -stdin flags")
		}
		return ParseConfig(os.Stdin, expandEnv)
	}
	if configPath == "" {
		configPath = DefaultConfigPath()
	}
	return ReadConfigFile(configPath, expandEnv)
}
```

This function would delegate to the existing `ReadConfigFile` function so things work as they already do when a config file is available. But then if the `-stdin` flag is passed then `os.Stdin` is passed to `ParseConfig` instead.

Appreciate you taking the time to read through my idea!

corylanou:
## Description
Adds a `-stdin` config source for CLI commands that already read Litestream configuration. The shared config loader now reads from either `-config`, standard input, or the default config path, and rejects `-config` plus `-stdin` together.

Scope:
- In scope: `databases`, `status`, `replicate`, `ltx`, `restore`, and `reset` config loading.
- In scope: rejecting `-stdin` with direct replica URL/positional replica URL flows where `-config` is already invalid.
- Not in scope: changing the YAML config format or adding a persistent config path for stdin-loaded MCP tools.

## Motivation and Context
Issue #1205 requests passing dynamically generated configuration through standard input so integrations do not need to create temporary config files that may contain credentials.

Fixes #1205

## How Has This Been Tested?
- `go test -run 'TestReadConfig|TestDatabasesCommand_Run/StdinJSONOutput' ./cmd/litestream`
- `go test ./cmd/litestream`
- `go build ./...`
- `go test ./...`
- `go test -race ./...`
- `golangci-lint run --new-from-rev=c30c5ffc364d6ee35b0b37d8ce4278371c8107a8`

`golangci-lint run` without a diff base currently reports existing `errcheck` findings outside this change; the diff-scoped run reports `0 issues`.

## Types of changes

## Checklist

### DIFF
--- a/cmd/litestream/databases.go
+++ b/cmd/litestream/databases.go
@@ -15,7 +15,7 @@ type DatabasesCommand struct{}
 // Run executes the command.
 func (c *DatabasesCommand) Run(_ context.Context, args []string) (err error) {
 	fs := flag.NewFlagSet("litestream-databases", flag.ContinueOnError)
-	configPath, noExpandEnv := registerConfigFlag(fs)
+	configPath, stdin, noExpandEnv := registerConfigFlag(fs)
 	jsonOutput := fs.Bool("json", false, "output raw JSON")
 	fs.Usage = c.Usage
 	if err := fs.Parse(args); err != nil {
@@ -25,10 +25,7 @@ func (c *DatabasesCommand) Run(_ context.Context, args []string) (err error) {
 	}
 
 	// Load configuration.
-	if *configPath == "" {
-		*configPath = DefaultConfigPath()
-	}
-	config, err := ReadConfigFile(*configPath, !*noExpandEnv)
+	config, err := ReadConfig(*configPath, *stdin, !*noExpandEnv)
 	if err != nil {
 		return err
 	}
@@ -89,6 +86,9 @@ Arguments:
 	-json
 	    Output raw JSON instead of human-readable text.
 
+	-stdin
+	    Read configuration from standard input.
+
 	-no-expand-env
 	    Disables environment variable expansion in configuration file.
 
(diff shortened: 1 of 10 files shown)


akoutmos:
🤩 Really looking forward to this feature. Thank you so much for addressing #1205!

benbjohnson (changes requested):
Can you change this so that it uses `-config -` to read from STDIN rather than adding a `-stdin` flag? I tried looking around for other tools that read config from STDIN and that seems to be the convention:

- `docker compose -f -`
- `caddy run --config -`
- `kubectl apply -f -`

akoutmos:
Hey there. Anyway I can help get this one over the line? Would really like to get this incorporated into my Elixir library.

Issue #1198, opened 2026-03-15
Checkpoint triggers unwanted full snapshot under high write load


corylanou:
## Summary

Under high write load (500+ writes/sec), checkpoints trigger unwanted full snapshots instead of small incremental L0 syncs. This was detected by the new `TestLTXBehavior` soak test which found **742 snapshot-on-checkpoint violations** in a 30-minute run.

## Evidence

```
[no-snap-on-checkpoint] FAIL: 742 snapshot-on-checkpoint violations detected
[l0-page-count] FAIL: 125/279 (44.8%) L0 files exceed 200 pages. Max seen: 1917.3 MB
```

The "max seen: 1917 MB" for an L0 file is a full database snapshot written at L0 instead of L9.

## Root Cause

The `syncedToWALEnd` flag (added in #927) tracks whether the last sync reached the WAL end, so `verify()` can distinguish expected checkpoint truncation from external WAL modification. However, under high write load:

1. `Sync()` calls `verifyAndSync()` → sync reaches WAL end → `syncedToWALEnd = true`
2. `checkpoint()` calls `verifyAndSync()` BEFORE checkpoint — new writes arrived, sync does NOT reach WAL end → `syncedToWALEnd = false`
3. `execCheckpoint()` truncates WAL
4. `checkpoint()` calls `verifyAndSync()` AFTER checkpoint — `verify()` sees truncated WAL, but `syncedToWALEnd = false` → triggers full snapshot with reason "wal truncated by another process"

The pre-checkpoint sync clears the flag when concurrent writes prevent it from reaching WAL end. The post-checkpoint sync then misidentifies the expected truncation as external.

## Impact

- Full database written as L0 LTX file (up to 1.9GB observed) instead of small incremental
- Downstream compaction failures ("nonsequential page numbers in snapshot") 
- Compaction chain stalls at L2/L3

## Fix

Set `syncedToWALEnd = true` in `checkpoint()` after `execCheckpoint()` succeeds, before the post-checkpoint `verifyAndSync()` call. The checkpoint function knows it just truncated the WAL, so the subsequent truncation detection should always be treated as expected.

Related: #927, #1183

rrva:
**Production occurrence on v0.5.11 (S3 backend) — snapshot-on-checkpoint at *low* write volume, silent until restore**

Confirming this path in production, with one detail that may broaden the known trigger: it fired on a brief **idle→burst transition**, not sustained high load.

[...]

PR #1323, opened 2026-06-28
WAL increment can drop newly-allocated pages → restore-fatal 'nonsequential page numbers' (repro + candidate fix, #1199)

### LINKED ISSUE #1199: Compaction fails with nonsequential page numbers from snapshot-on-checkpoint
## Summary

Under burst load (alternating idle/busy at 1000 writes/sec), compaction at L2 and L3 never completes. Repeated "compaction failed" errors appear with "nonsequential page numbers in snapshot" as the underlying cause.

## Evidence

From `TestLTXBehavior` soak test (burst-volume profile, 30 min):

```
[l0-page-count] FAIL: 541/1101 (49.1%) L0 files exceed 150 pages. Max seen: 21.3 MB
[compaction-timing-L2] Skipped: only 0 compactions (need >=2 for interval check)
[compaction-timing-L3] Skipped: only 0 compactions (need >=2 for interval check)
```

Log errors:
```
write ltx file: copy page 6903 header: nonsequential page numbers in snapshot tr...
compaction failed (repeated)
```

## Root Cause

This is a downstream effect of #1198 (snapshot-on-checkpoint bug). When checkpoints incorrectly trigger full snapshots at L0:

1. L0 files contain full database snapshots with non-sequential page ordering
2. L0→L1 compaction reads these files and the LTX compactor rejects them ("nonsequential page numbers")
3. L1 compaction fails, producing no valid L1 files
4. L2 and L3 have no input files to compact → 0 compactions

## Fix

Fixing #1198 (preventing unwanted snapshots at L0) will resolve this issue, since L0 files will return to small incremental syncs that compact cleanly.

Related: #1198, #1183

rrva:
> **Drive-by, AI-assisted, draft.** I hit this in production and used an AI assistant to investigate and write the fix; I haven't independently audited Litestream's sync internals, so I can't vouch it's the right fix or the right layer. It reproduces the bug, the fix makes it pass, suite + soak stay green — but the design call is yours. Take it, rework it, or just use it as a reproducer.

## The bug

Litestream 0.5.11, S3, `wal_autocheckpoint=0`. A transaction grew the DB 28250 → 28281 pages. Replication logged only routine INFO and the replica looked healthy — but a later cold-start restore failed, total loss:

```
decode database: decode page 28252: copy page 28281 header:
nonsequential page numbers in snapshot transaction: 28251,28281
```

Page 28281 was captured; 28252..28280 were not, yet the increment's `Commit` was 28281.

**Cause:** under a concurrent checkpoint the incremental sync's WAL page map can miss pages the transaction allocated (#1198). `writeLTXFromWAL` writes only the pages in that map but stamps `Commit` = new DB size, and the incremental LTX encoder checks page *ordering* but not *contiguity* — so the gapped increment replicates and compacts with no error. The gap only becomes fatal at restore, where files are re-encoded as a snapshot (`MinTXID==1`) and contiguity *is* enforced. Latent until cold start.

## The fix (containment, not root cause)

In `sync()`, when `commit > prevCommit`, every page in `prevCommit+1..commit` is newly allocated and must be in the page map. If any is missing, fall back to a full snapshot instead of writing the gapped increment (resetting `info.offset` to the WAL header so it's a true full snapshot, not the partial mid-WAL view). `prevCommit` comes from the previous LTX header via `verify()`. Lock page skipped; gated on growth, so the normal path and VACUUM/shrink are untouched.

**This is a write-side guard, not a root-cause fix.** It does not touch the `verify()`/`syncedToWALEnd` checkpoint race that produces the incomplete map (#1198). It catches the **loud** failure (missing *newly-allocated* page → the #1199 restore error); it would **not** catch a dropped update to an *existing* page (≤ prevCommit), which would be silent wrong data — I haven't confirmed whether that variant can occur.

## Reproduction (first principles, on `main`)

#1198 reproduces naturally on current `main`, not just 0.5.11. A stress harness (concurrent writers + **PASSIVE** checkpoints) observing the *real, unmodified* page map saw **6 and 8** natural gaps across ~400–500 growing syncs in two ~90s runs. It's checkpoint-mode-specific: **TRUNCATE-only → 0 gaps** (snapshot-on-truncate covers it), **PASSIVE under write load → gaps** — so `[no-snap-on-checkpoint]` passing doesn't mean it's gone.

That harness isn't committed (concurrent `database/sql` is timing-dependent / lock-order-flaky as a unit test). The committed e2e `TestDB_RestoreSurvivesIncompleteWALView` injects the same gap deterministically via a nil-by-default seam, then does a real `Replica.Restore`: it fails with the production error when the guard is off, and restores with `integrity_check=ok` when on.

## Tests

```bash
go test -run 'TestMissingNewlyAllocatedPage|TestDB_Verify_PrevCommitFromLTXHeader|TestDB_RestoreSurvivesIncompleteWALView' -v .
go test .            # full package: ok
go vet . && gofmt -l db.go db_internal_test.go   # clean
# soak (touches WAL handling): all profiles pass
go build -o bin/litestream ./cmd/litestream && go build -o bin/litestream-test ./cmd/litestream-test
go test -timeout 30m -tags 'integration,soak' -run TestLTXBehavior -short ./tests/integration/
```

Fixes #1199. Related: #1198.

---

🤖 AI-assisted per CONTRIBUTING.md: investigation and code are AI-generated; I reproduced the bug and ran the tests but have not independently audited the sync internals.

### DIFF
--- a/db.go
+++ b/db.go
@@ -176,6 +176,15 @@ type DB struct {
 
 	// Where to send log messages, defaults to global slog with database epath.
 	Logger *slog.Logger
+
+	// Test seam: when non-nil, called with the freshly-built incremental WAL
+	// page map in sync() so tests can simulate an incomplete WAL view (the
+	// #1198 condition). Never set in production.
+	corruptPageMapForTest func(pageMap map[uint32]int64, prevCommit, commit uint32)
+
+	// Test seam: when true, skips the missing-newly-allocated-page guard so a
+	// test can reproduce the pre-fix corruption. Never set in production.
+	disableMissingPageGuardForTest bool
 }
 
 // syncState holds mutable sync-tracking fields extracted from DB.
@@ -1522,6 +1531,7 @@ func (db *DB) verifyWithExecutor(ctx context.Context, exec *syncExecutor) (info 
 	info.offset = dec.Header().WALOffset + dec.Header().WALSize
 	info.salt1 = dec.Header().WALSalt1
 	info.salt2 = dec.Header().WALSalt2
+	info.prevCommit = dec.Header().Commit
 
 	// If LTX WAL offset is larger than real WAL then the WAL has been truncated.
 	if fi, err := os.Stat(db.WALPath()); err != nil {
@@ -1710,6 +1720,7 @@ type syncInfo struct {
 	snapshotting        bool   // if true, a full snapshot is required
 	reason              string // reason for snapshot
 	clearSyncedToWALEnd bool
+	prevCommit          uint32 // commit (page count) of the previous LTX, 0 if unknown
 }
 
 type syncResult struct {
@@ -1802,6 +1813,25 @@ func (exec *syncExecutor) applySyncResult(result syncResult) {
 	exec.synced = exec.synced || result.synced
 }
 
+// missingNewlyAllocatedPage reports whether the page map is missing any page
+// in the range (prevCommit, commit] that the database grew into. Those pages
+// are newly allocated by this transaction and therefore must be present in the
+// WAL view; an absent one means the incremental sync captured an incomplete
+// view of the transaction (see #1198) and a snapshot must be taken instead so
+// the replica does not gain a silently-corrupt increment (see #1199). The lock
+// page never carries data and is skipped.
+func missingNewlyAllocatedPage(pageMap map[uint32]int64, prevCommit, commit, lockPgno uint32) (uint32, bool) {
+	for pgno := prevCommit + 1; pgno <= commit; pgno++ {
+		if pgno == lockPgno {
+			continue
+		}
+		if _, ok := pageMap[pgno]; !ok {
+			return pgno, true
+		}
+	}
+	return 0, false
+}
+
 // sync copies pending bytes from the real WAL to LTX.
 // Returns synced=true if an LTX file was created (i.e., there were new pages to sync).
 func (db *DB) sync(ctx context.Context, checkpointing bool, exec *syncExecutor, info syncInfo) (result syncResult, err error) {
@@ -1891,6 +1921,10 @@ func (db *DB) sync(ctx context.Context, checkpointing bool, exec *syncExecutor, 
 		commit = walCommit
 	}
 
+	if db.corruptPageMapForTest != nil && !info.snapshotting {
+		db.corruptPageMapForTest(pageMap, info.prevCommit, commit)
+	}
+
 	var sz int64
 	if maxOffset > 0 {
 		sz = maxOffset - info.offset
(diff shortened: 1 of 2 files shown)


sroze:
We independently hit this exact defect in production-like use with Litestream
v0.5.13 (`superfly/ltx` v0.5.1), and decoded the preserved LTX chain.

SQLite page size: 4096 bytes.

The relevant L0 sequence was:

- TXID 1991: `Commit=41531`; includes pages 41530 and 41531.
- TXID 1992: `Commit=41536`, but its page index contains only page 62.
  Pages 41532–41536 are entirely absent.
- TXID 1993: `Commit=41561`; its newly allocated range starts at page
  41537 and continues through 41561.

Observed restore behavior:

- Restore through TXID 1991 passes Litestream's full integrity check.
- Restore through TXID 1992 fails:
  `decode database: decode page 41532: EOF`
- Including TXID 1993 fails:
  `copy page 41537 header: nonsequential page numbers in snapshot transaction: 41531,41537`

The checkpoint correlation is unusually strong:

- A standalon process opened the same busy SQLite DB and
  unconditionally executed `PRAGMA wal_checkpoint(PASSIVE)`.
- Process start: `10:44:34.017Z`.
- LTX timestamps for 1990/1991/1992:
  `10:44:34.027Z`, `.036Z`, and `.054Z`.
- Between 1991 and 1992, WAL salt-1 incremented, salt-2 changed, and
  WAL offset reset to 32, confirming a checkpoint/WAL reset.

The DB was also receiving concurrent writes from the long-lived server.

This appears to confirm the condition guarded by this PR exactly:
`Commit` grew from 41531 to 41536 while none of the newly allocated pages
41532–41536 existed in the incremental page map. The incomplete increment
was accepted, uploaded, and compacted without warning; corruption became
visible only during cold restore.

S3 checksums passed and the replica objects were preserved, so this was not
an interrupted upload or object-store loss.

petergaultney:
🍋:

Another production hit on v0.5.11 (Azure Blob replica, `wal_autocheckpoint=0`), plus a reproduction and a detector, in case they help. Our restore error:

```
decode database: decode page 4329636: copy page 4329925 header:
nonsequential page numbers in snapshot transaction: 4329635,4329925
```

## Reproduction

v0.5.11, one WAL-mode db per arm, with concurrent writers on separate connections, all `busy_timeout=30000` and `wal_autocheckpoint=0`:

- 3.5 MB row insert every 2s
- 256 KB row insert every 0.5s
- 1 KB row insert every 0.1s

Every arm got the same seeded workload for 30 minutes. A small detector built on `superfly/ltx` v0.5.1 read each local L0 file as it was written and flagged any whose `Commit` exceeded the previous file's `Commit` without containing every page in between (lock page excluded).

| arm | config | gapped L0 files | pages missing | `restore` of latest |
|---|---|---|---|---|
| A | defaults | 2 | 63, 63 | fails |
| B | `levels: [5s, 5m, 1h]`, `l0-retention: 1h` | 2 | 63, 63 | fails |
| C | as B, plus `checkpoint-interval: 0s`, `min-checkpoint-page-count: 100000000`, `truncate-page-n: 5000` (TRUNCATE only) | 1 | 855 | fails |

Those arms replicated to Azure Blob. A rerun against a local `file://` replica with the release v0.5.11 binary produced the same kind of file within 11 minutes (1 page missing, from a 1 KB insert that grew the db), so it isn't backend-specific.

- **Compaction and retention settings make no difference** (A vs B).
- **TRUNCATE-only checkpointing does not avoid it** (C), unlike the harness result in the description above.
- **Every gapped file is the ~208-byte file written right after litestream's own checkpoint** (the `_litestream_seq` bump). Its `Commit` includes the growth from one concurrent application write, 63 pages for a 256 KB row or 855 for a 3.5 MB row, and it contains none of that write's pages. The one in our production chain was 289 pages, which fits one ~1.2 MB write.

## What we think the window is

In `DB.checkpoint()`, `verifyAndSync()` runs, then `execCheckpoint()` releases the read lock and checkpoints. An application commit that lands between the two is backfilled into the db file by the checkpoint. The next write, usually the `_litestream_seq` insert, then restarts the WAL over it. Afterwards `verify()` takes the "wal restarted" path. `lastPageMatch()` only confirms that the last frame litestream synced is still intact, not that litestream had synced to the end of the old WAL, so the committed frames are skipped. When that write grew the db, the next LTX file carries the larger `Commit` without the pages, which is this issue. When it didn't grow the db, we'd expect the write to be dropped from the replica silently. We haven't demonstrated that case.

`TestDB_CheckpointPageGapWithConcurrentWrites` failed when we ran it on the v0.5.11 tag (`go test -run 'Checkpoint|Sync|Restore' .`), which looks like the same window.

We're testing a candidate fix for that window against this reproduction and will follow up with results.

PR #1324, opened 2026-06-29
[DO NOT MERGE] soak integration: Stack A (corruption) + Stack B (S3 resilience)

corylanou:
**Soak-validation branch only — do NOT merge.**

This combines the two independent in-flight fix stacks so the litestream-soak system can validate them together (no single existing PR head contains both, so every individual PR soaks red on the half it's missing):

- **Stack A — corruption / checkpoint hardening:** #1290 → #1291 → #1292 → #1166 → #1302, plus both sibling leaves #1312 (retention: retain prior snapshot txid, cherry-picked) and #1322 (disk-full staging).
- **Stack B — S3 fault resilience:** #1305 (retry through transport flaps) → #1306 (retry through provider throttle) → #1311 (resume remote compaction reads).

The two stacks touch disjoint files and combine cleanly. Local `go build`, `go vet`, and `go test . ./s3 ./internal` all pass.

Purpose: produce a single branch that should pass the full soak fleet (general + fault scenarios), as a green reference for the stacked-PR release. Intended to be soaked, reviewed, and then have its constituent stacks merged piecemeal — this integration branch itself is not a merge target.

PR #1325, opened 2026-06-30
fix(retention): retain prior snapshot txid

corylanou:
Standalone extraction of the retention fix (originally commit `06052c9`, also in stacked PR #1312) so it can be reviewed and merged independently of the larger corruption stack.

**What:** retention now preserves lower-level files back to the prior retained snapshot TXID, so an in-flight restore plan can't lose files it still needs.

**Independence:** clean cherry-pick onto current `main`; touches only `db.go` (+ `db_test.go`). `go build ./...` and `TestStore_EnforceSnapshotRetention*` pass.

**Validation:** this fix is part of the soak-validated integration branch (#1324) that passes the full general fleet + all S3 fault scenarios.

### DIFF
--- a/db.go
+++ b/db.go
@@ -2491,9 +2491,11 @@ func (db *DB) EnforceSnapshotRetention(ctx context.Context, timestamp time.Time)
 	defer itr.Close()
 
 	var deleted []*ltx.FileInfo
+	var snapshots []*ltx.FileInfo
 	var lastInfo *ltx.FileInfo
 	for itr.Next() {
 		info := itr.Item()
+		snapshots = append(snapshots, info)
 		lastInfo = info
 
 		// If this snapshot is before the retention timestamp, mark it for deletion.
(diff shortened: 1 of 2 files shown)


db.go
```diff
@@ -2491,29 +2491,37 @@ func (db *DB) EnforceSnapshotRetention(ctx context.Context, timestamp time.Time)
...
 	}
 
+	for i, info := range snapshots {
+		if slices.Contains(deleted, info) {
+			continue
+		}
+		if i > 0 {
+			minSnapshotTXID = snapshots[i-1].MaxTXID
+		} else {
+			minSnapshotTXID = info.MaxTXID
```
benbjohnson: If there's only one item in `snapshots` then you don't get the prior snapshotTXID or zero but rather it sets it to the current snapshot TXID. Should the `else` block be eliminated since there's no previous snapshot TXID to set it to?

corylanou: Fixed in the latest commit: snapshot retention now leaves the lower-level retention boundary at zero when there is no prior retained snapshot, so lower-level files are not pruned based on the only snapshot. I also added a regression test for the single-snapshot case.

benbjohnson (approved):

PR #1328, opened 2026-07-02
perf(s3): single PutObject for small ltx objects

### LINKED ISSUE #1327: perf(s3): single PutObject for small LTX objects (skip multipart part-buffers)
## Background

Profiling the many-database tier (100 mostly-idle ~5 MB SQLite DBs, one Litestream process) showed the AWS S3 upload manager's 5 MiB part buffers are the dominant allocation traffic and the `memclr`/zeroing CPU cost, and the multipart path hard-floors at 5 MiB so `part-size` config can't shrink it for tiny objects. Approved direction: for small objects, do a single `PutObject` instead of always streaming through the multipart manager ("we stream the data in but should be able to do a rough estimate of file size").

## Approach — size-aware first, buffer only as fallback

In `s3/replica_client.go` `WriteLTXFile` (currently always `c.uploader.Upload`), add a small-object fast path that avoids the multipart part-pool:

- **Known size** — the normal L0 replication path passes the local LTX file as an `*os.File` into `WriteLTXFile` (`replica.go:184`), so the size is available (stat / known length):
  - size `<` effective part size → direct `c.s3.PutObject` with an exact `ContentLength` and a seekable body.
  - size `>=` effective part size → existing `c.uploader.Upload` with a seekable body.
- **Unknown size** → read into a grow-on-demand buffer up to the threshold; if EOF before threshold, `PutObject` from `bytes.NewReader(buf)`; otherwise fall back to `uploader.Upload` with `io.MultiReader(bytes.NewReader(prefix), r)` so nothing is re-read.

Threshold = effective configured `PartSize` (default `manager.DefaultUploadPartSize`, honor `c.PartSize`), and preserve the SDK's minimum-part-size validation — don't let a too-small configured part size silently succeed.

## Correctness traps (test these)

- `ContentLength` must be exact on the direct `PutObject`.
- Retryability only holds if the single-PUT body is **seekable** (`bytes.NewReader` or `*os.File`). Do **not** wrap it in `internal.ReadCounter` — that strips seekability and is the current self-inflicted problem (`s3/replica_client.go:688`, `internal/internal.go:59`).
- **Parameter parity:** factor a shared input-building helper so both paths set the same metadata timestamp, `StorageClass` (#1321), SSE-C / SSE-KMS, bucket/key. Don't invent a `ContentType` if `main` doesn't set one.
- The direct `PutObject` must use the already-initialized `c.s3` client so checksum/signing/custom-endpoint middleware still applies.
- Keep the existing ETag success check + metrics accounting.

## Base branch

Base on current `origin/main` — the S3 retryer (#1305/#1306) and storage-class (#1321) work is already merged there. Do **not** stack on the old `fix-s3-retry-quota-exhaustion` / #1311 branches; they lack the storage-class `WriteLTXFile` changes.

## Verify

Sub-threshold object goes via `PutObject` (assert no multipart); over-threshold still uses multipart; byte-for-byte round-trip both ways; StorageClass/SSE parity on both paths. Test against MinIO + the S3 mock. (No manifest/caching — declined separately.)

corylanou:
## Description

`WriteLTXFile` wrapped every upload body in a non-seekable `ReadCounter(MultiReader(...))`, forcing the AWS upload manager down its part-buffer path — which allocates a fresh 5 MiB part buffer per call (the manager's pool returns its capacity when each `Upload` completes, so buffers never survive between calls), even for a 200-byte LTX file.

This PR routes uploads by size:

- **Known size** (seekable body — the L0 replication path passes the local `*os.File`, `replica.go:186-192`): size measured via `Seek`; below the effective part size → direct `c.s3.PutObject` with exact `ContentLength` and a seekable (retryable) body; at/above it → existing `c.uploader.Upload` with the raw seekable body, which the SDK slices into zero-copy `SectionReader`s instead of part buffers.
- **Unknown size** (compaction/snapshot pipes): grow-on-demand buffer up to the part size; EOF before that → `PutObject` from `bytes.NewReader(buf)`; otherwise fall back to `uploader.Upload` with `io.MultiReader(prefix, rest)` so nothing is re-read.
- **Parameter parity**: a shared `putObjectInput` helper builds the one `*s3.PutObjectInput` both paths use (timestamp metadata, `StorageClass` from #1321, SSE-C/SSE-KMS). No `ContentType` invented. The direct `PutObject` goes through the already-initialized `c.s3`, so the transport retryer (#1305), signing, checksum, and endpoint middleware all still apply.
- **Misconfiguration stays loud**: a configured `part-size` below the SDK's 5 MiB floor fails every upload with the SDK's own message (`part size must be at least 5242880 bytes`) instead of silently succeeding for small objects.

**Scope**: only `s3/replica_client.go` upload routing. No manifest/caching work (declined separately in the issue), no changes to other backends, no config surface changes.

**Deliberate wire change**: an object of *exactly* the part size is now one `PutObject` instead of a one-part multipart upload (the uploader sees the total size once the body is seekable) — 1 request instead of 3. `TestReplicaClient_MultipartUploadThreshold/AtThreshold_5MB` updated with a comment.

## Motivation and Context

Profiling the many-database tier (100 mostly-idle ~5 MB SQLite DBs in one process) showed the upload manager's 5 MiB part buffers as the dominant allocation traffic and `memclr` CPU cost, and the multipart floor means `part-size` config can't shrink them for tiny objects.

Measured by the new allocation regression test (TotalAlloc across 16 uploads of an 8 KiB LTX, mock S3 endpoint):

| Path | before | after |
|---|---|---|
| known-size `*os.File` | 5.08 MiB/upload | 0.09 MiB/upload |
| unknown-size pipe | 5.08 MiB/upload | 0.07 MiB/upload |

Fixes #1327

## How Has This Been Tested?

TDD: the allocation test was written first and failed at 5.08 MiB/upload on both paths before the change.

```bash
go test -race ./s3/                                  # unit + regression tests, all green
go test -race ./...                                  # full suite green
go test -race . -v -run TestReplicaClient -integration -replica-clients s3   # against MinIO
pre-commit run --all-files                           # gofmt/vet/staticcheck green
```

New tests: `TestReplicaClient_WriteLTXFile_SmallUploadAllocations` (the perf guard, logs measured MiB/upload), `_SinglePutKnownSize` (exact `Content-Length`, byte-for-byte body, no multipart), `_SinglePutRetries` (first PUT 500 → SDK replays the seekable body intact — guards the "don't wrap in ReadCounter" trap), `_StreamSmallSinglePut`, `_StreamLargeMultipartRoundTrip` (parts reassemble byte-for-byte across the buffered-prefix handoff), `_PartSizeBelowMinimum` (fails with zero requests on the wire), `_MultipartParamParity` (StorageClass/SSE-C/timestamp metadata on the multipart initiate).

End-to-end against real MinIO with server-side API tracing (`mc admin trace`):

- `litestream replicate` of a WAL-mode DB: small L0 files and small streamed L1/L9 compactions arrive as single `s3.PutObject`; a 12.1 MiB L0 file and a large streamed L1 compaction arrive as `NewMultipartUpload` + 3 `PutObjectPart` + `Complete` — all four routing paths observed on the wire.
- `litestream restore` round-trip: `PRAGMA integrity_check` ok, row content and 12,582,912 blob bytes identical to origin.
- `part-size: 1MB` via config: 46 loud errors across snapshot/compaction/monitor, zero write APIs reached the server.

## Types of changes

## Checklist

### DIFF
--- a/s3/replica_client.go
+++ b/s3/replica_client.go
@@ -680,38 +680,156 @@ func (c *ReplicaClient) OpenLTXFile(ctx context.Context, level int, minTXID, max
 
 // WriteLTXFile writes an LTX file to the replica.
 // Extracts timestamp from LTX header and stores it in S3 metadata to preserve original creation time.
+// Objects smaller than the part size are written with a single PutObject call
+// to avoid the multipart upload manager's 5 MiB part buffers (issue #1327).
 func (c *ReplicaClient) WriteLTXFile(ctx context.Context, level int, minTXID, maxTXID ltx.TXID, r io.Reader) (*ltx.FileInfo, error) {
 	if err := c.Init(ctx); err != nil {
 		return nil, err
 	}
 
-	// Use TeeReader to peek at LTX header while preserving data for upload
-	var buf bytes.Buffer
-	teeReader := io.TeeReader(r, &buf)
-
-	// Extract timestamp from LTX header
-	hdr, _, err := ltx.PeekHeader(teeReader)
-	if err != nil {
-		return nil, fmt.Errorf("extract timestamp from LTX header: %w", err)
-	}
-	timestamp := time.UnixMilli(hdr.Timestamp).UTC()
-
-	// Combine buffered data with rest of reader
-	rc := internal.NewReadCounter(io.MultiReader(&buf, r))
-
 	filename := ltx.FormatFilename(minTXID, maxTXID)
 	key := c.Path + "/" + fmt.Sprintf("%04x/%s", level, filename)
 
-	// Store timestamp in S3 metadata for accurate timestamp retrieval
-	metadata := map[string]string{
-		MetadataKeyTimestamp: timestamp.Format(time.RFC3339Nano),
+	partSize := int64(manager.DefaultUploadPartSize)
+	if c.PartSize > 0 {
+		partSize = c.PartSize
 	}
 
+	// The uploader rejects part sizes below the SDK minimum on every upload.
+	// Fail identically here so the single-put path cannot mask the
+	// misconfiguration for small objects.
+	if partSize < manager.MinUploadPartSize {
+		return nil, fmt.Errorf("s3: upload to %s: part size must be at least %d bytes", key, manager.MinUploadPartSize)
+	}
+
+	size, timestamp, etag, err := c.uploadLTX(ctx, key, r, partSize)
+	if err != nil {
+		return nil, err
+	}
+
+	internal.OperationTotalCounterVec.WithLabelValues(ReplicaClientType, "PUT").Inc()
+	internal.OperationBytesCounterVec.WithLabelValues(ReplicaClientType, "PUT").Add(float64(size))
+
+	// ETag indicates successful upload
+	if etag == nil {
+		return nil, fmt.Errorf("s3: upload failed: no ETag returned")
+	}
+
+	return &ltx.FileInfo{
+		Level:     level,
+		MinTXID:   minTXID,
+		MaxTXID:   maxTXID,
+		Size:      size,
+		CreatedAt: timestamp,
+	}, nil
+}
+
+// uploadLTX uploads LTX data from r to key, choosing between a single
+// PutObject and the multipart uploader based on the object size.
+func (c *ReplicaClient) uploadLTX(ctx context.Context, key string, r io.Reader, partSize int64) (int64, time.Time, *string, error) {
+	// The L0 replication path passes the local LTX file, so the size is
+	// known up front and the body stays seekable for SDK retries.
+	if rs, ok := r.(io.ReadSeeker); ok {
+		if start, err := rs.Seek(0, io.SeekCurrent); err == nil {
+			return c.uploadSizedLTX(ctx, key, rs, start, partSize)
+		}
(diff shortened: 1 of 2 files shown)


ry:
We hit exactly this failure mode in a multi-machine workload with 1000 independently replicated SQLite databases across five 2 GiB VMs, roughly 200 databases per VM. The workload continuously wrote to the databases while exercising concurrent activation and replication.

With Litestream v0.5.15, synchronized small-LTX uploads caused Litestream to be OOM-killed on every node. Lowering `GOMEMLIMIT` from 512 MiB to 256 MiB did not resolve it.

We pinned this PR’s commit (5a26244) and repeated the same workload with the normal 512 MiB limit. Litestream stayed around 118–143 MiB RSS during load, versus 456–1,005 MiB before the fix. All 1k databases passed application-level sequence and checksum verification, with zero OOMs, process restarts, or workload errors.

We are currently pinning this commit in our application. I hope it gets landed!

benbjohnson (approved):

benbjohnson:
@ry Thanks for the feedback on your testing. That really helps a lot. We just cut a new release with this fix: https://github.com/benbjohnson/litestream/releases/tag/v0.5.16

PR #1337, opened 2026-07-07
fix(db): hold write lock across checkpoint to prevent replica corruptio

oliverkofoed:
## Summary
In a system I operate, sometimes "litestream restore" will create an invalid database file. 

I've got a test case (tests/integration/concurrent_replication_corruption_test.go) in
this PR that reliably reproduces the error: one litestream process replicates a database
file to a file target while a concurrent writer inserts rows into that same database. The
test then restores from the target in a loop and checks integrity — and within a few
seconds hits a restore that produces an invalid database file (fails PRAGMA
integrity_check, or fails to decode).

The restore loop is only how the test *detects* the problem — the corruption is baked
into the replica during ordinary replication, before any restore. So this is a bug in the
capture/checkpoint path, not in restore.

This PR also includes a potential fix, which can be disregarded in favor of a more
appropriate one. AI was used to help reduce the bug to a reproducible test case and 
to write the fix.

## Problem
During continuous replication with a concurrent writer, replicas would
intermittently fail `PRAGMA integrity_check` after restore, with signatures like:

- `wrong # of entries in index ...`
- `Page N: never used`

`checkpointWithExecutor` did not hold the database write lock between (a) copying
the WAL tail into an LTX file and (b) running `PRAGMA wal_checkpoint`. A concurrent
writer could commit frames in that window; the checkpoint backfilled them into the
database file, and the subsequent WAL restart (triggered by litestream's own
`_litestream_seq` write) discarded them from the WAL. `verify()` then saw a
benign-looking single-generation salt change and continued incrementally on the new
WAL, so those pages never reached any LTX file.

Instrumentation confirmed the race: 32 events in a single 5s run where the
checkpoint backfilled more frames than litestream had captured (e.g. `nCkpt=17` vs
`captured=15`).

## Solution
Two complementary mechanisms in `db.go`:

1. **Close the race:** acquire the write lock (`BEGIN` + `_litestream_lock` insert,
   always rolled back — the same trick the post-checkpoint copy already used) before
   the pre-checkpoint WAL copy, and hold it through PASSIVE checkpoints (which never
   need SQLite's writer lock). No frame can be committed between capture and
   checkpoint.
2. **Safety net:** `execCheckpoint` now returns `(frames-in-WAL, frames-backfilled)`.
   If a checkpoint may have backfilled uncaptured frames (provable overrun for
   PASSIVE, always assumed for FULL/RESTART/TRUNCATE), a persistent `snapshotPending`
   flag forces the next sync to write a full snapshot from the database file,
   repairing the LTX chain. The flag survives transient sync failures and clears only
   after a snapshot is written.

## Scope
**In scope:**
- Checkpoint locking + safety-net snapshot in `db.go`
- Regression test in `tests/integration/`

**Not in scope:**
- No changes to compaction, restore, or storage backends

## Test Plan
```bash
# Regression test (fails pre-fix in ~5s, passes post-fix)
go test -tags integration -race -run TestConcurrentWriteReplicationCorruption ./tests/integration/

# Full race suite
go test -race ./...

Post-fix: the regression test passes the full window with zero safety-net triggers;
go test -race ./... passes.

### DIFF
--- a/db.go
+++ b/db.go
@@ -202,6 +202,15 @@ type syncState struct {
 	// a checkpoint. This prevents issue #997 where PASSIVE checkpoints
 	// trigger a feedback loop because stale file size exceeds threshold.
 	lastSyncedWALOffset int64
+
+	// snapshotPending forces the next sync to write a full snapshot from
+	// the database file instead of copying incrementally from the WAL. Set
+	// after a checkpoint that may have backfilled WAL frames into the
+	// database file before they were captured to an LTX file; the WAL alone
+	// can no longer reconstruct the database state. Cleared only once a
+	// snapshot has been written successfully so the repair survives
+	// transient sync failures.
+	snapshotPending bool
 }
 
 type syncExecutor struct {
@@ -239,6 +248,7 @@ const (
 	diagPhaseSyncComplete                   diagPhase = "sync_complete"
 	diagPhaseCheckpointLock                 diagPhase = "checkpoint_lock"
 	diagPhaseCheckpointReadWALHeader        diagPhase = "checkpoint_read_wal_header"
+	diagPhaseCheckpointCopyBeforeLock       diagPhase = "checkpoint_copy_before_lock"
 	diagPhaseCheckpointCopyBefore           diagPhase = "checkpoint_copy_before"
 	diagPhaseCheckpointExec                 diagPhase = "checkpoint_exec"
 	diagPhaseCheckpointVerifyRestart        diagPhase = "checkpoint_verify_restart"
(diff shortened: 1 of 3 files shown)


corylanou:
Thanks for this, @oliverkofoed — you identified a real window, and the diagnosis was correct. It has since been fixed on `main` by a different route, so I'm closing this as superseded rather than leaving it to rot.

This PR is based on `2b34013`, before the stability chain landed. Comparing it against current `main` (`4e3f0c0`), the two things it does are now covered:

- **Holding the write lock across a PASSIVE checkpoint** — `main` now blocks writers with a writer barrier and performs a second sealed sync to the WAL boundary before proceeding (`db.go:2453`), then handles the WAL restart with a post-checkpoint sync.
- **Fallback snapshot when the checkpoint may have hidden commits** — `main` now takes the boundary snapshot unconditionally in TRUNCATE mode, because `wal_checkpoint(TRUNCATE)` reports zero frames and therefore can never prove that no commits landed between the sealed sync and the checkpoint (`db.go:2529`).

`main` additionally backfills WAL growth pages from the database file (`db.go:2299`) and makes snapshot capture mutually exclusive with checkpoints by taking `chkMu.RLock` while still holding the executor semaphore (`db.go:2675`).

Landed in #1292 (`d26cb54`), with #1166 (`9ca1201`) covering fsync ordering on the restore/follow apply path. I re-checked your diff against current `main` in refute mode specifically looking for a window it still closes, and did not find one.

If you see a case that's still open on current `main`, please reopen or file an issue with the repro — happy to be wrong about this.

PR #1358, opened 2026-07-10
fix(mcp): report running daemon version

### LINKED ISSUE #1353: MCP litestream_version reports PATH binary version, not the running daemon
## Summary

The MCP `litestream_version` tool shells out to whatever `litestream` binary is on `$PATH` (`cmd/litestream/mcp.go:250`) rather than reporting the running daemon's own version.

## Impact

If the daemon was started from a different binary than the one on `$PATH` (source build, container, versioned install path), the MCP tool reports the wrong version — precisely the situation where an AI assistant querying the daemon needs the truth.

## Suggested fix

Return the in-process version string (same value `litestream version` prints for the running build) instead of exec'ing the PATH binary.

corylanou:
## Description
The MCP `litestream_version` tool now returns the daemon's in-process `Version` value instead of launching whichever `litestream` executable appears first on `$PATH`. The response preserves the existing newline formatting, and the tool description now identifies the running daemon as the version source.

A regression test removes `litestream` from `$PATH`, resolves a distinct in-process version through the machinery added by #1338, and verifies that the MCP response still reports it.

**Scope:** This changes only `litestream_version`; other MCP tools that intentionally invoke CLI operations remain unchanged.

This PR is stacked on #1338 and uses its version-resolution machinery.

## Motivation and Context
The previous handler called `exec.CommandContext(ctx, "litestream", "version")`, so its answer described the PATH binary rather than the daemon serving the MCP request. If no PATH binary existed, the regression test failed with:

```text
exec: "litestream": executable file not found in $PATH
```

Returning the already-resolved process value makes the MCP result match the running daemon regardless of how it was launched or what is installed on `$PATH`.

Fixes #1353

## How Has This Been Tested?
- Verified `go test ./cmd/litestream -run '^TestVersionTool$' -count=1 -v` failed before the production change and passed afterward.
- `go test ./... -count=1`
- `go test -race ./... -count=1`
- `go vet ./...`
- `go build -o bin/litestream ./cmd/litestream`

## Types of changes
- [ ] Breaking change (would cause existing functionality not to work as expected)

## Checklist

### DIFF
--- a/cmd/litestream/mcp.go
+++ b/cmd/litestream/mcp.go
@@ -243,16 +243,11 @@ func RestoreTool(configPath string) (mcp.Tool, server.ToolHandlerFunc) {
 
 func VersionTool() (mcp.Tool, server.ToolHandlerFunc) {
 	tool := mcp.NewTool("litestream_version",
-		mcp.WithDescription("Print the Litestream binary version."),
+		mcp.WithDescription("Print the running Litestream daemon version."),
 	)
 
 	return tool, func(ctx context.Context, req mcp.CallToolRequest) (*mcp.CallToolResult, error) {
-		cmd := exec.CommandContext(ctx, "litestream", "version")
-		output, err := cmd.CombinedOutput()
-		if err != nil {
-			return mcp.NewToolResultError(strings.TrimSpace(string(output)) + ": " + err.Error()), nil
-		}
-		return mcp.NewToolResultText(string(output)), nil
+		return mcp.NewToolResultText(Version + "\n"), nil
 	}
 }
 
--- a/cmd/litestream/mcp_test.go
+++ b/cmd/litestream/mcp_test.go
@@ -1,0 +1,33 @@
+package main
+
+import (
+	"testing"
+
+	"github.com/mark3labs/mcp-go/mcp"
+)
+
+func TestVersionTool(t *testing.T) {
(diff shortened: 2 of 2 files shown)


corylanou:
Closing in favor of #1375, which supersedes this.

Both PRs replace the `$PATH` binary exec in `litestream_version` with the in-process `Version`. #1375 came out of the MCP audit (issue #1365, which explicitly took over from #1353) and is a superset of this change:

- It also fixes `NewMCP` advertising a hardcoded `"1.0.0"` as the server identity, which this PR does not touch.
- It adds `TestMCPServerInitializeVersion`, a wire-level test that stands up an `httptest` server, performs a real JSON-RPC `initialize` handshake, and asserts `ServerInfo.Version` in the response — not just the tool output.
- Its tool description says "the running Litestream instance's version", matching the two-versions distinction from #1365: the tool reports the Litestream **instance** the MCP represents, while the handshake carries the MCP server's own build.

There is also a structural reason to prefer #1375: this PR is stacked on #1338 (the CLI dev-build version work), which unnecessarily couples the MCP stack to an unrelated CLI change.

Nothing here is lost. The one small difference is that this PR's test drives `resolveVersion(...)` rather than setting a literal version string; that is a test-only nuance and not worth carrying the duplicate.

Tracked in #1365 / #1375.

Issue #1367, opened 2026-07-16
refactor(mcp): call Litestream in-process instead of exec-ing a PATH binary


corylanou:
## Problem

Most MCP tools run `exec.CommandContext(ctx, "litestream", …)` (`cmd/litestream/mcp.go:96,116,167,235,287,313,339`). Consequences: they depend on a `litestream` on `$PATH` that may differ from the running build, pay process-spawn cost per call, lose typed errors, and couple correctness to CLI text output. The server may target **local paths or remote replica URLs** (`s3://`, `gs://`) or a remote Litestream instance (e.g. Fly.io) — the implementation must keep supporting all of these.

> Note: `litestream_version` is **out of scope** here — its exec removal and version semantics are owned by #1365. This issue converts the remaining tools and must not reintroduce any PATH-binary behavior.

## Fix

Replace the shell-outs with in-process calls to Litestream's library / daemon-IPC API, reusing existing surface:

- restore → `litestream.NewReplicaFromConfig` + `Replica.Restore` / `Replica.CalcRestoreTarget` (`replica.go`)
- ltx → `ReplicaClient.LTXFiles` (`replica_client.go`)
- status → `NewDBFromConfig` + `DB.SyncStatus` / `DB.MaxLTX` (`db.go`)
- databases → `ReadConfigFile` + `NewDBFromConfig`
- reset → `DB.ResetLocalState`
- daemon-backed tools (see #1372) → the existing Unix-socket IPC client

Explicitly handle both local paths and remote replica URLs. This same in-process/IPC path is what later lets tools talk to a **connected** (possibly remote) daemon instead of exec — including the remote-version behavior noted in #1365. Can be done tool-by-tool.

## Acceptance

- No MCP tool spawns a `litestream` subprocess (the `version` tool is already converted by #1365).
- Tools work against local databases and remote replica URLs.
- No tool reintroduces PATH-binary drift.

## Cross-refs

Complements #1365 (which owns the `litestream_version` / #1353 fix and the embedded-vs-remote version semantics). Touches every remaining handler, same code as #1368 — recommend landing **before** #1368, or combining in one PR at the maintainer's discretion. #1367 + #1370 enable the remote "connected-daemon version" behavior described in #1365.

PR #1375, opened 2026-07-16
fix(mcp): report the running instance version, not a PATH binary

### LINKED ISSUE #1365: fix(mcp): report the running instance's version, not a PATH binary (closes #1353)
## Problem

Two distinct "versions" are wrong/conflated in the MCP server today, and #1353 is about the first:

1. **The `litestream_version` tool** execs whatever `litestream` is on `$PATH` (`cmd/litestream/mcp.go:250`). It should report the version of the **Litestream instance the MCP server represents / is connected to** — i.e. the running daemon — not an unrelated binary that happens to be first on `$PATH`. This is #1353.
2. **The MCP server's own handshake identity** is hardcoded `"1.0.0"` (`cmd/litestream/mcp.go:33`), unrelated to the real build.

These must not be conflated:
- *"What Litestream instance am I talking to?"* (version of the daemon being operated on) → **tool result**.
- *"What build is running the MCP server component?"* → **MCP handshake identity**.

They happen to coincide **today** only because the MCP server runs **embedded inside the `replicate` daemon process** (`cmd/litestream/replicate.go:184-189`): the in-process build `Version` (`cmd/litestream/main.go:40`) equals the daemon's version, which the daemon already reports over IPC (`server.go:511` sets `InfoResponse.Version = s.Version`, wired at `replicate.go:306`).

## Fix (current embedded architecture)

- `litestream_version` returns the running instance's version from the same source the daemon reports it — in-process `Version` (== `s.Server.Version` == `/info`'s `InfoResponse.Version`) — with **no exec**.
- Set the MCP server handshake version in `server.NewMCPServer` to the real build `Version` instead of `"1.0.0"`.

## Future (standalone / remote) — cross-ref only, out of scope here

Once the MCP server can run **standalone** (#1370) or against a **remote daemon** (#1367), the MCP process is no longer the daemon and the two versions diverge. In that mode the tool must source the instance version from the **connected daemon over IPC** (`/info` → `InfoResponse.Version`, `server.go:509,538-544`), not from the MCP process's own binary (it may additionally surface the MCP server's own build for diagnostics). Those issues own that mechanism; this issue fixes the embedded case and must not bake in an assumption that blocks it.

## Acceptance

- `litestream_version` reports the running instance's version with no subprocess, matching `litestream version` of the daemon build.
- The MCP handshake advertises the real build version, not `"1.0.0"`.
- Closes #1353.

## Cross-refs

Sole owner of the #1353 version bug. #1367 (in-process refactor) converts the **other** tools and must not touch/reintroduce version reporting; #1367 and #1370 own the remote/standalone "connected-daemon version" behavior described above.

corylanou:
## Description
The MCP version tool now reports the in-process build version for the running embedded daemon instead of launching whichever `litestream` binary `$PATH` resolves. The MCP initialize response also advertises the real build version rather than the static `1.0.0` identity.

This change is limited to embedded MCP version reporting and handshake identity. Standalone or remote connected-daemon version sourcing remains out of scope.

## Motivation and Context
A daemon started from a source build, container, or versioned install path could report the version of an unrelated `litestream` executable. The regression reproduced this by removing `litestream` from `$PATH`, which returned an executable-not-found tool error, while initialization independently advertised `1.0.0` instead of the running development build.

Closes #1365

## How Has This Been Tested?
- `go test ./... -count=1`
- `go test -race ./... -count=1`
- `go test ./cmd/litestream -run "Test(MCPServerInitializeVersion|VersionTool)$" -count=1 -ldflags "-X main.Version=v1365-test"`

## Types of changes

## Checklist

### DIFF
--- a/cmd/litestream/mcp.go
+++ b/cmd/litestream/mcp.go
@@ -30,7 +30,7 @@ func NewMCP(ctx context.Context, configPath string) (*MCPServer, error) {
 
 	mcpServer := server.NewMCPServer(
 		"Litestream MCP Server",
-		"1.0.0",
+		Version,
 		server.WithToolCapabilities(false),
 		server.WithRecovery(),
 		server.WithLogging(),
@@ -243,16 +243,11 @@ func RestoreTool(configPath string) (mcp.Tool, server.ToolHandlerFunc) {
 
 func VersionTool() (mcp.Tool, server.ToolHandlerFunc) {
 	tool := mcp.NewTool("litestream_version",
-		mcp.WithDescription("Print the Litestream binary version."),
+		mcp.WithDescription("Print the running Litestream instance's version."),
 	)
 
-	return tool, func(ctx context.Context, req mcp.CallToolRequest) (*mcp.CallToolResult, error) {
-		cmd := exec.CommandContext(ctx, "litestream", "version")
-		output, err := cmd.CombinedOutput()
-		if err != nil {
-			return mcp.NewToolResultError(strings.TrimSpace(string(output)) + ": " + err.Error()), nil
-		}
-		return mcp.NewToolResultText(string(output)), nil
+	return tool, func(context.Context, mcp.CallToolRequest) (*mcp.CallToolResult, error) {
+		return mcp.NewToolResultText(Version + "\n"), nil
 	}
 }
 
--- a/cmd/litestream/mcp_test.go
+++ b/cmd/litestream/mcp_test.go
@@ -1,0 +1,70 @@
+package main
+
+import (
+	"encoding/json"
+	"net/http"
+	"net/http/httptest"
+	"strings"
+	"testing"
+
+	"github.com/mark3labs/mcp-go/mcp"
+)
+
+func TestMCPServerInitializeVersion(t *testing.T) {
+	server, err := NewMCP(t.Context(), "")
+	if err != nil {
+		t.Fatal(err)
+	}
+	httpServer := httptest.NewServer(server)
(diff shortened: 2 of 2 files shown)


corylanou:
Closing as superseded by #1378.

#1378 migrated `cmd/litestream/mcp.go` to the official `modelcontextprotocol/go-sdk` as an independent rewrite from `main`, and in doing so it already implements both of this issue's acceptance criteria: `mcp.Implementation` is constructed with the in-process `Version` rather than the hardcoded `"1.0.0"`, and `VersionTool` returns `Version` directly with no subprocess. Its test suite covers both — the handshake `ServerInfo.Version` and the `litestream_version` tool result, asserted through a real client session against an injected version.

Because both branches rewrote the same functions, this one now conflicts with #1378 for no remaining gain, so it is cheaper to close than to reconcile.

Two test nuances here are not exactly duplicated on #1378 and are worth copying into the surviving suite: this PR empties `PATH` before invoking `VersionTool`, which proves the absence of a subprocess dependency as an explicit precondition, and it exercises a raw `initialize` at protocol version `2025-03-26`.

Correction to an earlier version of this comment: it claimed #1365 was only partly delivered because `InfoTool` still runs `litestream version` via PATH. That was wrong about scope. #1365 scopes itself explicitly to the `litestream_version` tool and the handshake identity, and states that #1367 owns converting the other tools. Both acceptance criteria are delivered by #1378, so #1365 is complete once that lands. The `InfoTool` occurrence is a real but separate PATH-version bug; it is fixed in #1376.

PR #1385, opened 2026-07-20
fix(restore): resume follow mode when saved TXID is ahead of latest s…

atmin:
## Description

Follow-mode crash-recovery resume validated the saved `-txid` against the latest snapshot's `MaxTXID` and refused to resume when the saved TXID was greater. Compare against the newest TXID across *all* levels instead, so a follow file that is ahead of the last snapshot but still backed by existing deltas resumes correctly. A stale/foreign `-txid` past all replicated data is still rejected; the "behind the earliest snapshot" check is unchanged.

## Motivation and Context

A live follow file continuously applies level-0 deltas past the latest snapshot, while snapshots are only written every `SnapshotInterval` (default 24h). So the saved `-txid` is almost always ahead of the last snapshot, and resume was rejected in the normal case — including `litestream restore -follow` crash recovery — forcing a full re-restore instead of an incremental catch-up.

## How Has This Been Tested?

New regression test `TestReplica_Restore_Follow_ResumeAheadOfSnapshot` (fails before this change with `saved TXID … is ahead of latest snapshot`, passes after):

```
go test -race -run 'TestReplica_Restore_Follow' .
```

All pass locally (full race suite green across all packages; `go fmt`, `go vet`, `goimports`, `staticcheck` on the changed files clean).

## Types of changes

## Checklist

### DIFF
--- a/replica.go
+++ b/replica.go
@@ -652,8 +652,20 @@ func (r *Replica) Restore(ctx context.Context, opt RestoreOptions) (err error) {
 				if latestSnapshot.MinTXID > txid {
 					return fmt.Errorf("cannot resume follow mode: saved TXID %s is behind the earliest snapshot (min TXID %s); replica history has been pruned -- delete %s and %s-txid to re-restore", txid, latestSnapshot.MinTXID, opt.OutputPath, opt.OutputPath)
 				}
-				if txid > latestSnapshot.MaxTXID {
-					return fmt.Errorf("cannot resume follow mode: saved TXID %s is ahead of latest snapshot (max TXID %s); delete %s and %s-txid to re-restore", txid, latestSnapshot.MaxTXID, opt.OutputPath, opt.OutputPath)
+
+				// A follow file legitimately runs ahead of the latest snapshot, so bound the check by the newest TXID across all levels.
+				var maxTXID ltx.TXID
+				for level := 0; level <= SnapshotLevel; level++ {
+					info, infoErr := r.MaxLTXFileInfo(ctx, level)
+					if infoErr != nil {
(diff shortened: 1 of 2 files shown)


strefethen:
Hit this in production on v0.5.11 today, and confirmed the same guard is unchanged in v0.5.17 — +1 for getting this merged and released.

**Setup:** `litestream replicate` on the primary writes to a `file://` replica; a separate host runs a follower:

```
litestream restore -f -o /var/lib/app/db.sqlite file:///srv/litestream-replica/db
```

**What happened:** after a host reboot we removed our `ExecStartPre` that deletes the database before each start, expecting the v0.5.10 `-txid` crash-recovery resume to work. The `-txid` sidecar was present and current, but the restart failed immediately:

```
cannot resume follow mode: saved TXID 00000000000102ad is ahead of latest snapshot (max TXID 000000000000fda8); delete /var/lib/app/db.sqlite and /var/lib/app/db.sqlite-txid to re-restore
```

The follower had legitimately applied L0–L3 deltas past the last snapshot, and those deltas were still present in the replica, so an incremental catch-up was possible. The check only compares against the latest snapshot's `MaxTXID`, so resume fails in the normal case, as the PR description says.

**Impact:** our database is ~11 GB, so every follower restart becomes a full re-restore. That's ~3 minutes on healthy NVMe, but today, with a degraded disk, the restore ran at under 1 MB/s and would have taken hours with no readable replica. We've kept `ExecStartPre=-/bin/rm -f <db> <db>-wal <db>-shm` as a workaround, because without it the follower can't recover on its own after a restart.

Comparing against the newest TXID across all levels, as this PR does, matches what we'd expect. Happy to test a release build against our setup.

PR #1389, opened 2026-07-23
Propagate AWS_SESSION_TOKEN to AWS client

amzn-daadelbe:
## Description
AWS can issue temporary credentials (see [the docs](https://docs.aws.amazon.com/IAM/latest/UserGuide/id_credentials_temp_use-resources.html)). However, the current implementation doesn't pass-through AWS_SESSION_TOKEN, which is required when using temporary credentials.

## Motivation and Context
Since some users may be using AWS temporary credentials, `litestream` should propagate `AWS_SESSION_TOKEN` alongside ACCESS_KEY_ID and SECRET_ACCESS_KEY.

## How Has This Been Tested?
I have tested this in an AWS Lambda that uses temp credentials to provision

## Types of changes

## Checklist

### DIFF
--- a/s3/replica_client.go
+++ b/s3/replica_client.go
@@ -86,6 +86,7 @@ type ReplicaClient struct {
 	// AWS authentication keys.
 	AccessKeyID     string
 	SecretAccessKey string
+	SessionToken    string
 
 	// S3 bucket information
 	Region            string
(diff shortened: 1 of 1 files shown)


s3/replica_client.go
```diff
@@ -227,6 +228,11 @@ func NewReplicaClientFromURL(scheme, host, urlPath string, query url.Values, use
 	} else if v := os.Getenv("LITESTREAM_SECRET_ACCESS_KEY"); v != "" {
 		client.SecretAccessKey = v
 	}
+	if v := os.Getenv("AWS_SESSION_TOKEN"); v != "" {
+		client.SessionToken = v
+	} else if v := os.Getenv("LITESTREAM_SESSION_TOKEN"); v != "" {
```
corylanou, line 233: Can you drop the LITESTREAM_SESSION_TOKEN fallback and keep just the AWS_SESSION_TOKEN read? Session tokens are AWS STS specific, so there's no reason for a Litestream-prefixed alias.

corylanou (changes requested):

PR #1396, opened 2026-07-25
fix(ci): stabilize nightly workflow

corylanou:
## Description

Stabilizes the nightly workflow by fixing its accumulated race-test, MinIO harness, high-volume disk-growth, snapshot-classification, and failure-notification problems.

- restores the prior default logger after the SFTP host-key test and synchronizes captured log access
- configures each test Store before opening its DB so logger configuration cannot race the DB monitor
- captures the MinIO container ID from stdout only and runs `minio/mc` in the MinIO container's network namespace
- retains the 500 writes/sec high-volume profile while reducing its payload from 4KB to 1KB
- records checkpoint mode and database identity in behavioral events and fails closed on malformed or unattributable checkpoint-boundary snapshots
- requires a valid same-database `TRUNCATE` checkpoint to precede a boundary snapshot in log order and fall within the five-second attribution window
- validates every paginated job independently so malformed jobs become synthetic failures without hiding recognized failures or throwing from the notifier

The snapshot exception is deliberately narrow. PR #1292 made a full L0 `checkpoint boundary snapshot` required for emergency `TRUNCATE` checkpoints. The assertion permits that reason only when an earlier log event is a valid `TRUNCATE` checkpoint for the same database and its timestamp is within five seconds. Missing database identity, cross-database checkpoints, later-in-log checkpoints, missing modes, malformed timestamps, unsupported modes, `PASSIVE` checkpoints, missing preceding checkpoints, and differently reasoned checkpoint-adjacent snapshots still fail. Issue #1198 remains open for the cost of those intentional full L0 snapshots.

## Motivation and Context

The 25 most recent `nightly-stability` runs were failing. Evidence run [30065650568](https://github.com/benbjohnson/litestream/actions/runs/30065650568) showed four independent harness failures:

- the SFTP host-key test left a mutable global slog handler installed, and later background logging raced on its captured slice
- MinIO startup combined Docker pull stderr with stdout, producing an invalid container ID that was passed to the legacy `--link` flag
- the 500 writes/sec high-volume profile used 4KB payloads, growing the source, replica, and restore working set until the runner ran out of disk
- the notifier's `github.paginate()` mapping callback returned an undefined jobs collection, then dereferenced `job.conclusion`

Once disk exhaustion was fixed, the high-volume lane also exposed that its older snapshot assertion did not distinguish PR #1292's intentional emergency-TRUNCATE boundary from the unwanted snapshot-on-checkpoint behavior tracked by #1198.

[...]

PR #1397, opened 2026-07-25
fix(store): preserve final sync during shutdown

corylanou:
## Description

Preserves the final database flush when Store-owned monitors are slow or ignore cancellation, while eliminating the snapshot encoder race with database handle cleanup.

- cancels Store monitors immediately, then lets each database complete its final WAL-to-LTX sync, replica sync, and handle cleanup before waiting for those monitors
- makes the post-flush monitor wait interruptible by `Close(ctx)` or the CLI second-signal channel
- registers snapshot readers while holding the DB execution semaphore
- closes and joins every active snapshot encoder before final sync and before `DB.Close()` clears `db.f`
- closes and joins the encoder when a snapshot backend returns early, whether it returns success or failure

## Motivation and Context

A full race sweep while validating PR #1396 caught `DB.Close()` clearing database fields while `monitorCompactionLevel()` was still taking a snapshot. The initial fix moved `s.wg.Wait()` before `DB.Close()`, but cold review found that ordering introduced a durability regression: file fsync and SFTP operations can ignore cancellation, so shutdown could block before the final WAL-to-LTX and replica sync. A second signal could not interrupt that wait.

The review also found that waiting for the Store monitor did not completely eliminate the original race. `snapshotReader` launches an encoder goroutine that was not joined when a backend returned early; the monitor could finish while that encoder still used `db.f`.

Both major findings and the test-coverage finding were confirmed. The revised design does not make final-flush correctness depend on a backend monitor stopping:

1. Store cancellation prevents normal new work.
2. `DB.Close()` acquires the DB execution semaphore, closes all registered snapshot pipes, and joins their encoders.
3. The database performs its final local and replica sync and clears its handles.
4. Store waits for any remaining backend monitor work only after the durability work is complete; caller cancellation or a second signal can interrupt that wait.

This deliberately avoids placing an unbounded backend wait in front of the final flush. Discovered while validating #1396.

## How Has This Been Tested?

The regressions fail on the previous PR head `4364eab`:

- an early-success snapshot backend lets `Snapshot` return while its encoder still holds the checkpoint lock
- a non-cancelable snapshot upload prevents `Store.Close()` from reaching the final sync

The replacement Store regression runs through the real snapshot monitor and `snapshotReader`, leaves the backend blocked after cancellation, writes pending WAL data, and verifies that shutdown still advances both the local and replica TXIDs before a second signal interrupts the remaining monitor wait.

```bash
go test -race -count=10 -run '^(TestDB_SnapshotJoinsEncoderOnEarlyReplicaReturn|TestStore_CloseFlushesBeforeInterruptibleMonitorWait|TestStore_WaitForMonitorsHonorsContext)$' .
go test -race -count=10 -timeout 45m ./...
pre-commit run --all-files
go build -o bin/litestream-test ./cmd/litestream-test
go test -tags 'integration,soak' -run '^TestLTXBehavior$' -short -v -timeout 10m ./tests/integration/
```

The low-, high-, and burst-volume behavioral profiles all passed shutdown, restoration, and integrity validation.

## Types of changes

## Checklist

- [x] I have updated the documentation accordingly (not needed; shutdown contract is unchanged)

### DIFF
--- a/db.go
+++ b/db.go
@@ -79,6 +79,11 @@ type DB struct {
 	syncState syncState
 	syncDiag  diagState
 
+	snapshotReaders struct {
+		sync.Mutex
+		m map[*snapshotReadCloser]struct{}
+	}
+
 	// last file info for each level
 	maxLTXFileInfos struct {
 		sync.Mutex
@@ -326,6 +331,7 @@ func NewDB(path string) *DB {
 		Logger:               slog.With(LogKeyDB, filepath.Base(path)),
 	}
 	db.maxLTXFileInfos.m = make(map[int]*ltx.FileInfo)
+	db.snapshotReaders.m = make(map[*snapshotReadCloser]struct{})
 	db.openLTXFile = defaultOpenLTXFile
 
 	db.dbSizeGauge = dbSizeGaugeVec.WithLabelValues(db.path)
@@ -828,6 +834,8 @@ func (db *DB) Close(ctx context.Context) (err error) {
 		return err
 	}
 	defer db.execSem.Release(1)
+
+	db.closeSnapshotReaders()
 
 	// Perform a final db sync, if initialized.
 	if db.db != nil {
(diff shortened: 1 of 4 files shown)


MAxentum:
I reproduced an SFTP upload hanging past its context deadline on v0.5.17. Runnable reproduction attached.

I tested an AI-assisted transport-abort fix alongside your c4a2240 changes. The combined default suite, focused race tests, and short restore/integrity soak passed.

Would you prefer a complementary PR after this lands, or a patch here?

[litestream-sftp-repro.zip](https://github.com/user-attachments/files/32131255/litestream-sftp-repro.zip)

MAxentum:
I've put the tested combination on a [branch in my fork](https://github.com/MAxentum/litestream/tree/codex/sftp-shutdown-integration) so the code is available for review. It includes your c4a2240 snapshot-lifetime and monitor-wait changes, adapted to current main, plus the SFTP transport-abort work. [Combined diff](https://github.com/MAxentum/litestream/compare/1be2782b642111afbb9fb06d6ea10a211e6a0b84...b15e29c62af15298945c8ffec485533d38eb331a).

No separate PR opened; happy to split out the transport additions or adapt the integration to your preferred approach.

PR #1423, opened 2026-08-04
Resolve the default config path in the reset command (fixes #1404)

### LINKED ISSUE #1404: reset: ignores LITESTREAM_CONFIG and the default config path
## Summary

Every other config-backed subcommand resolves its configuration through `DefaultConfigPath()`, which honors `LITESTREAM_CONFIG` and falls back to `/etc/litestream.yml`. `reset` does not. It reads the config only when `-config` is passed explicitly, so both the environment variable and the default path are ignored — while its own help text advertises the default path.

## Reproduction

```
$ cat c.yml
dbs:
  - path: /tmp/x.db
    replica:
      path: /tmp/xrep

$ LITESTREAM_CONFIG=$PWD/c.yml litestream databases
path       replica
/tmp/x.db  file

$ LITESTREAM_CONFIG=$PWD/c.yml litestream reset -dry-run /tmp/x.db
Error: database does not exist: /tmp/x.db
```

`databases` resolves the database from the config. `reset` does not see it at all.

## Cause

`cmd/litestream/reset.go`:

```go
// Load configuration to find the database (if config exists)
var dbConfig *DBConfig
if *configPath != "" {
    config, configErr := ReadConfigFile(*configPath, !*noExpandEnv)
    ...
}
```

The gate is on the raw flag value rather than on a resolved path. Compare with `databases.go`, `status.go`, `replicate.go`, `restore.go` and `ltx.go`, which all fall back to `DefaultConfigPath()`.

## Suggested fix

Resolve through `DefaultConfigPath()` when `-config` is empty, matching the other subcommands.

Found while investigating #1400.

youdie006:
## Description

`reset` only loaded a Litestream config when `-config` was passed explicitly (`cmd/litestream/reset.go:48`), so it ignored `LITESTREAM_CONFIG` and the default config path. Every other subcommand resolves `DefaultConfigPath()` when `-config` is empty (`databases.go`, `status.go`, `replicate.go`, `restore.go`, `ltx.go`), and `DefaultConfigPath()` honors `LITESTREAM_CONFIG` before falling back to `/etc/litestream.yml`. Because `reset` skipped that step, `dbConfig` stayed `nil` and it operated on the default meta path via `litestream.NewDB(dbPath)` instead of the `meta-path` configured for the database (`NewDBFromConfig`).

This resolves the default config path when `-config` is not given, matching the other subcommands, but tolerates a missing config file (`errors.Is(err, ErrConfigFileNotFound)`) so `reset` still works on a database that is not listed in any config.

**Scope:** changes only the config-path resolution in `reset`. Behavior with no config file present is unchanged, and the other subcommands are untouched.

## Motivation and Context

`reset` even advertises the default config path in its own help text, but ignored it — `LITESTREAM_CONFIG=cfg litestream reset <db>` never saw the config.

Fixes #1404

## How Has This Been Tested?

Added `TestResetCommand_RunWithConfigEnv` (`cmd/litestream/reset_test.go`): it places the Litestream state at a non-default `meta-path` discoverable only through a config file, sets `LITESTREAM_CONFIG`, runs `reset` without `-config`, and asserts the state at the configured meta path is removed.

```
go test ./cmd/litestream/ -run TestResetCommand -v
```

Verified red/green: the new test fails on `main` (config ignored → default meta path used) and passes with this change. `go build ./...`, `go vet ./cmd/litestream/`, and `gofmt` are clean (go 1.25).

## Types of changes

## Checklist
- [x] I have tested my changes (`go test ./cmd/litestream/`)

---
AI disclosure: prepared with AI assistance (Claude Code); the bug was traced against the code, the fix derived, and the red/green test verified before submitting.


corylanou (changes requested):
Thanks for addressing #1404. This PR correctly makes reset honor LITESTREAM_CONFIG and the implicit default config path.

There is one behavior change that needs adjustment: ErrConfigFileNotFound is currently ignored regardless of how the path was selected. As a result, an explicitly missing -config file, which previously returned an error, now silently falls back to the default metadata path.

Please preserve an error for an explicitly supplied -config path. A missing path supplied through LITESTREAM_CONFIG should also return an error once we begin honoring that variable. Only a missing implicit platform default should be tolerated, preserving standalone litestream reset <db> behavior.

Please add tests covering these cases:

  - Missing explicit -config returns an error.
  - Missing LITESTREAM_CONFIG path returns an error.
  - Missing implicit platform default retains standalone reset behavior.

youdie006:
Thanks for the review — you were right, and the consequence was worse than a silent fallback. I probed it before and after:

```
before:  run_err=<nil>                                   ltx_exists=false
after:   run_err=cannot read config: config file not found: .../missing.yml   ltx_exists=true
```

With a mistyped `-config` path the command reported success and removed the local LTX state, because by the time `ErrConfigFileNotFound` was tested, `*configPath` had already been overwritten with `DefaultConfigPath()` and all provenance was gone.

## What changed

Config resolution now distinguishes how the path was selected:

| path source | missing file |
|---|---|
| explicit `-config` | error |
| `LITESTREAM_CONFIG` | error |
| implicit platform default | tolerated |

`resolveConfigPath` in `cmd/litestream/main.go` returns the path along with whether it was explicitly chosen, and `DefaultConfigPath` delegates to it, so `LITESTREAM_CONFIG` is still read in exactly one place. `reset.go` becomes a one-line predicate change.

On your "use an existing helper if there is one" — there wasn't. Every other subcommand (`databases.go:28-34`, `status.go:30-36`, `replicate.go:84-89`, `restore.go:107-110`, `ltx.go:48-56`) uses the same two-step `if *configPath == "" { *configPath = DefaultConfigPath() }` idiom and then hard-fails, so none of them needed the distinction. I added the helper rather than re-reading `LITESTREAM_CONFIG` inside `reset.go`, which would have let the two definitions drift. That does put 14 lines in `main.go`, outside this PR's original two files — if you would rather keep the diff strictly inside `reset.go`, that is a two-line change and I will redo it.

The other five subcommands are unaffected: same branch order, same values, and their tests pass.

## Tests

The three cases you listed:

- `TestResetCommand_RunWithMissingConfigFlag`
- `TestResetCommand_RunWithMissingConfigEnv`
- `TestResetCommand_RunWithMissingDefaultConfig`

The first two assert `ErrConfigFileNotFound` *and* that the LTX file survives, so they also fail if the command errors after doing work. The third passes both before and after by design — it is the no-regression guard for standalone `reset`.

Verified red then green: with `reset.go` reverted and the tests kept, the two error cases fail with `expected ErrConfigFileNotFound, got <nil>` while the default-path case still passes; all six pass with the fix. `t.Setenv` throughout for env isolation.

`go test` over CI's package list is green, and `go vet`, `gofmt`, `goimports -local github.com/benbjohnson/litestream` and `staticcheck` are all clean. Not covered here: the Windows default path (`main_windows.go:17`) — `DefaultConfigPath()` semantics are unchanged so the risk is low, but this environment is Linux-only.

Disclosure: I use AI assistance in my work, and I review and verify everything before it goes out. The probe and the red-green runs above are my own.

PR #1440, opened 2026-08-16
docs: document backend-specific replica layouts

### LINKED ISSUE #1426: file:// and s3:// replicas use different object layouts; a file replica copied to a bucket is unrestorable
Version: litestream v0.5.16 (darwin-arm64), DigitalOcean Spaces (S3-compatible).

The file:// and s3:// replica clients use different object layouts:

- file replica: `<path>/ltx/<level>/<min16>-<max16>.ltx` (bare level numbers: `ltx/0/`, `ltx/1/`, `ltx/9/`)
- s3 replica: `<path>/<level as %04d>/<min16>-<max16>.ltx` — no `ltx/` segment (observed from a real replication: `.../0001/0000000000000001-0000000000000001.ltx`, `.../0009/...`)

Consequence: a file replica copied/rcloned into a bucket cannot be restored — `litestream restore s3://...` reports `no matching backup files available` for a tree its own `restore file://...` accepts. Verified both directions: re-keying the same LTX files into the `%04d` layout makes the S3 restore succeed (byte-identical output database).

Expected: either one layout for both replica types, or documentation stating they are not interchangeable (the docs currently don't mention the difference).

Repro:
1. `litestream replicate -exec "sleep 3" t.db file:///tmp/rep` → note `/tmp/rep/ltx/0/...`
2. Copy `/tmp/rep` recursively to `s3://bucket/path` (aws cli/rclone).
3. `litestream restore -o out.db s3://bucket/path` → "no matching backup files available".
4. Re-upload the same files as `s3://bucket/path/0000/<name>.ltx` → restore succeeds.

corylanou:
## Summary

Documents that `file://` and `s3://` replicas use different established raw storage layouts and cannot be moved between backends with a recursive object copy.

## Problem

Issue #1426 reports that a file replica copied directly into an S3 bucket cannot be restored through the S3 client. A reproduction from current `main` confirmed these paths for the same database:

| Backend | Observed LTX path |
|---|---|
| `file://` | `<root>/ltx/<unpadded decimal level>/<min16>-<max16>.ltx` |
| `s3://` | `<prefix>/<four-digit lowercase hexadecimal level>/<min16>-<max16>.ltx` |

The concrete file replica contained `ltx/0/`, `ltx/1/`, and `ltx/9/`, while native S3 replication contained `0000/`, `0001/`, and `0009/`. Both use the same 16-digit hexadecimal LTX filename, and neither writes a separate replica index or manifest object.

Changing either writer would make existing replicas invisible to its reader, so normalizing the raw layout would be a breaking storage change.

## Solution

Document both layouts and the supported migration path: restore the file replica locally, then create a new S3 replica from the restored database.

No runtime behavior, write path, or existing object name changes.

## Scope

**In scope:**
- Document the file and S3 raw storage paths.
- Document that raw replica trees are backend-specific.
- Document the supported migration path.

**Not in scope:**
- Automatic runtime layout detection.
- A new doctor or replica-copy command.
- Changing or normalizing existing backend layouts.

## Test Plan

- `go test ./s3 -count=1`
- `go build ./...`
- `go test ./...`
- `go test -race ./...`
- Full pull request CI, including storage integration and the LTX behavioral gate

## Related

Fixes #1426


corylanou:
## Design opinion: read-side fallback

I agree with rejecting a transparent read-side fallback for the current read/write S3 client.

A correct fallback could not stop at listing `ltx/<decimal-level>/`; `OpenLTXFile` plus retention and deletion would also need layout-aware routing. More importantly, after the first new write, the bucket would contain both the copied file layout and native `<prefix>/<four-digit lowercase hex level>/` objects. Those trees could contain overlapping transaction ranges or diverging histories, while compaction and retention would have no durable marker identifying which tree is authoritative. Restore and cleanup behavior would therefore become ambiguous.

The rejection would be too broad only if the fallback were strictly read-only and all writes, compaction, retention, and deletion were prohibited, or if an explicit one-time import validated and migrated the entire replica before enabling writes. Either approach is a separate migration feature. For the current client, the documented incompatibility plus an actionable restore-locally/re-replicate error is the safer behavior.

benbjohnson (changes requested):
I don't think we should add a call on s3.fileIterator to check it every time the snapshot layer is checked the first time. Copying from disk to S3 is fairly rare.

I would suggest two options:

1. Add a `litestream doctor` (or similarly named command) that would do some validation across Litestream and its replicas to find any potential issues.

2. Add a `litestream replica copy` command that could create two separate ReplicaClients and copy from one to the other correctly.

corylanou:
Addressed in 149f15e: removed the automatic file-layout probe from S3 snapshot iteration, so routine reads no longer issue an extra compatibility LIST request. The verified backend-layout and migration guidance remains documented. A broader doctor or replica-copy command can be designed separately without expanding this focused change.

benbjohnson (approved):

Issue #1400, opened 2026-07-27
Thoughts from a first-time user


BroadlyWhitaker:
Hi,

I just found litestream and have been playing with it for a couple of hours. 
Overall, I quite like it. It's a single binary, was very easy to get working, and it does what it says on the packaging.

But, having used it briefly, and using the docs to try and understand what was unclear I thought I'd provide some feedback on things I disliked, from trivial to not so trivial. It's not meant to put the project down, it's just that new users experience a tool in a way which other can't, once they get used to the quirks. Some of these may be trivially fixable, other may not be trivial but well worth addressing. Again, I like the project and find it useful.

Here goes, in no particular order:

# CLI arg handling:  -cfg

due to Go quirky cli argument handling, I can only pass -config after I specify a
sub-command. But this means an alias like
alias lt='my/local/litestream -config ~/my_local_config'
doesn't work. 
this makes working in the shell tedious if the config file is not stored 
in the default location. litestream is not inherently a privilged
program, why require write access to /etc? (I care more about the alias working, really)

# global settings should not pretend to be per-db settings

The "Per-database snapshot settings" section of 
https://litestream.io/reference/config/ 
says

"""
A snapshot block can be set on an individual database, but snapshot settings are ultimately applied globally. The per-database block is a convenience for expressing the global snapshot configuration from the database level:
"""

This is confusing and redundant. Since "the effective snapshot schedule is global", i.e shared by all databased, these settings should only be settable in the global scope of the config. 
It's confusing and counter-productive to introduce needlessly complex "promotion semantics" for no reason.

# Configuration default values are not documented

Neither the settings docs, nor the complete config example in the config docs, nor the example config included in the binary releases, document what the defaults are for configuration values.

For example, the "complete config example" in the config docs lists 5m as the compaction interval for level1,
but when you run the tool without overriding the value, you get:
```
level=INFO msg="starting compaction monitor" system=store level=1 interval=30s
```

specific defaults are mentioned in some places in the documentation, but these are handwritten, non-exhaustive, and do not derive from code so may lose sync if and when defaul values are changed.

# Compaction triggers at unexplained times

```
time=2026-07-27T17:28:21.158+03:00 level=INFO msg="starting compaction monitor" system=store level=1 interval=30s
time=2026-07-27T17:28:21.158+03:00 level=INFO msg="starting compaction monitor" system=store level=2 interval=5m0s
[...]

PR #1297, opened 2026-05-28
fix(cli): polish agent-facing errors

corylanou:
## Description
Polishes the agent-facing CLI behavior from #1296:

- Emits `[]` instead of `null` for empty top-level JSON arrays from `databases`, `status`, and `ltx`.
- Renders top-level CLI user errors as plain stderr output with optional `Try:` hints instead of slog-framed stdout records.
- Adds required-argument hints for `register`, `unregister`, `start`, `stop`, `sync`, `ltx`, `reset`, and `restore`.
- Normalizes idempotent runtime statuses to `already_<state>` for register/unregister, matching the existing start/stop convention.
- Removes stale `ltx -replica` help text and documents the control socket config in `replicate -help`.
- Makes explicit subcommand help exit successfully.

## Motivation and Context
Issue #1296 verified that the #1232 CLI surface mostly works, but a few edges still made scripted and agent usage inconsistent.

Observed before this PR:

```bash
litestream databases -json -config empty.yml
# null

litestream register >out 2>err || true
# out contained: time=... level=ERROR msg="failed to run" error="database path required"
# err was empty

litestream ltx -help
# documented -replica even though the command does not register that flag
```

Verified after this PR with a built binary:

```bash
databases-json=[]
status-json=[]
ltx-json=[]
register-stdout-bytes=0
register-stderr=Error: database path required
Try: litestream register -replica s3://bucket/prefix /path/to/db
register-help-exit=0 stdout-has-usage=yes stderr-bytes=0
ltx-help-replica=absent
replicate-help-socket=yes
```

Fixes #1296

Closes #1232 — this PR completes the final error-surface items in the tracking checklist; the deferred resource+verb RFC was spun out to its own design issue (#1298).

## Scope
**In scope:**
- CLI output/error/help polish requested in #1296.
- Control socket response status string normalization for register/unregister no-op cases.
- JSON output docs for the idempotent status convention.
- Regression tests for the visible CLI contracts.

**Not in scope:**
- Changes to replication, restore mechanics, storage backends, or LTX file formats.
- Broader lint cleanup unrelated to this CLI issue.

## How Has This Been Tested?

```bash
[...]

PR #1453, opened 2026-08-19
cli: explain misplaced flags before the subcommand

### LINKED ISSUE #1405: cli: flags before the subcommand print bare usage with no explanation
## Summary

`litestream -config foo.yml replicate` prints the bare top-level usage screen and exits 1, with no message explaining that the flag is in the wrong position. Users reasonably read that as "the flag does not exist".

From #1400:

> due to Go quirky cli argument handling, I can only pass -config after I specify a sub-command. But this means an alias like `alias lt='my/local/litestream -config ~/my_local_config'` doesn't work.

## Reproduction

```
$ litestream -config ./c.yml databases

litestream is a tool for replicating SQLite databases.

Usage:

	litestream <command> [arguments]
...
$ echo $?
1
```

## Cause

`Main.Run()` consumes `args[0]` as the subcommand before any flag parser runs. When that token starts with `-`, it falls through to the `m.Usage(); return flag.ErrHelp` branch. Nothing distinguishes "unknown flag" from "misplaced flag" from "no arguments at all".

## Suggested fix

When the first token starts with `-` and is not a help flag, emit a targeted error instead of bare usage:

```
litestream: flags must come after the subcommand
  try: litestream databases -config ./c.yml
  or:  LITESTREAM_CONFIG=./c.yml litestream databases
```

Mentioning `LITESTREAM_CONFIG` here matters — it is what actually solves the reporter's aliasing use case, and it is currently only documented inside an environment-variable table that is otherwise all credentials.

Note their stated premise about `/etc` is mistaken (Litestream only reads the config at runtime), but the ergonomic complaint stands on its own.

Related: #1400

ekanshul:
## Summary
Replace the bare usage screen printed for `litestream -config c.yml databases` with a targeted error that says flags go after the subcommand, and points `-config` users at `LITESTREAM_CONFIG`.

## Problem
`Main.Run()` takes `args[0]` as the subcommand before any flag parser runs. A leading flag falls into the `m.Usage(); return flag.ErrHelp` branch, so the user gets the full command list on stdout, exit 1, and nothing on stderr. Users read that as "the flag does not exist" (#1405, #1400).

Before (origin/main 63225f1):

```
$ litestream -config /nonexistent.yml databases

litestream is a tool for replicating SQLite databases.

Usage:

	litestream <command> [arguments]
... (full command list) ...
$ echo $?
1
```

After:

```
$ litestream -config /nonexistent.yml databases
Error: flags must come after the subcommand
Try: litestream <command> -config PATH
 or: LITESTREAM_CONFIG=PATH litestream <command>
$ echo $?
1
```

Any other leading flag gets the neutral form:

```
$ litestream -socket status info
Error: flags must come after the subcommand
Try: litestream <command> [flags]
```

## Solution
In `Run()`, a first token that starts with `-` and is not `-h`/`-help`/`--help` now returns a `usageError` (the existing `Error:` / `Try:` mechanism from #1297) instead of printing usage. `misplacedFlagError` looks only at that token: if it names `config` it adds an `or:` line using `LITESTREAM_CONFIG`, which is what solves the shell-alias case in #1400; otherwise the hint is a fixed form.

The hint deliberately does not try to reconstruct the user's command. `Run()` is not the flag parser, so it cannot tell a command from a flag value: in `litestream -socket status info`, `status` may be the value of `-socket` and `info` the intended command. An earlier revision of this PR scanned for a known command name and reordered the arguments, which reversed that case and needed its own copy of the command list. Empty args, `-h`/`-help`/`--help`, `help`, unknown commands, and flags after the subcommand behave exactly as before.

## Scope
**In scope:**
- `cmd/litestream/main.go`: split the existing branch in `Run()`, plus `misplacedFlagError`
- Unit test in `main_test.go` and subprocess CLI tests in `main_cli_test.go`

**Not in scope:**
- Accepting flags before the subcommand. Making `-config` a real root flag with a top-level `flag.FlagSet` is a separate change.
- Documenting `LITESTREAM_CONFIG` more prominently on the website (separate repo)

## Test Plan
```bash
gofmt -l ./cmd/litestream
go vet ./cmd/litestream/...
go test -race -count=1 -run 'TestMainFlagPlacement|TestMain_RunHelp' ./cmd/litestream/... -v
go test -race -count=1 ./cmd/litestream/...
```
`TestMainFlagPlacement` covers `-config PATH`, `--config=PATH`, another leading flag, `-h`/`-help`/`--help`, no arguments, and a correctly positioned flag. All pass locally (macOS, Go 1.26; `staticcheck` and `goimports` clean, `uvx pre-commit run --all-files` clean).

## Related
- Fixes #1405
- Related to #1400

This change was made with AI assistance (Claude Code) and reviewed by the author.

### DIFF
--- a/cmd/litestream/main.go
+++ b/cmd/litestream/main.go
@@ -14,6 +14,7 @@ import (
 	"path"
 	"path/filepath"
 	"runtime/debug"
+	"slices"
 	"strconv"
 	"strings"
 	"time"
@@ -230,12 +231,57 @@ func (m *Main) Run(ctx context.Context, args []string) (err error) {
 		if cmd == "help" || cmd == "-h" || cmd == "-help" || cmd == "--help" {
 			m.Usage()
 			return nil
-		} else if cmd == "" || strings.HasPrefix(cmd, "-") {
+		} else if cmd == "" {
 			m.Usage()
 			return flag.ErrHelp
+		} else if strings.HasPrefix(cmd, "-") {
+			return misplacedFlagsError(append([]string{cmd}, args...))
 		}
 		return fmt.Errorf("litestream %s: unknown command", cmd)
 	}
+}
+
+// commandNames lists the subcommands dispatched by Main.Run.
+var commandNames = []string{
+	"databases", "info", "list", "ltx", "register", "replicate", "reset",
+	"restore", "start", "status", "stop", "sync", "unregister", "version",
+}
+
+// misplacedFlagsError returns the error for flags passed before the
+// subcommand, e.g. "litestream -config c.yml databases". Flags are only
+// parsed by subcommands, so the hint reorders the user's own arguments into
+// a form that can be pasted back. When -config is the misplaced flag it also
+// offers LITESTREAM_CONFIG, which is what works for shell aliases.
+func misplacedFlagsError(args []string) error {
+	var cmd, configPath string
+	var rest, envRest []string // envRest omits the -config flag.
+	for i := 0; i < len(args); i++ {
+		arg := args[i]
+		if cmd == "" && slices.Contains(commandNames, arg) {
+			cmd = arg
+			continue
+		}
+		rest = append(rest, arg)
+		if name := strings.TrimLeft(arg, "-"); name == "config" && i+1 < len(args) {
+			i++
+			configPath = args[i]
+			rest = append(rest, configPath)
+		} else if strings.HasPrefix(name, "config=") {
+			configPath = strings.TrimPrefix(name, "config=")
+		} else {
+			envRest = append(envRest, arg)
+		}
+	}
+	if cmd == "" {
+		cmd = "<command>"
+	}
+
+	hint := strings.Join(append([]string{"litestream", cmd}, rest...), " ")
+	if configPath != "" {
+		// main() prints the hint after "Try: ", so align the alternative under it.
+		hint += "\n or: " + strings.Join(append([]string{"LITESTREAM_CONFIG=" + configPath, "litestream", cmd}, envRest...), " ")
+	}
+	return &usageError{message: "flags must come after the subcommand", hint: hint}
 }
 
 // Usage prints the help screen to STDOUT.
--- a/cmd/litestream/main.go
+++ b/cmd/litestream/main.go
@@ -14,7 +14,6 @@ import (
 	"path"
 	"path/filepath"
 	"runtime/debug"
-	"slices"
 	"strconv"
 	"strings"
 	"time"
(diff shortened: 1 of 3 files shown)


corylanou (changes requested):
Thanks for addressing the bare usage output here. I agree with the behavior goal, but I don't think we should reconstruct the invocation by scanning for a hard-coded command name.

The helper is not the parser, so it cannot reliably distinguish a command from a flag value. For example:

```console
$ litestream -socket status info
Error: flags must come after the subcommand
Try: litestream status -socket info
```

Here, `status` can be the value of `-socket` and `info` the intended command, but the suggestion reverses that interpretation. There are related correctness issues:

- A config path containing spaces is joined without shell quoting, so the result is not pasteable.
- `litestream -config c.yml version` suggests `litestream version -config c.yml`, although `version` does not accept `-config`.
- The command names already exist in the dispatcher and usage text. This adds another list that must remain synchronized; #1380 is currently adding the `mcp` command.

Could we keep this PR focused on producing a targeted, correct diagnostic without trying to infer and rewrite the full command? Something like:

```go
} else if strings.HasPrefix(cmd, "-") {
	return misplacedFlagError(cmd)
}
```

```go
func misplacedFlagError(arg string) error {
	name := strings.TrimLeft(strings.SplitN(arg, "=", 2)[0], "-")
	if name == "config" {
		return &usageError{
			message: "flags must come after the subcommand",
			hint: "litestream <command> -config PATH\n" +
				" or: LITESTREAM_CONFIG=PATH litestream <command>",
		}
	}
	return &usageError{
		message: "flags must come after the subcommand",
		hint:    "litestream <command> [flags]",
	}
}
```

That removes `commandNames`, `slices`, and the token scanner while still fixing #1405 and pointing alias users to `LITESTREAM_CONFIG`. It also cannot emit a confidently incorrect command.

The CLI tests can cover `-config PATH`, `--config=PATH`, another leading flag, explicit help, no arguments, and a correctly positioned flag. I would avoid asserting reconstructed user arguments.

If we want `litestream -config ./c.yml databases` to actually work, I would prefer a separate change that makes `-config` a real root flag using a top-level `flag.FlagSet`. Go's parser stops at the first non-flag argument, so it can parse root flags, select the subcommand, and leave the remaining arguments for the subcommand parser: https://pkg.go.dev/flag#hdr-Command_line_flag_syntax

ekanshul:
Rewritten to your design in a3e8342. `commandNames`, the `slices` import and the token scanner are gone, and the helper is your snippet as written:

```
$ litestream -config /etc/litestream.yml databases
Error: flags must come after the subcommand
Try: litestream <command> -config PATH
 or: LITESTREAM_CONFIG=PATH litestream <command>

$ litestream -socket status info
Error: flags must come after the subcommand
Try: litestream <command> [flags]
```

The second one is your example, and it is now a test case. Since the hint no longer names a command, the ambiguity you pointed out cannot be resolved wrongly, and the quoting and `version` problems go away with it.

`TestMainFlagPlacement` covers your six cases: `-config PATH`, `--config=PATH`, another leading flag, explicit help, no arguments, and a correctly positioned flag. No assertion mentions reconstructed arguments. The correctly positioned case only asserts the misplaced-flag message is absent, so it does not pin another command's error text. I also folded `-help` back in alongside `-h` and `--help`, since each of those starts with `-` and has to keep reaching the help branch rather than the new one.

I have updated the PR description, which still described the scanner.

One input this PR changes that we have not discussed, in case you want it handled differently:

```
# main
$ litestream -version
litestream is a tool for replicating SQLite databases.
... (usage, which lists the version command)
$ echo $?
0

# this PR
$ litestream -version
Error: flags must come after the subcommand
Try: litestream <command> [flags]
$ echo $?
1
```

`-version` is not a misplaced flag, it is a subcommand spelled as one, so there is no `litestream <command> -version` form to try. The usage screen was arguably the better answer there. This has been in the PR since the first revision rather than something the rewrite introduced, and it applies to `-v` too. Happy to special-case those two tokens to the usage screen, or to leave it as is if you would rather not grow the helper.

Not addressed here, per your review: making `-config` a real root flag with a top-level `flag.FlagSet`.

`go build ./...`, `go vet ./...`, `go test -race ./...` and `uvx pre-commit run --all-files` are clean.

PR #1454, opened 2026-08-19
config: warn about unrecognized YAML keys

### LINKED ISSUE #1403: config: unknown YAML keys are silently ignored, hiding misconfiguration
## Summary

`ParseConfig()` uses plain `yaml.Unmarshal`, so any key Litestream does not recognize is silently discarded. A user can write a configuration that looks correct, loads with exit 0, and does nothing.

This is not hypothetical — it is the root cause of a live documentation defect. The official v0.3 → v0.5 migration guide's block labeled "NEW FORMAT (recommended)" sets `retention: 72h` on a replica. That field does not exist. Everyone who followed that guide believes they have a 72h backup window and is actually running the 24h default.

## Reproduction

```
$ cat c.yml
definitely-not-a-setting: true
dbs:
  - path: /tmp/x.db
    replica:
      path: /tmp/xrep

$ litestream databases
path       replica
/tmp/x.db  file
$ echo $?
0
```

Real-world instances found so far, all silently ignored:

- `replica.retention` — `ReplicaSettings` has no such field (documented in the migration guide)
- `levels[].retention` — `CompactionLevelConfig` has only `Interval` (documented in the migration guide)

Historical precedent: #790 was the same failure class for age encryption — accepted configuration that did nothing.

Note that a *known* key with the wrong type does fail loudly (`exec:` given a list produces `cannot unmarshal !!seq into string`). It is only unknown keys that vanish.

## Suggested fix: warn, do not fail

**This must not become a hard error.** Litestream runs unattended in production. If an upgrade started rejecting configs that have loaded fine for months, every deployment carrying a stray or misspelled key would fail to start on restart — turning a silent misconfiguration into an outage, and taking replication down for people who were not having a problem yet.

The goal is visibility, not enforcement. On startup, log a warning naming each unrecognized key and continue:

```
level=WARN msg="unrecognized config key, ignoring" key=dbs[0].replica.retention
level=WARN msg="unrecognized config key, ignoring" key=levels[0].retention
```

That is enough to make the migration-guide case self-diagnosing: the user sees `retention` is being dropped instead of assuming it took effect.

I don't think this should ever be promoted to a hard failure. There is no version boundary where silently breaking existing deployments is worth it, and a warning already solves the actual problem, which is that the misconfiguration is invisible. If a stricter mode is ever wanted it should be opt-in, e.g. a `--strict-config` flag or a validate-only subcommand people can run in CI, never the default path.

## Implementation notes

1. This repo depends on `gopkg.in/yaml.v2`, not v3. The relevant APIs are `yaml.UnmarshalStrict()` or `Decoder.SetStrict(true)`. `KnownFields(true)` is a yaml.v3 API and will not apply here.

   Note both v2 strict APIs return an *error* rather than a warning list, so the likely shape is to run the strict decode separately, collect the resulting field errors, log them as warnings, and keep using the lenient decode's result as the live config.

2. `ParseConfig()` performs **two** unmarshal passes. The first intentionally targets a struct containing only `snapshot`, in order to detect whether the global snapshot fields were explicitly set. Making that preliminary pass strict would reject every other valid top-level key. Strictness must be applied only to the full `Config` decode.

Found while investigating #1400.

ekanshul:
## Summary
Log a WARN for every YAML key in the config file that the decoder cannot map to a known setting, and for every duplicate mapping key. The config still loads; these are warnings, not errors.

## Problem
`ParseConfig()` decodes with plain `yaml.Unmarshal` (gopkg.in/yaml.v2), so unknown keys are silently dropped and the user believes they took effect. On origin/main (63225f1):

```
$ cat c.yml
definitely-not-a-setting: true
dbs:
  - path: /tmp/x.db
    replica:
      path: /tmp/xrep
      retention: 72h
$ litestream databases -config c.yml
path       replica
/tmp/x.db  file
$ echo $?
0
```

Real cases from #1403: `replica.retention` (no such field, but the v0.3 -> v0.5 migration guide shows it) and `levels[].retention` (`CompactionLevelConfig` only has `interval`).

## Solution
After the existing lenient decode and validation succeed, and after `InitLog` has configured the logger, `ParseConfig` runs `yaml.UnmarshalStrict` a second time into a scratch `Config` used only for diagnostics. The live configuration is still the lenient decode, so behavior is unchanged for every config that loads today. Each `yaml.TypeError` detail is logged:

```
$ litestream databases -config c.yml
time=... level=WARN msg="configuration contains unrecognized or duplicate key" detail="line 1: field definitely-not-a-setting not found in type main.Config"
time=... level=WARN msg="configuration contains unrecognized or duplicate key" detail="line 6: field retention not found in type main.ReplicaConfig"
path       replica
/tmp/x.db  file
$ echo $?
0
```

This relies on the YAML library as the source of truth for tags, inline structs, custom unmarshalling and duplicate keys, rather than a second implementation of field resolution that has to stay synchronized with it. An earlier revision of this PR walked the decoded document against the `Config` field tags to produce a dotted path (`dbs[0].replica.retention`); the production code for that was 111 lines against 11 here, and the tradeoff is that the detail names the line, field and destination type instead of the path.

The preliminary snapshot-only decode pass is untouched. The returned `Config` and all error paths are unchanged. Warnings go through the configured logger, so they follow `logging.type`, `logging.stderr` and `LOG_LEVEL` like every other startup message.

## Behavior Change

| Scenario | Before | After |
|----------|--------|-------|
| Unknown top-level key | silently ignored, exit 0 | WARN with the detail, exit 0 |
| `replica.retention` / `levels[].retention` | silently ignored | WARN naming the line, field and type |
| Duplicate mapping key | last value silently wins | WARN, last value still wins |
| Valid config | no output | no output (unchanged) |
| Known key with wrong type | hard error | hard error (unchanged) |

## Scope
**In scope:**
- Strict second decode in `ParseConfig`, warning per `yaml.TypeError` detail
- Tests for top-level, `dbs[].replica`, `levels[]`, nested/multiple, a duplicate key, and a valid config

**Not in scope:**
- Making unknown keys a hard error, or a `--strict` / `--strict-config` flag (per the issue this should stay a warning; an opt-in strict mode could be a follow-up)
- Fixing the migration guide text that shows `replica.retention` (website repo)

## Test Plan
```bash
go test -race -v -run 'TestParseConfig_StrictWarnings' ./cmd/litestream/...
go test -race -run 'TestParseConfig|TestReadConfigFile|TestConfig_' ./cmd/litestream/...
go test -race ./cmd/litestream/...
go vet ./cmd/litestream/...
```
All pass locally (macOS, Go 1.26; gofmt, goimports and staticcheck clean).

## Related
- Fixes #1403
- Related to #1400

This change was made with AI assistance (Claude Code) and reviewed by the author.

### DIFF
--- a/cmd/litestream/main.go
+++ b/cmd/litestream/main.go
@@ -7,14 +7,12 @@ import (
 	"fmt"
 	"io"
 	"log/slog"
-	"maps"
 	"math"
 	"net/url"
 	"os"
 	"os/user"
 	"path"
 	"path/filepath"
-	"reflect"
 	"runtime/debug"
 	"strconv"
 	"strings"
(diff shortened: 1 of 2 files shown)


corylanou (changes requested):
Thanks for explaining the `UnmarshalStrict` alternative in the PR description. I tested that approach locally, and I think we should use it instead of maintaining the reflection-based field walker.

We can preserve the existing lenient decode as the live configuration and run `yaml.UnmarshalStrict` into a scratch value only for diagnostics:

```go
var strictConfig Config
if err := yaml.UnmarshalStrict(buf, &strictConfig); err != nil {
    var typeErr *yaml.TypeError
    if !errors.As(err, &typeErr) {
        return config, fmt.Errorf("strictly decode config: %w", err)
    }
    for _, detail := range typeErr.Errors {
        slog.Warn(
            "configuration contains unrecognized or duplicate key",
            "detail", detail,
        )
    }
}
```

This keeps the backward-compatible warning behavior while relying on the YAML library as the source of truth for tags, inline structs, custom unmarshalling, and duplicate keys. It also removes the need to keep a second implementation of YAML field resolution synchronized with the library and our configuration types.

The tradeoff is that the warning contains the YAML line, field name, and destination type instead of a full path:

```text
detail="line 5: field retention not found in type main.ReplicaConfig"
```

I think that is sufficient to make the misconfiguration self-diagnosing, and the substantially smaller maintenance surface is worth the less polished path. The [`yaml.v2` documentation](https://pkg.go.dev/gopkg.in/yaml.v2#UnmarshalStrict) also confirms that strict decoding handles both unknown fields and duplicate mapping keys.

I verified this implementation locally with:

```text
go test -race ./...
go vet ./...
pre-commit run --files cmd/litestream/main.go cmd/litestream/main_test.go
```

The production change drops from 111 lines to 11 lines. Tests cover an unknown key warning while continuing, a duplicate key warning while continuing, and valid configuration producing no warning.

With this reduced implementation, I’m comfortable leaving the code in `main.go`, where the rest of Litestream’s configuration parsing currently lives.

ekanshul:
Switched to `yaml.UnmarshalStrict` in 1919581, as you wrote it. The walker, and the `reflect` and `maps` imports, are gone; the production change is 11 lines of code plus a comment, down from 111.

```
$ litestream databases -config c.yml
time=... level=WARN msg="configuration contains unrecognized or duplicate key" detail="line 1: field definitely-not-a-setting not found in type main.Config"
time=... level=WARN msg="configuration contains unrecognized or duplicate key" detail="line 6: field retention not found in type main.ReplicaConfig"
path       replica
/tmp/x.db  file
```

I agree the shorter path is worth the less polished detail string, and you are right that strict decoding picks up duplicate keys for free.

On the tests: I kept the cases the PR already had rather than reducing to the three from your spike, since `levels[].retention` is one of the two cases named in #1403 and it seemed worth keeping a regression test for it. `TestParseConfig_StrictWarnings` now covers a top-level key, `dbs[].replica`, `levels[]`, a nested config producing several warnings at once, a duplicate key, and a valid config. The valid case is deliberately the rich one, since the risk in switching to strict decoding is warning about keys that are actually fine, and it is the shapes the library has to resolve itself that would break first: the inline `ReplicaSettings`, the custom `ByteSize` unmarshaler on `part-size`, and the anonymous `age` struct. All clean.

Three things I ran into that are worth knowing, none of which change my view that this is the right trade:

1. For a key inside the anonymous `age` struct the detail is hard to read, because the type has no name:

   ```
   detail="field bogus not found in type struct { Identities []string \"yaml:\\\"identities\\\"\"; Recipients []string \"yaml:\\\"recipients\\\"\" }"
   ```

   Giving that struct a named type would fix it, but that is a change to the config types and outside this PR.

2. The line number is the line in the env-expanded document, so with `$VAR` expansion on (the default) a multi-line value shifts it. The user still gets the field name and type, which is usually enough to find it.

3. The warnings only appear if `config.Validate()` passes, since that returns first. So a typo that breaks validation, like `pth` instead of `path`, still reports only the validation error and never says which key caused it. That placement is the same as before this PR, so I have left it alone, but it is the case where the diagnostic would help most. Happy to hoist `InitLog` above `Validate()` and move the strict decode up with it if you want that fixed here rather than separately.

`go build ./...`, `go vet ./...`, `go test -race ./...`, `gofmt`, `goimports` and `staticcheck` are all clean. PR description updated.

Issue #1460, opened 2026-08-21
vfs gets wedged on `non-contiguous ltx file` when an L1 compaction straddles `maxTXID1`


darkgnotic:
## Bug Description

A read-only Litestream VFS follower (live-tailing an S3 replica) can wedge permanently, e.g.:

```
cannot fetch new ltx files  error="poll L1: non-contiguous ltx file: level=1, current=0000000000000006, next=000000000000000c-000000000000001d"
```

Once in this state, `monitorReplicaClient` logs the error and retries every poll interval forever — the follower never advances past the stuck TXID and never self-heals. It keeps serving stale reads at the frozen position (queries succeed with stale data; nothing surfaces the failure to callers).

The backup data in object storage is **fully intact and contiguous** — this is not corruption and not a listing/consistency artifact. 

```
0000000000000001-000000000000000b.ltx    <- covers 1..b (INCLUDES 7,8,9,a,b)
000000000000000c-000000000000001d.ltx    <- covers c..1d
000000000000001e-000000000000002b.ltx
... (contiguous onward)
```

It is a state bug in the VFS tailer: the L1 high-water mark (`maxTXID1`) can be seeded to a value that is **not an L1 file boundary**, and a subsequent L1 compaction can emit a file whose TXID range *straddles* that watermark. The straddling file is then invisible to the level-1 seek forever, and the next L1 file trips the strict contiguity check.

## Environment

**Litestream version:**

```text
v0.5.16
```

**Operating system & version:**

alpine arm64

**Installation method:**

custom binary linking in vfs.go@v0.5.16, built from source

**Storage backend:**

s3

## Steps to Reproduce

1. Start a writer replicating to an S3 path; let it write L0 files up to TXID N **before** the first L0→L1 compaction.
2. Open a read-only VFS follower and let it build its index at that moment (L1 empty) → `maxTXID1` seeds to N.
3. Let the writer run its first compaction so it emits an L1 file with `MinTXID <= N <= MaxTXID` (a range that straddles N), then a second L1 file starting at `MaxTXID+1`.
4. The follower wedges on `non-contiguous ltx file` at L1 and never recovers.

**Expected behavior:**

The poll correctly handles the new L1 files

**Actual behavior:**

The L1 level in object storage was fully contiguous — no missing file:

```
0000000000000001-000000000000000b.ltx    <- covers 1..b (INCLUDES 7,8,9,a,b)
000000000000000c-000000000000001d.ltx    <- covers c..1d
[...]

PR #1464, opened 2026-08-21
fix(vfs): tolerate L1 files that straddle the poll watermark

darkgnotic:
## Description

The VFS live-tailer's incremental poll (`pollLevel`) and every `ReplicaClient`
backend's `LTXFiles(seek)` implementation assumed a resumed watermark always
falls exactly on a compaction-level file boundary (`file.MinTXID == watermark+1`).
That assumption breaks whenever a file's range *straddles* the watermark —
`MinTXID <= watermark < MaxTXID` — which happens:

* on open, `maxTXID1` is seeded from `pos.TXID` (an L0/snapshot position) when
   L1 is still empty. That seed is correct and necessary (L1 files get pruned by
   retention, so L1 doesn't generally start at TXID 1) — but it need not land on
   a future L1 file's boundary.

In both cases, every backend's seek filter (`MinTXID < seek` -> skip) makes the
straddling file invisible, and `pollLevel`'s exact-match contiguity check
(`MinTXID == watermark+1`) then rejects the *next* file as non-contiguous. At L1
and above this returns a hard error; nothing retries or recovers, so a follower
wedges permanently at the stale watermark — polling forever, logging
"non-contiguous ltx file", and serving stale reads with no indication to
callers.

This PR makes the seek + contiguity logic straddle-tolerant:

- `pollLevel` (`vfs.go`) now classifies each candidate file against the current
  watermark instead of requiring an exact match:
  1. Already covered (`MaxTXID <= watermark`) -> skip (defends against within-level 
   overlap; not produced by litestream today)).
  2. Covers the next TXID, including straddling (`MinTXID <= watermark+1 <=
     MaxTXID`) -> apply and advance the watermark to `MaxTXID`. Re-applying the
     already-covered portion of the page index is an idempotent overwrite, so
     this is safe.
  3. Real gap (`MinTXID > watermark+1`) -> unchanged: defer at L0, error above.
- Every backend's `LTXFiles` seek filter changes from `MinTXID < seek` to
  `MaxTXID < seek`, so a straddling file is returned instead of silently
  dropped: `s3`, `file`, `sftp`, `webdav`, `nats`, `oss`.
- `gs` and `abs` additionally used `seek` as a *list-key prefix*
  (`prefix += seek.String()`), which excludes a straddling file at the listing
  level regardless of the post-filter. Both now list the full level directory
  and filter by `MaxTXID < seek` in the iterator, matching `s3`/`oss`. This
  trades a narrower list for correctness — same list volume `s3`/`oss` already
  have.
- `mock` needs no change; it delegates entirely to a caller-supplied func.

The LTXFiles seek-semantics change is a no-op for every existing caller except pollLevel: seek==0 callers are unaffected by definition, L0 callers only ever see single-TXID files (MinTXID==MaxTXID), and compaction always seeks on a level-aligned boundary. Only pollLevel passes a seek that can fall inside a file's range.

## Motivation and Context

Found via a production incident: a read-only VFS follower tailing S3 wedged
permanently on 

`poll L1: non-contiguous ltx file: level=1, current=...6, next=...c-...1d`

even though the backup data in object storage was fully intact and contiguous. 
Root-caused to the seed-vs-boundary mismatch above (case 1) — the L1 file that
would have bridged the gap (`1-b`, covering the missing range) was present in
storage the whole time but permanently invisible to the seek. 

Fixes Issue #1460 

## How Has This Been Tested?

- Added `TestVFSFile_StraddlingL1FileAfterSeed`, which reproduces the exact
  production sequence (snapshot seeds `maxTXID1=6`, first L1 compaction emits a
  straddling `1-b`, next L1 file is `c-1d`) — fails on the old code
  (`maxTXID1` stuck at 6) and passes with this fix.
- Added `TestVFSFile_MergedWiderL1FileAfterConsume`, a defensive test that 
pollLevel doesn't wedge on within-level overlap; not a state litestream produces today.
- Updated `mockReplicaClient.LTXFiles` in tests to mirror the fixed backend
  seek semantics (`MaxTXID >= seek`).
- Full existing `-tags vfs` suite passes with no regressions:
  `go test -tags vfs ./...`.
- `go build -tags vfs ./...`, `go vet` clean on all backend packages.
- Per-backend package tests (`s3`, `gs`, `oss`, `nats`, `webdav`, `file`) pass;
  `abs`/`sftp`/`mock` have no existing test files to run.
- `gofmt -l` clean on all touched files.

Note: this also fixes a latent, unrelated compile break in the `-tags vfs` test
suite — `ReplicaClient` gained `SetLogger` in a prior commit but the VFS test
doubles (`mockReplicaClient`, `countingReplicaClient`, `writeTestReplicaClient`)
were never updated, so `go test -tags vfs ./...` hasn't built since. Added
no-op `SetLogger` implementations to unblock running (and adding to) these
tests; worth wiring `-tags vfs` into CI separately so this doesn't recur.

## Types of changes

## Checklist

### DIFF
--- a/abs/replica_client.go
+++ b/abs/replica_client.go
@@ -380,12 +380,13 @@ func newLTXFileIterator(ctx context.Context, client *ReplicaClient, level int, s
 		seek:   seek,
 	}
 
-	// Create paginator for listing blobs with level prefix
+	// Create paginator for listing blobs with level prefix. List the whole
+	// level and filter by seek in the iterator (like the s3 backend). We can't
+	// narrow the list with a seek-derived key prefix: a file whose range
+	// straddles the seek has a MinTXID below it, so a prefix on the seek value
+	// would exclude the very file that covers the next needed TXID.
 	dir := litestream.LTXLevelDir(client.Path, level)
 	prefix := dir + "/"
-	if seek != 0 {
-		prefix += seek.String()
-	}
 
 	itr.pager = client.client.NewListBlobsFlatPager(client.Bucket, &azblob.ListBlobsFlatOptions{
 		Prefix:  &prefix,
@@ -457,8 +458,10 @@ func (itr *ltxFileIterator) loadNextPage() bool {
 			Size:    *item.Properties.ContentLength,
 		}
 
-		// Skip if below seek TXID
-		if info.MinTXID < itr.seek {
+		// Skip only if the file's whole range is below the seek TXID. A file
+		// whose range straddles the seek (MinTXID < seek <= MaxTXID) still
+		// covers the next TXID a follower needs, so it must be returned.
+		if info.MaxTXID < itr.seek {
 			continue
 		}
 
--- a/file/replica_client.go
+++ b/file/replica_client.go
@@ -114,7 +114,8 @@ func (c *ReplicaClient) LTXFiles(ctx context.Context, level int, seek ltx.TXID, 
 		minTXID, maxTXID, err := ltx.ParseFilename(fi.Name())
 		if err != nil {
 			continue
-		} else if minTXID < seek {
+		} else if maxTXID < seek {
+			// Skip if the whole range is below seek.
 			continue
 		}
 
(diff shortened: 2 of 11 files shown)


darkgnotic:
@corylanou, I am somewhat uneasy about changing the semantics of the `LTXFiles()` iterator, though it makes sense for the `pollLevel()` logic. Claude seems to think that it is safe for all other call sites, but let me know if we need to parameterize the iterator for different behaviors.

PR #1465, opened 2026-08-22
fix(vfs): stop wedging the level-1 tailer on a straddling first compaction

### LINKED ISSUE #1460: vfs gets wedged on `non-contiguous ltx file` when an L1 compaction straddles `maxTXID1`

yzxcj797:
Fixes #1460.

## Investigation (reproduced from the report's timeline)

`rebuildIndex` seeds the level-1 tailer position from an L0-derived value when the index is built while L1 is still empty:

```go
maxTXID1 := maxLevelTXID(infos, 1)
if maxTXID1 == 0 {
    maxTXID1 = pos.TXID   // e.g. 6 — NOT an L1 file boundary
}
```

The writer's first L0→L1 compaction then emits a file whose range **straddles** that value (`1..0xb` around position 6). The level-1 seek filters by `MinTXID`, so the straddling file is invisible forever; when the next L1 file (`0xc..0x1d`) appears, the strict contiguity check fails on every poll and the follower freezes at TXID 6 while storage remains fully contiguous:

```
poll L1: non-contiguous ltx file: level=1, current=0000000000000006, next=000000000000000c-000000000000001d
```

## Fix

`maxTXID1` stays **zero** when there are no L1 files, so the first L1 file — which starts at TXID 1 regardless of how far L0 has advanced — is consumed contiguously, and the position tracks real L1 boundaries from then on.

## Scope

- This PR changes only the seeding in `rebuildIndex` (plus tests). It does **not** change the `MinTXID`-based seek semantics of any storage backend, nor the `pollLevel` contiguity rules — with a correctly-zero seed neither can encounter a straddle.
- The vfs-tagged test suite did not compile on main (the in-file test mocks predate the `SetLogger` addition to `ReplicaClient`); the three mocks get a no-op `SetLogger` so the tagged suite builds again. Pre-existing failures unrelated to this change (e.g. `TestReplica_SyncOnceLimitsLTXFiles`, `TestDB_SyncChunksWALAtCommitBoundary`) reproduce identically with and without this fix.

## Tests

Updated `TestVFSFile_OpenSeedsLevel1PositionFromPos` (which pinned the old seeding) to `TestVFSFile_OpenKeepsLevel1PositionAtZeroWithoutL1Files`, and added `TestVFSFile_PollConsumesStraddlingL1Compaction`, which replays the report's exact timeline: open with L0 at 6 and L1 empty, then present L1 `1..0xb` and `0xc..0x1d`, then poll.

```
go test -tags vfs -run 'TestVFSFile_PollConsumesStraddlingL1Compaction|TestVFSFile_OpenKeepsLevel1PositionAtZero' -count=1 .
# ok
```

Differential: reverting only the seeding (tests kept) fails with exactly the report's wedge error — `poll L1: non-contiguous ltx file: level=1, current=0000000000000006, next=000000000000000c-000000000000001d`.

### DIFF
--- a/vfs.go
+++ b/vfs.go
@@ -1209,10 +1209,13 @@ func (f *VFSFile) rebuildIndex(ctx context.Context, infos []*ltx.FileInfo, targe
 	}
 
 	maxTXID1 := maxLevelTXID(infos, 1)
-	// Seed maxTXID1 from pos when there are no L1 files
-	if maxTXID1 == 0 {
-		maxTXID1 = pos.TXID
-	}
+	// When there are no L1 files yet, maxTXID1 must stay zero: the first
+	// L1 compaction emits a file starting at TXID 1 whose range can
+	// straddle the current L0 position. Seeding from the L0-derived
+	// pos.TXID skipped that file in the level-1 seek forever and wedged
+	// the follower on "non-contiguous ltx file" once the next L1 file
+	// appeared (#1460). Starting from zero consumes the first L1 file
+	// (and any later ones) contiguously instead.
 
 	f.mu.Lock()
 	defer f.mu.Unlock()
--- a/vfs_test.go
+++ b/vfs_test.go
@@ -360,7 +360,7 @@ func TestVFSFile_OpenSeedsLevel1Position(t *testing.T) {
 	}
 }
 
-func TestVFSFile_OpenSeedsLevel1PositionFromPos(t *testing.T) {
+func TestVFSFile_OpenKeepsLevel1PositionAtZeroWithoutL1Files(t *testing.T) {
 	client := newMockReplicaClient()
 	snapshot := buildLTXFixture(t, 1, 's')
 	snapshot.info.Level = SnapshotLevel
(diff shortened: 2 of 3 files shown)


darkgnotic:
Thank you @yzxcj797 .  🙏 

FWIW, I asked Claude about this solution and it seemed to think that it would break compaction in write mode (e.g. introduced in #953). But TBH, I would love for this to be the correct solution, as it's much simpler.  🤞

yzxcj797:
@darkgnotic thanks for flagging that — I checked the write-mode path against #953 and I don't see an interaction:

- `maxTXID1` is consumed in exactly one place: `poll()` → `pollLevel(ctx, 1, f.maxTXID1, ...)` (vfs.go:2511/2538), the read-side L1 tailer. The only other reference is the `MaxTXID1()` accessor, which has no callers outside tests.
- The write-mode compactor (`compactor.go`) never reads `f.maxTXID1`; its level/retention decisions come from the replica client listing and its own state, so a zero seed can't steer compaction.
- In write mode the poll goroutine still runs ("in case another writer creates the database"). With the zero seed it consumes the first L1 compaction output (one-time, bounded by the first compaction's files) instead of skipping past it; from then on it tracks real L1 boundaries exactly as before. Page-index fetches that hit a file being compacted away fail the poll and retry next tick — the same failure/retry mode the pre-existing fresh-consumption path already has.

So the effect is strictly within the read-side tailer that the issue is about.

PR #1471, opened 2026-08-25
fix(config): reject snapshot retention shorter than interval

### LINKED ISSUE #1451: Nothing complains if I set retention interval lower than snapshot interval
Litestream starts up successfully with a config where `retention_period` is set shorter than `snapshot_interval`. 

This feels nonsensical and an obvious red flag. If snapshots are created less often than they're retained for, the retention setting has no meaningful effect. At minimum this should produce a startup warning; arguably it should be a hard validation error.

Not sure what logic there is to prevent the last remaining snapshot from being deleted, I assume there is one. Even so, this should be checked and the user should b warned.

UniversePeak:
## Summary
Reject configurations where snapshot retention is shorter than the snapshot interval.

## Problem
A configuration with `snapshot.interval` greater than `snapshot.retention` currently starts successfully even though retention cannot preserve a full snapshot interval. This makes the retention setting ineffective and provides no startup signal.

## Solution
Validate the relationship for both global and per-database snapshot settings, returning a dedicated configuration validation error.

## Scope
**In scope:**
- Global `snapshot.retention` versus `snapshot.interval` validation
- Per-database snapshot relationship validation
- Focused parser regression tests

**Not in scope:**
- Runtime retention behavior
- Changes to existing positive-duration validation

## Test Plan
- `go test ./cmd/litestream -run 'TestConfig' -count=1`
- `go vet ./cmd/litestream`

Both commands pass locally. The full package suite has unrelated Windows-path and file-locking failures in this checkout.

## Related
- Fixes #1451

### DIFF
--- a/cmd/litestream/main.go
+++ b/cmd/litestream/main.go
@@ -50,17 +50,18 @@ var errStop = errors.New("stop")
 
 // Sentinel errors for configuration validation
 var (
-	ErrInvalidSnapshotInterval         = errors.New("snapshot interval must be greater than 0")
-	ErrInvalidSnapshotRetention        = errors.New("snapshot retention must be greater than 0")
-	ErrInvalidCompactionInterval       = errors.New("compaction interval must be greater than 0")
-	ErrInvalidSyncInterval             = errors.New("sync interval must be greater than 0")
-	ErrInvalidL0Retention              = errors.New("l0 retention must not be negative")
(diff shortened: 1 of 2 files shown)


UniversePeak:
This PR has been open for over a week with all reported checks passing, including the focused config tests, lint, and cross-platform build checks. I’m following up on the review status when you have a chance. The change is limited to validating global and per-database snapshot retention against their intervals, with no runtime retention behavior changes.

Issue #1354, opened 2026-07-10
v0.5.14: dead TCP socket/fd accumulation on S3 replica drives host into kernel tcp_mem exhaustion (regression vs 0.5.12)


jaywedgeworth22:
### Summary

On v0.5.14 (linux amd64, official release tarball), a long-running `litestream replicate` process accumulates dead TCP socket file descriptors at a sustained ~1-2/s until the **kernel enters TCP memory pressure**, which clamps the receive window of every connection on the host and collapses all bulk TCP throughput machine-wide. Replication itself keeps working (INFO logs stay clean) — the damage is to the host.

### Environment

- litestream v0.5.14, linux x86_64 (release tarball binary)
- Running inside a Docker container (default bridge networking) as `litestream replicate -config <cfg> -exec "<app>"`
- One SQLite DB, ~180MB, WAL mode, steady write churn (~1 write tx/s, tens of KB per tx)
- Replica: `type: s3` against Cloudflare R2 (custom `endpoint`, `region: auto`), with weeks of accumulated multi-level LTX history
- DB-level `snapshot: {interval: 24h, retention: 720h}`

### Observed

- Socket fd count on the litestream PID grows ~1-2/s in a sawtooth (pool-eviction waves release batches, but the high-water mark keeps climbing). Measured peak: **16,840 open socket fds** after ~6h.
- The sockets are dead: `lsof` shows them as `sock ... "protocol: TCP"` (not identifiable/established); `/proc/net/sockstat` showed `TCP: inuse 13 ... alloc 16921 mem 183169` — i.e. ~17k allocated TCP socks not in any live table, pinning ~715MB (183k pages) of unread receive-queue memory.
- That drove the host to its `tcp_mem` max (182670 pages on an 8GB machine): dmesg logged `TCP: out of memory -- consider tuning tcp_mem`, and the kernel clamped `rcv_wnd` to ~6KB for *every* connection on the box (observed via `ss -ti`), so any high-RTT bulk transfer on the machine collapsed to ~20KB/s. (This is how we found it — unrelated services on the host started failing.)
- Process RSS also grows unboundedly (~1.7GB after ~6h).
- `strace -f -e socket,connect,close` over 12s of steady state: **271 socket() + 271 connect()** — a fresh DNS lookup + fresh TLS connection per S3 operation, no keep-alive reuse, consistent with heavy retry churn even though operations ultimately succeed.

### Not reproducible in a minimal setup

[...]

PR #1472, opened 2026-08-25
fix: resumable LTX reader aborts restore on transient S3 connection resets

darkgnotic:
## Description

The `ResumableReader` used to wrap LTX file streams during restore treated its retry budget as a **lifetime** cap rather than a **consecutive-failure** cap. `retryN` was incremented on every reconnect and never reset, so "3 retries" meant "3 reconnects total for the file's entire transfer" instead of "3 consecutive failures without progress."

This change resets `retryN` to 0 whenever a `Read` returns bytes, so the cap now counts consecutive failures at a stuck offset. A connection that periodically drops but keeps advancing recovers instead of aborting; a stream genuinely stalled at one offset still fails after 3 tries.

No lifetime cap is added: termination is guaranteed by the storage backend returning EOF at the true end of a finite object, backstopped by the 3-consecutive-no-progress limit. (A fixed lifetime cap would be actively harmful — a legitimately large file over a flaky link could exceed it while making perfect progress, reproducing this same bug at a higher threshold.)

The reconnect backoff is now a struct field defaulted from the package constant, purely as a test seam so reconnect-heavy cases can run without the real delay.

## Motivation and Context

Our process crash-looped indefinitely, never completing its litestream restore (disk/memory see-sawing, pod restarting every ~30–40 min). The underlying error:

```
Error: decode database: decode page 30201904: read page header 69: read compressed data:
  max retries exceeded reading ltx file (level=0, min=0000000000000046, max=0000000000000046,
  offset=25141595): read tcp ...:443: read: connection reset by peer
```

The connection reset at `offset 1210312` (attempt 2) and again at `offset 25141595` (attempt 4) on the same file — ~24 MB of successful progress between the two resets, yet the counter climbed to 4 and tripped the limit of 3. The network path is healthy (S3 egresses via the VPC S3 gateway endpoint); these are ordinary transient resets the reader is designed to recover from. Because each restore restarts from scratch, any file large enough to see a few resets means the restore never converges.

## How Has This Been Tested?

`go test ./...`, `go vet ./...`, and `gofmt` all clean. Focused coverage in `internal/resumable_reader_test.go`:

- **`ResetsRetriesOnForwardProgress`** — a stream that resets ~10× with a chunk of progress between each now completes; would have failed at the 4th reset under the old code (direct regression for the reported failure).
- **`CompletesAcrossPartialReads`** (reworked from `BoundsConnectionsAcrossPartialReads`) — a one-byte-per-connection backend now succeeds, with reconnects bounded by file size. The old assertion ("fail after `maxRetries+1` connections") was the bug encoded as expected behavior, so it was inverted.
- All existing subtests (normal read, reconnect-on-error, premature-EOF, backoff, context-cancel, max-retries-exceeded, etc.) pass unchanged.

## Types of changes

## Checklist

### DIFF
--- a/internal/resumable_reader.go
+++ b/internal/resumable_reader.go
@@ -45,9 +45,16 @@ type ResumableReader struct {
 	size    int64 // expected total file size from FileInfo; 0 means unknown
 	offset  int64
 	rc      io.ReadCloser
-	retryN  int
-	err     error
-	logger  *slog.Logger
+	// retryN counts consecutive failures since the last forward progress, so a
+	// stream that periodically drops but keeps advancing is not limited by it;
+	// only a stream stuck at one offset trips resumableReaderMaxRetries.
+	retryN int
+	err    error
+	logger *slog.Logger
+
+	// backoff is defaulted from resumableReaderBackoff by NewResumableReader and
+	// overridable in tests to keep reconnect-heavy cases fast.
+	backoff time.Duration
 }
 
 // NewResumableReader creates a ResumableReader. Primarily exposed for testing.
@@ -61,9 +68,14 @@ func NewResumableReader(ctx context.Context, client LTXFileOpener, level int, mi
 		size:    size,
 		rc:      rc,
 		logger:  logger,
+		backoff: resumableReaderBackoff,
 	}
 }
 
+// resumableReaderMaxRetries bounds *consecutive* failed reads at a single
+// offset — a stream making no forward progress. It resets every time a Read
+// returns bytes, so a connection that periodically drops but keeps advancing
+// (the common S3/Tigris case during a large restore) is not limited by it.
 const resumableReaderMaxRetries = 3
 
 // resumableReaderBackoff is the base delay between retry attempts, doubling
@@ -98,6 +110,12 @@ func (r *ResumableReader) Read(p []byte) (int, error) {
 
 		n, err := r.rc.Read(p)
 		r.offset += int64(n)
+		if n > 0 {
+			// Forward progress: the connection recovered, so clear the
+			// consecutive-failure budget. Only reads that advance zero bytes
+			// count against resumableReaderMaxRetries.
+			r.retryN = 0
+		}
 
 		if err == nil {
 			return n, nil
@@ -174,7 +192,7 @@ func (r *ResumableReader) retry(err error) error {
 	case <-r.ctx.Done():
 		r.err = r.ctx.Err()
 		return r.err
-	case <-time.After(resumableReaderBackoff << (r.retryN - 1)):
+	case <-time.After(r.backoff << (r.retryN - 1)):
 	}
 	return nil
 }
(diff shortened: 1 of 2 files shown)


corylanou:
Thanks for the detailed production evidence here. I reproduced the restore-liveness failure and have already implemented and verified a combined solution locally on top of #1493's current head (`00422ba`). Nothing has been pushed.

#1493 addresses a separate lifecycle problem in the same reader: `Close` must be terminal, concurrent reads must not reopen afterward, and a stream returned by an in-flight reopen must be closed immediately. I used that behavior as the reconciliation baseline.

One concern with this PR's current `n > 0` reset is that it reverses the protection added by #1360 for #1354. The #1360 HTTP regression declares the remaining `Content-Length` but returns one byte per connection. Go returns that byte as a short successful read before reporting the truncation on the next read. Resetting on any positive byte therefore lets the transfer open one connection per byte instead of returning a retry-limit error so replication backoff can engage.

The local combined implementation uses a narrower progress invariant: renew the retry budget only after a positive read completely satisfies the caller's request (`err == nil && n == len(p)`). Bytes delivered with an error or as a short read still advance the resume offset, but they do not clear the connection-churn protection.

This matches the production LTX path, which uses `io.ReadFull` for file headers, page headers, size prefixes, and payloads. A healthy resumed stream completes those requests; the one-byte truncated backend does not.

Local results on top of #1493:

- Five disconnects with 64 KiB of successful progress between each completed a 393,216-byte transfer using six total opens. The #1493/#1360 lifetime policy failed at the fourth disconnect and offset 262,144.
- The preserved one-byte/truncated-response test returned `max retries exceeded` with no more than four connections.
- Zero-progress exhaustion, premature EOF, reopen errors, offsets, provider backoff, and context cancellation passed.
- Close-during-reopen and the compactor no-reopen/no-leak regressions passed under `-race`, including 10 repeated runs.
- `go test -race -count=1 ./...`, `go vet ./...`, staticcheck, all pre-commit hooks, and the three-profile short `TestLTXBehavior` gate passed. The behavioral gate completed in 505 seconds with restoration and integrity checks green for every profile.

Would you be comfortable enabling maintainer edits on this PR and giving us permission to add the locally proven reconciliation as a normal follow-up commit? That would let the combined approach go through this PR's full review and CI process without rewriting your history.

If you prefer, you can also incorporate the patch yourself, or we can open a separate upstream reconciliation PR with clear attribution to both #1472 and #1493.

darkgnotic:
Thank you @corylanou!

It looks my rocicorp/litestream, under an organizational account, doesn't support the **Allow maintainer edits** feature.

Please, do feel free to open a separate upstream reconciliation PR and close this one!

corylanou:
Thank you again for the investigation and collaboration. The locally proven combined solution is now open as #1500, with attribution to this PR and #1493 and the full test evidence. Closing this PR as superseded by #1500, per your request.

PR #1479, opened 2026-08-27
fix(store): serialize snapshots and compactions per database

### LINKED ISSUE #1477: Concurrent L9 snapshot and L1 compaction for one (large) db can exhaust memory
## Bug Description

Litestream can run an L9 snapshot and an L1 compaction concurrently for the same database. On a large database, both operations retain database-sized LTX page indexes. In our case, the overlap exhausted a 12 GiB container and caused Litestream to be OOM-killed.

This differs from #331, which concerned concurrent snapshots across many databases. This case involves snapshot and compaction overlap for one database.

- Each compaction level and the snapshot level run in independent goroutines:
  https://github.com/benbjohnson/litestream/blob/main/store.go#L212-L233
- `CompactDB` does not serialize work per database:
  https://github.com/benbjohnson/litestream/blob/main/store.go#L769-L830

## Environment

**Litestream version:**

```text
v0.5.16
github.com/superfly/ltx v0.5.2
```

Linux container running in Kubernetes on AWS EKS.

**Installation method:**

Docker. Litestream is bundled in: https://github.com/rocicorp/mono/blob/72a8696182d6fa1ed4bcabcdcf9e36e529c61fb3/packages/zero/Dockerfile#L24 (we have a couple fixes on top of 0.5.16 unrelated to this issue).

**Storage backend:**

S3

## Steps to Reproduce

1. Create or replicate an approximately 130 GiB SQLite database using the default 4096-byte page size. This produces approximately 34 million pages.
2. Start Litestream against a fresh S3 replica with a 12 GiB memory limit, the configuration below, `GOMEMLIMIT=8GiB`, and multipart concurrency 4.
3. Allow the initial full L0 and L9 snapshot work to begin.
4. After the initial L0 completes, observe L1 compaction begin while the L9 snapshot remains active.
5. Observe container memory approach 12 GiB and the process terminate with `OOMKilled` / exit code 137.

**Expected behavior:**

Maintenance work for one database should have bounded concurrency and memory usage. A snapshot and compaction for the same database should not multiply database-sized memory requirements enough to OOM the process.

**Actual behavior:**

The L9 snapshot and L1 compaction overlapped. Container working memory increased from approximately 5.75 GiB to 11.63 GiB before the process was OOM-killed.

The process restarted against the existing local database, created a new backup generation, and began repeating the same initial backup cycle.

## Configuration

```yaml
snapshot:
  interval: 4h

levels:
  - interval: 2m # L1

dbs:
  - path: /data/replica.db
    replica:
      url: s3://REDACTED
      concurrency: 4
      part-size: 16777216
```

Relevant process and container settings:

```text
GOMEMLIMIT=8GiB
Container memory limit: 12GiB
SQLite database size: approximately 130.4GiB
SQLite page size: 4096 bytes
Multipart concurrency: 4
Multipart part size: 16MiB
```

</details>

## Logs

```text
2026-08-26T19:02:10Z L9 multipart upload initiated:
  0009/0000000000000001-0000000000000001.ltx

2026-08-26T19:05:05.208631806Z
level=INFO msg="ltx file uploaded" system=store db=replica.db
replica=s3 level=0 minTXID=0000000000000001
maxTXID=0000000000000001 size=63603458713

2026-08-26T19:06:01Z L1 multipart upload initiated while L9 remained active:
  0001/0000000000000001-0000000000000001.ltx

2026-08-26T19:06:05Z container working memory: 11603Mi
2026-08-26T19:07:10Z container working memory: 11913Mi
2026-08-26T19:08:22Z container working memory: 11749Mi

2026-08-26T19:08:44.439274146Z
level=INFO msg="forcing truncate checkpoint" system=store
db=replica.db wal_size=3545309472 threshold=499999112

2026-08-26T19:08:44.439315810Z
level=INFO msg="checkpoint skipped, snapshot in progress"
system=store db=replica.db mode=TRUNCATE

2026-08-26T19:08:55Z
Kubernetes termination reason: OOMKilled
Exit code: 137
```

There was no application-level OOM message because the process was terminated by the container memory cgroup.

</details>

## Additional Context

The database contains approximately:

```text
130.4 GiB / 4096 bytes = approximately 34.2 million pages
```

Code inspection suggests the dominant scaling comes from LTX page indexes.

The LTX encoder retains one map entry per output page:

- Map declaration:
  https://github.com/superfly/ltx/blob/v0.5.2/encoder.go#L17-L23
- Entry added for every page:
  https://github.com/superfly/ltx/blob/v0.5.2/encoder.go#L268-L277
- Map iterated and sorted during finalization:
  https://github.com/superfly/ltx/blob/v0.5.2/encoder.go#L120-L155

For approximately 34.2 million entries, one `map[uint32]PageIndexElem` is estimated to require roughly 2.9 GiB after Go map bucket overhead. Old and new bucket arrays can temporarily coexist while the map grows.

Compaction can also decode and retain input page indexes during finalization:

- `io.ReadAll` and decoder index retention:
  https://github.com/superfly/ltx/blob/v0.5.2/decoder.go#L77-L98
- Input index map construction:
  https://github.com/superfly/ltx/blob/v0.5.2/decoder.go#L369-L405

During snapshot and L1 overlap, the L9 output index, L1 output index, and L1 input index can coexist.

`GOMEMLIMIT=8GiB` and reducing multipart concurrency from the default to 4 did not prevent the OOM. We temporarily increased the container memory limit while investigating.

corylanou:
## Summary

`Store.Open` runs one goroutine per compaction level plus one for the snapshot level, and `CompactDB` did nothing to serialize work per database, so an L9 snapshot and an L1 compaction can run at the same time on the same database. Each holds a database-sized LTX page index, so peak memory multiplies with database size; a 130 GiB database was OOM-killed in a 12 GiB container this way (#1477).

- `Store.CompactDB` now takes a per-database `TryLock`. A refused operation returns the new `ErrMaintenanceBusy` instead of queueing, so one database's long snapshot cannot stall the level monitor for every other database.
- The level monitor treats busy like "no compaction" but wakes early to retry: 1s after the first refusal, doubling up to `MaintenanceBusyRetryInterval` (10s) while the database stays busy. An early pass only re-attempts the databases that were refused (plus any registered since the last regular pass) and skips snapshot retention; regular passes stay anchored to the level's schedule. Without the early retry a snapshot that lost startup contention to L1 would wait for the next snapshot boundary (up to 24h); without the 1s start, levels whose boundaries coincide (e.g. L1 15s / L2 30s in the LTX behavioral gate) would drift by a full 10s on every collision.
- The snapshot reader's `Close` now cancels and joins the producer goroutine, so a snapshot whose replica write fails early cannot leave its WAL page map and encoder alive after the lock is released.
- Compaction across different databases is unaffected (covered by the new test).

This bounds concurrent maintenance memory to one operation per database. The page indexes themselves still scale with database size; shrinking the encoder's index is https://github.com/superfly/ltx/pull/95, and this PR pins `go.mod` to that branch's head so it builds and soaks as the intended final combination. **The pin must be replaced with the tagged ltx release before merge.**

Fixes #1477

## Evidence

Measured with the litestream-soak `snapshot-compaction-overlap` rig (https://github.com/corylanou/litestream-soak/pull/196, https://github.com/corylanou/litestream-soak/pull/197): it runs snapshot and L1 compaction sequentially as a baseline, then holds the snapshot's replica stream at 95% so the encoder's page index stays resident while the L1 compaction runs, sampling `runtime.MemStats` (GOGC=25) and writing a heap profile at each phase's peak. "Overlap" is the peak heap growth during that phase; the pass limit is 1.10× the larger sequential phase plus one multipart upload's fixed buffers. The "this PR" rows were built by the rig straight from this PR's head (`go.mod` as committed, ltx pinned to superfly/ltx#95), not via a local replace.

| build | DB | seq snapshot | seq compact-l1 | overlap | limit | result |
|---|---|---|---|---|---|---|
| main 3b468c9 + ltx v0.5.2 | 1 GiB / 262,803 pages | 69.0 MB | 101.3 MB | **170.0 MB** | 145.0 MB | fail (overlap/sum = 1.00) |
| serialization only (ltx v0.5.2) | 1 GiB | 68.6 MB | 99.3 MB | **99.1 MB** | 142.8 MB | pass (ratio 1.00) |
| this PR (serialization + ltx#95) | 1 GiB | 48.7 MB | 79.2 MB | **81.3 MB** | 120.7 MB | pass (ratio 1.03) |
| main 3b468c9 + ltx v0.5.2 | 4 GiB / 1,051,216 pages | 168.6 MB | 306.3 MB | **470.6 MB** | 370.5 MB | fail (overlap/sum = 0.99) |
| serialization only (ltx v0.5.2) | 4 GiB | 173.1 MB | 300.7 MB | **286.5 MB** | 364.3 MB | pass (ratio 0.95) |
| this PR (serialization + ltx#95) | 4 GiB | 68.8 MB | 203.4 MB | **199.6 MB** | 257.3 MB | pass (ratio 0.98, `gate_release_reason=timeout`, 364 busy retries) |

On main the overlap's memory is the sum of the two operations; with serialization it is the larger of the two; with ltx#95 the encoder index itself shrinks from ~95 B/page to ~18 B/page. Peak heap profiles (`inuse_space`, 4 GiB run):

```
main, overlap phase:        233.81MB ltx.(*Encoder).EncodePage   (two live encoders)
                             64.00MB s3/manager.(*maxSlicePool).newSlice (two in-flight uploads)
                             48.03MB ltx.DecodePageIndex
serialization only:         114.54MB ltx.(*Encoder).EncodePage   (one encoder)
                             32.00MB s3/manager.(*maxSlicePool).newSlice
                             13.20MB ltx.DecodePageIndex
this PR, snapshot phase:     32.00MB s3/manager.(*maxSlicePool).newSlice
                             15.70MB ltx.(*pageIndex).append
this PR, overlap phase:      84.45MB ltx.DecodePageIndex        (now the largest term; follow-up in ltx)
                             32.00MB s3/manager.(*maxSlicePool).newSlice
```

Overlap cost: ~450 B/page on main, ~270 B/page with serialization alone, ~190 B/page for this PR as pinned — at the reporter's 34.2M pages that is ≈15 GB → ≈9.3 GB → ≈6.5 GB. The decoder-side `DecodePageIndex` map is now the biggest remaining per-page term (~80 B/page) and is called out as follow-up work in superfly/ltx#95.

The same head is also deployed as the `pr-1479` litestream-soak fleet (15 workers incl. two 100-database workers that upload hourly heap/alloc/goroutine/CPU pprof captures) for a longer-running comparison against `main`.

## Known limits

- `TryLock` is not fair; under continuous contention on one database a level could repeatedly lose to another. Defaults make that unlikely (snapshots are daily, compactions minutes apart).
- Only `Store.CompactDB` takes the lock. Direct `DB.Snapshot`/`DB.Compact` callers (e.g. `replicate -force-snapshot`, which runs with monitors disabled) are not serialized.
- The pre-existing race between the monitor's `IsOpen` check and `DisableDB`/`UnregisterDB` is unchanged; a re-registered path gets a fresh `DB` and therefore a fresh lock.

## Test plan

- [x] `tests/integration` LTX behavioral gate (`-short`) passes locally with the 1s→10s backoff (it failed on the first push, where a 10s retry after every L1/L2 boundary collision fell outside the ±50% timing window)
- [x] `TestStore_CompactDB_SerializesPerDBMaintenance`: same-DB compaction refused with `ErrMaintenanceBusy` while a snapshot is held, other DB unaffected, lock available after release
- [x] `TestBusyRetryDelay` for the monitor delay cap
- [x] `go test ./` root package, `go vet ./...`, pre-commit (goimports, vet, staticcheck)
- [x] Rig A/B above, 1 GiB and 4 GiB, including a build from this PR's exact head
- [x] Five adversarial Codex review passes; all confirmed findings addressed (producer join + cancellation, busy retry cap, early passes scoped to refused/not-ready/newly registered DBs, schedule anchored at the regular boundary, test cleanup ordering); final pass approved with no confirmed regressions

### DIFF
--- a/db.go
+++ b/db.go
@@ -65,19 +65,23 @@ const (
 // with indefinite write blocking (issue #724). All checkpoints now use either
 // PASSIVE (non-blocking) or TRUNCATE (emergency only) modes.
 type DB struct {
-	mu        sync.RWMutex
-	execSem   *semaphore.Weighted
-	path      string        // part to database
-	metaPath  string        // Path to the database metadata.
-	db        *sql.DB       // target database
-	f         *os.File      // long-running db file descriptor
-	rtx       *sql.Tx       // long running read transaction
-	pageSize  int           // page size, in bytes
-	notify    chan struct{} // closes on WAL change
-	chkMu     sync.RWMutex  // checkpoint lock
-	opened    bool          // true if Open() was called and Close() not yet called
-	syncState syncState
-	syncDiag  diagState
+	mu       sync.RWMutex
+	execSem  *semaphore.Weighted
+	path     string        // part to database
+	metaPath string        // Path to the database metadata.
+	db       *sql.DB       // target database
+	f        *os.File      // long-running db file descriptor
+	rtx      *sql.Tx       // long running read transaction
+	pageSize int           // page size, in bytes
+	notify   chan struct{} // closes on WAL change
+	chkMu    sync.RWMutex  // checkpoint lock
+	// maintenanceMu serializes snapshots & compactions per database; held via
+	// TryLock by Store.CompactDB so overlapping maintenance is refused with
+	// ErrMaintenanceBusy rather than queued (issue #1477).
+	maintenanceMu sync.Mutex
+	opened        bool // true if Open() was called and Close() not yet called
+	syncState     syncState
+	syncDiag      diagState
 
 	// last file info for each level
 	maxLTXFileInfos struct {
@@ -2693,12 +2697,23 @@ func (p *snapshotReadPosition) close() {
 
 type snapshotReadCloser struct {
 	*io.PipeReader
-	pos *snapshotReadPosition
+	pos    *snapshotReadPosition
+	cancel context.CancelFunc // stops the producing goroutine
+	done   <-chan struct{}    // closed when the producing goroutine has exited
 }
 
+// Close stops the producing goroutine and waits for it to exit, so that no
+// snapshot state (WAL page map, LTX encoder page index) outlives the reader.
+// Cancelling the producer's context interrupts the WAL scan that runs before
+// the first pipe write; closing the pipe interrupts any write in progress.
+// Store.CompactDB relies on this to keep a failed snapshot's memory from
+// overlapping the next maintenance operation.
 func (r *snapshotReadCloser) Close() error {
-	defer r.pos.close()
-	return r.PipeReader.Close()
+	r.cancel()
+	err := r.PipeReader.Close()
+	<-r.done
+	r.pos.close()
+	return err
 }
 
 // SnapshotReader returns the current position of the database & a reader that contains a full database snapshot.
@@ -2812,8 +2827,12 @@ func (db *DB) snapshotReader(ctx context.Context, pos *snapshotReadPosition) (io
 
 	// Execute encoding in a separate goroutine so the caller can initialize before reading.
 	pr, pw := io.Pipe()
+	ctx, cancel := context.WithCancel(ctx)
+	done := make(chan struct{})
 	go func() {
+		defer close(done)
 		defer pos.close()
+		defer cancel() // release the derived context even if the reader is never closed
 
 		walFile, err := os.Open(db.WALPath())
 		if err != nil {
@@ -2893,7 +2912,7 @@ func (db *DB) snapshotReader(ctx context.Context, pos *snapshotReadPosition) (io
 		_ = pw.Close()
 	}()
 
-	return &snapshotReadCloser{PipeReader: pr, pos: pos}, nil
+	return &snapshotReadCloser{PipeReader: pr, pos: pos, cancel: cancel, done: done}, nil
 }
 
 func snapshotHeaderWALRange(maxOffset, frameSize int64) (offset, size int64) {
--- a/go.mod
+++ b/go.mod
@@ -27,7 +27,7 @@ require (
 	github.com/prometheus/client_golang v1.17.0
 	github.com/psanford/sqlite3vfs v0.0.0-20260519004904-f9180fa2acc9 // direct
 	github.com/studio-b12/gowebdav v0.11.0
-	github.com/superfly/ltx v0.5.3-0.20260910220417-f299ac5b951c
+	github.com/superfly/ltx v0.5.3
 	golang.org/x/crypto v0.56.0
 	golang.org/x/sys v0.47.0
 	google.golang.org/api v0.155.0
--- a/go.mod
+++ b/go.mod
@@ -27,7 +27,7 @@ require (
 	github.com/prometheus/client_golang v1.17.0
 	github.com/psanford/sqlite3vfs v0.0.0-20260519004904-f9180fa2acc9 // direct
 	github.com/studio-b12/gowebdav v0.11.0
-	github.com/superfly/ltx v0.5.2
+	github.com/superfly/ltx v0.5.3-0.20260910220417-f299ac5b951c
 	golang.org/x/crypto v0.56.0
 	golang.org/x/sys v0.47.0
 	google.golang.org/api v0.155.0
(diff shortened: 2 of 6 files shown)


darkgnotic:
Thanks @corylanou. 

Just to update you on our thoughts here, with your pending change to spill the page index to disk (https://github.com/benbjohnson/litestream/pull/1483 - thank you!), along with the reduction of the size of most compactions (https://github.com/benbjohnson/litestream/pull/1514), we are beginning to wonder whether serializing compactions is necessary, or even desired.

For example, with the latter change to keep snapshots out of the compaction ladder, compactions will be increment sized, and it will be nice to be able to do a (potentially long) L9 snapshot without blocking the smaller, L0->L1 compactions. 

Perhaps what it comes down to is, if we are be able to effectively cap memory usage with your spill-to-disk improvement, perhaps we can continue letting compactions run in parallel.

Interested in your thoughts!

Issue #1467, opened 2026-08-24
Large L2 backlog does not complete despite resumable source reads


MischaU8:
## Bug Description

This may be a remaining large-backlog case related to #1308 and #1326.

L0, L1, and snapshots succeeded, but L2 remained empty. L2 source reads eventually failed with `max retries exceeded` despite making progress between transient disconnects. Each retry attempted the full accumulated backlog: approximately 2,680 L1 objects and 41.7 GiB.

## Environment

**Litestream version:**

The issue occurred with v0.5.16. I also tested a build from `main`.

```text
v0.5.16
```

**Operating system & version:**

Ubuntu 24.04.4 LTS, x86_64.

**Installation method:**

Standalone binary managed by systemd; follow-up build compiled from source.

**Storage backend:**

DigitalOcean Spaces through the S3-compatible backend.

## Steps to Reproduce

I do not have a provider-independent minimal reproduction.

1. Accumulate a large L1 backlog while L2 encounters transient source-read failures.
2. Allow the next hourly L2 compaction to process the pending sources.
3. Observe that it retries the full backlog and eventually exhausts the reader’s retry budget.

**Expected behavior:**

L2 eventually completes when transient failures are separated by substantial forward progress.

**Actual behavior:**

L2 remained empty while L1 continued accumulating.

## Configuration

<details><summary>litestream.yml</summary>

```yaml
snapshot:
  interval: 24h
  retention: 168h

levels:
  - interval: 5m
  - interval: 1h
  - interval: 24h

dbs:
  - path: /path/to/database.db
    replica:
      url: s3://<bucket>.<region>.digitaloceanspaces.com/<prefix>
      sync-interval: 60s
      part-size: 25M
```

</details>

## Logs

<details><summary>Log output</summary>

```text
level=ERROR msg="compaction failed" level=2
error="read multipart upload data failed:
max retries exceeded reading ltx file:
unexpected EOF"
```

The source objects remained readable, including range requests from the reported failure offsets.

</details>

## Additional Context

The database is approximately 26.8 GiB and produces about 4.5 GiB of L1 data per day:

- approximately 288 L1 objects/day;
- approximately 16 MiB per L1 object;
- normally about 12 inputs, or 200 MiB, per hourly L2 compaction.

The large operation resulted from more than nine days of failed L2 work.

[...]

PR #1493, opened 2026-08-31
fix(replica): stop resumable readers reopening streams after close

### LINKED ISSUE #1492: Compaction source streams reopened after Compact returns are never closed, leaking one connection per source file when an upload fails mid-compaction
## Bug Description

`Compactor.Compact` does not wait for its compaction goroutine before returning. When `WriteLTXFile` fails while that goroutine is still reading source files, `Compact` returns and closes every source stream it opened, but the goroutine keeps reading. Each subsequent read on a closed stream fails, `internal.ResumableReader` treats the failure as a transient connection error, reopens the stream from the current offset, and keeps going until it finally hits the closed pipe. Nothing ever closes those reopened streams.

`ResumableReader.Close` only closes the current underlying stream; it does not mark the reader as closed, so nothing stops the reopen. With an S3 backed replica every leaked stream is an HTTP response body: the connection is pinned out of the pool, the peer eventually times it out and closes its side, the socket sits in CLOSE_WAIT, and the file descriptor is held until the process dies.

The leak scales with the number of source files: during the header decode phase the compaction goroutine reads every input sequentially, so after `Compact` returns it can resurrect one stream per input before it exits. A wedged compaction (for example one corrupt L0 file that makes every attempt fail, while the backlog grows because L0 retention only prunes compacted files) turns this into unbounded fd growth.

To be clear about what does work: the error return path of `Compact` itself is fine. The `defer` added in #979 closes every stream `Compact` opened, and I verified with a close tracking client that a corrupt input file alone (compaction goroutine fails first) leaks nothing. The leak needs the other ordering: upload failure first, goroutine still reading. Long wedged compactions against remote storage make that ordering likely, because the upload connection sits idle for the whole multi minute header decode phase and providers time out idle connections.

## Environment

**Litestream version:**

Reproduced on v0.5.15, v0.5.16, and current `main` (25ac743).

```text
v0.5.15, v0.5.16, main @ 25ac743
```

**Operating system & version:** Linux (production), macOS 15 (repro test); not OS specific

**Installation method:** binary download (production), source (repro)

**Storage backend:** S3 compatible (Cloudflare R2 in production); the repro test uses the file backend with a wrapper client

## Steps to Reproduce

Drop the test below into the repo root as `compactor_leak_repro_test.go` (it reuses the existing `createTestLTXFile` and `countCompactorPipeWriters` helpers from `compactor_test.go`) and run:

```bash
go test -race -run TestCompactor_CompactDoesNotReopenSourcesAfterClose -v .
```

Current `main` fails with:

```text
compactor_test.go: 3 of 6 source streams still open after Compact: [level=0 txid=0000000000000001-0000000000000001 offset=0 level=0 txid=0000000000000002-0000000000000002 offset=0 level=0 txid=0000000000000003-0000000000000003 offset=0]
```

Three streams are opened by `Compact`, all three are closed by its defer, and then all three are reopened by the detached goroutine and leaked.

```go
// TestCompactor_CompactDoesNotReopenSourcesAfterClose verifies that no source
// stream is left open once Compact and its compaction goroutine have finished,
// even when the destination write fails while the goroutine is still reading.
func TestCompactor_CompactDoesNotReopenSourcesAfterClose(t *testing.T) {
	client := newLeakCheckCompactionClient(t.TempDir())
	compactor := litestream.NewCompactor(client, slog.Default())

	createTestLTXFile(t, client, 0, 1, 1)
	createTestLTXFile(t, client, 0, 2, 2)
	createTestLTXFile(t, client, 0, 3, 3)
	client.failWrites = true

	if _, err := compactor.Compact(context.Background(), 1); err == nil {
		t.Fatal("expected error")
	}

	// Compact has returned and closed the streams it opened. Wait for the
	// detached compaction goroutine to finish before checking for leaks.
	deadline := time.NewTimer(5 * time.Second)
	defer deadline.Stop()
	ticker := time.NewTicker(10 * time.Millisecond)
	defer ticker.Stop()
	for countCompactorPipeWriters() > 0 {
		select {
		case <-deadline.C:
			t.Fatal("compactor goroutine did not exit")
		case <-ticker.C:
		}
	}

	if total, open, labels := client.streamStats(); open > 0 {
		t.Fatalf("%d of %d source streams still open after Compact: %v", open, total, labels)
	}
}

// leakCheckCompactionClient tracks every stream opened via OpenLTXFile and
// whether it was closed. The first read of the first source stream parks until
// that stream is closed, holding the compaction goroutine inside a source read
// while WriteLTXFile fails. This pins down the ordering seen in production
// when an upload to a remote replica fails mid-compaction.
type leakCheckCompactionClient struct {
	litestream.ReplicaClient
	failWrites bool
	parked     chan struct{} // closed once a source read is parked

	mu      sync.Mutex
	gated   bool // whether the parking stream has been handed out
	streams []*leakCheckStream
}

func newLeakCheckCompactionClient(path string) *leakCheckCompactionClient {
	return &leakCheckCompactionClient{
		ReplicaClient: file.NewReplicaClient(path),
		parked:        make(chan struct{}),
	}
}

func (c *leakCheckCompactionClient) OpenLTXFile(ctx context.Context, level int, minTXID, maxTXID ltx.TXID, offset, size int64) (io.ReadCloser, error) {
	rc, err := c.ReplicaClient.OpenLTXFile(ctx, level, minTXID, maxTXID, offset, size)
	if err != nil {
		return nil, err
	}

	c.mu.Lock()
	defer c.mu.Unlock()
	s := &leakCheckStream{
		rc:    rc,
		label: fmt.Sprintf("level=%d txid=%s-%s offset=%d", level, minTXID, maxTXID, offset),
		done:  make(chan struct{}),
	}
	if !c.gated {
		c.gated = true
		s.parked = c.parked
	}
	c.streams = append(c.streams, s)
	return s, nil
}

func (c *leakCheckCompactionClient) WriteLTXFile(ctx context.Context, level int, minTXID, maxTXID ltx.TXID, r io.Reader) (*ltx.FileInfo, error) {
	if !c.failWrites {
		return c.ReplicaClient.WriteLTXFile(ctx, level, minTXID, maxTXID, r)
	}

	// Fail the upload only once the compaction goroutine is inside a read of
	// its first source stream.
	select {
	case <-c.parked:
	case <-time.After(5 * time.Second):
		return nil, fmt.Errorf("timeout waiting for compaction goroutine to park")
	}
	return nil, fmt.Errorf("simulated upload failure")
}

func (c *leakCheckCompactionClient) streamStats() (total, open int, labels []string) {
	c.mu.Lock()
	defer c.mu.Unlock()
	for _, s := range c.streams {
		total++
		if !s.isClosed() {
			open++
			labels = append(labels, s.label)
		}
	}
	return total, open, labels
}

type leakCheckStream struct {
	rc     io.ReadCloser
	label  string
	parked chan struct{} // if non-nil, the first read parks until close
	once   sync.Once
	done   chan struct{} // closed when the stream is closed

	mu     sync.Mutex
	closed bool
}

func (s *leakCheckStream) Read(p []byte) (int, error) {
	if s.parked != nil {
		s.once.Do(func() { close(s.parked) })
		<-s.done
	}

	s.mu.Lock()
	closed := s.closed
	s.mu.Unlock()
	if closed {
		return 0, os.ErrClosed
	}
	return s.rc.Read(p)
}

func (s *leakCheckStream) Close() error {
	s.mu.Lock()
	if s.closed {
		s.mu.Unlock()
		return nil
	}
	s.closed = true
	s.mu.Unlock()

	close(s.done)
	return s.rc.Close()
}

func (s *leakCheckStream) isClosed() bool {
	s.mu.Lock()
	defer s.mu.Unlock()
	return s.closed
}
```

Imports needed on top of the existing test imports: `os` and `sync`.

**Expected behavior:**

Once `Compact` and its compaction goroutine have finished, every source stream that was opened is closed, regardless of which side failed first.

**Actual behavior:**

Every source stream that the compaction goroutine touches after `Compact` returns is reopened and never closed. With S3 backed replicas each one is a pinned HTTP connection.

## Mechanism (file and line references on main @ 25ac743)

1. `compactor.go:119` collects sources as `[]io.Reader`; `compactor.go:120-126` defers a close of every reader that implements `io.Closer`. Remote sources are opened at `compactor.go:148` and wrapped in `internal.NewResumableReader` at `compactor.go:152`.
2. `compactor.go:158-167` runs `ltx.Compactor.Compact` in a goroutine that streams into a pipe; `compactor.go:169` uploads the pipe via `WriteLTXFile`. `Compact` never waits for the goroutine: on upload error it closes the pipe (`compactor.go:170`) and returns (`compactor.go:172`), which fires the deferred close of all source streams while the goroutine may still be inside a source read.
3. `internal/resumable_reader.go:145-150`: `Close` closes the current body but does not mark the reader closed.
4. `internal/resumable_reader.go:129-141`: a read failure (including "read on closed stream") drops the stream and retries; `internal/resumable_reader.go:82-97` then reopens from the current offset with a fresh `OpenLTXFile` call. This is correct while compaction is running (that is the #1308/#1326 resumable behavior) but after `Close` it resurrects a stream that has no owner left.
5. The goroutine decodes input headers sequentially (`ltx.Compactor.Compact` reads all headers before writing), so it can resurrect every input, one by one, before it first writes to the closed pipe and exits. All resurrected streams leak.

## Production observation

On a production host replicating a single low write SQLite database to Cloudflare R2 (30s L1 compaction schedule), a corrupt L0 file made every compaction attempt fail, and L0 retention cannot prune uncompacted files, so each attempt opened an ever larger set of source streams and attempts stretched to minutes. Observed on v0.5.15 and unchanged after upgrading to v0.5.16:

- sustained fd growth around 460 fds per minute while compaction retried
- roughly 28,000 open fds and 2.5 GB RSS at peak, then repeated kernel OOM kills
- socket census dominated by dead connections: hundreds in CLOSE_WAIT plus thousands of fds for sockets the kernel had already fully closed, exactly what pinned, never closed response bodies look like

To watch it live: run `litestream replicate` against any S3 endpoint with a compaction that repeatedly fails while uploads abort mid attempt, and watch `lsof -p $(pgrep litestream) | wc -l` climb monotonically across attempts while `ss` shows CLOSE_WAIT accumulating.

## Additional Context

- Not a duplicate of #1354 / #1360: those fixed an unbounded retry budget inside a live `ResumableReader` (connection churn while the reader was still owned). This issue is about streams reopened after the reader was closed, so the fix there does not help (verified on v0.5.16, which contains #1360).
- Related background: #1308 and #1326 introduced the resumable source reads in compaction; #1467 describes large backlog compaction pain that makes the wedged scenario here more likely.
- Aggravator worth noting: the ltx v0.5.1 to v0.5.2 page format change (page header flags, LZ4 block format) is silent and one way. A binary on ltx v0.5.1 (litestream v0.5.15 and earlier) that reads a valid file written under ltx v0.5.2 (v0.5.16 and later) fails page decode with `no flags allowed, reserved for future use`, so a version downgrade can wedge compaction on perfectly valid files, with no actual corruption involved. Same phenomenon class as the closed #754 from the v0.5.0 bump. Any input that fails page decode, for whatever reason, becomes the permanent wedge described above.
- A fix PR follows: make `ResumableReader.Close` sticky (guard the stream with a mutex, fail subsequent reads with `os.ErrClosed`, and never reopen after close), which also makes the detached goroutine exit on its next read instead of resurrecting every input. Includes the repro test above.

akeemjenkins:
## Summary

Makes `internal.ResumableReader.Close` sticky so source streams can no longer be reopened, and therefore leaked, by the detached compaction goroutine after `Compactor.Compact` has returned and closed them. Adds a regression test that fails on the previous code.

Fixes #1492

## Problem

`Compactor.Compact` (compactor.go:158-172) does not wait for its compaction goroutine before returning. When `WriteLTXFile` fails while that goroutine is still reading source files, `Compact` returns and its defer (compactor.go:120-126) closes every source stream, but the goroutine keeps reading. Each read on a closed stream fails, which `ResumableReader` treated as a transient connection failure: it reopened the stream from the current offset (internal/resumable_reader.go, reopen path in `Read`) and kept going until it finally hit the closed pipe. Nothing ever closed those reopened streams. `Close` only closed the current body and did not mark the reader closed, so nothing stopped the reopen.

Evidence, detailed in #1492:

- The new test on the previous code fails with `3 of 6 source streams still open after Compact`: three streams opened, closed by the defer, then all three reopened by the detached goroutine and leaked.
- Reproduced identically on v0.5.15, v0.5.16, and main (25ac743).
- Production observation on a host replicating a low write SQLite database to Cloudflare R2 with a wedged L1 compaction: sustained fd growth (~460 fds per minute), CLOSE_WAIT accumulation, ~28,000 fds and 2.5 GB RSS at peak, repeated OOM kills.
- Verified the other ordering is already correct: when the compaction goroutine fails first (for example a corrupt source file), the existing defer closes everything and nothing leaks, so this PR does not touch `Compact` itself.

## Solution

- Guard `ResumableReader`'s underlying stream with a mutex and add a `closed` flag. `Close` may be called from a different goroutine than `Read` (Compact's cleanup vs its compaction goroutine), and previously those accesses to `rc` were unsynchronized.
- After `Close`, reads fail with `os.ErrClosed` and the reader never reopens the remote stream. If a reopen is in flight when `Close` lands, the freshly opened stream is closed immediately.
- Side benefit: the detached compaction goroutine now exits on its next source read instead of resurrecting and reading every remaining input against a dead pipe.
- Behavior while the reader is live (resume on disconnect, premature EOF handling, bounded retries from #1360) is unchanged; the existing `internal` tests still pass.

## Scope

**In scope:**

- `internal/resumable_reader.go`: sticky, concurrency safe `Close`; no reopen after close.
- `compactor_test.go`: regression test `TestCompactor_CompactDoesNotReopenSourcesAfterClose` plus its close tracking fake replica client.

**Not in scope:**

- Making `Compact` join its compaction goroutine before returning.
- The wedged compaction scenario itself (a corrupt L0 file failing every attempt while the L0 backlog grows); that trigger is what made this leak fatal in production but is a separate issue.

## Test Plan

```bash
# Regression test (fails before this change, passes after)
go test -race -run TestCompactor_CompactDoesNotReopenSourcesAfterClose -v .

# Resumable reader behavior unchanged
go test -race ./internal/

# Full suite
go test -race ./...

# Soak gate per CONTRIBUTING.md (compaction is core replication)
go build -o bin/litestream-test ./cmd/litestream-test
go test -tags 'integration,soak' -run 'TestLTXBehavior$' -short -timeout 25m -v ./tests/integration/
```

All of the above pass on this branch (macOS, Go 1.27). The soak gate passes with all profiles green (`TestLTXBehavior` 506s: low-volume, high-volume, burst-volume, including restoration and integrity checks). Note the explicit `-timeout 25m`: the short mode suite needs a bit over 10 minutes, so go test's default 10m timeout kills it before the last profile otherwise. Before the fix, the regression test fails with:

```text
3 of 6 source streams still open after Compact: [level=0 txid=0000000000000001-0000000000000001 offset=0 ...]
```

`go vet ./...`, `gofmt`, and `goimports -local github.com/benbjohnson/litestream` are clean on the changed files.

## Types of changes

## Checklist

## Related

- Fixes #1492
- Distinct from #1354 / #1360 (bounded retries inside a live reader); verified on v0.5.16 that the fix there does not cover this leak
- Background: #1308 / #1326 introduced resumable source reads in compaction; #1467 describes the large backlog conditions that make the failure ordering here likely

## AI assistance disclosure

Investigated and drafted with AI assistance (Claude, per AI_PR_GUIDE.md), including the production investigation, the repro test, and the fix. Reviewed and submitted by a human account owner.

### DIFF
--- a/compactor_test.go
+++ b/compactor_test.go
@@ -6,8 +6,10 @@ import (
 	"fmt"
 	"io"
 	"log/slog"
+	"os"
 	"runtime"
 	"strings"
+	"sync"
 	"testing"
 	"time"
 
@@ -116,6 +118,50 @@ func TestCompactor_CompactClosesPipeOnWriteError(t *testing.T) {
 			t.Fatalf("compactor goroutine leaked: got %d, want <= %d", countCompactorPipeWriters(), before)
 		case <-ticker.C:
 		}
+	}
+}
+
+// TestCompactor_CompactDoesNotReopenSourcesAfterClose verifies that no source
+// stream is left open once Compact and its compaction goroutine have finished,
+// even when the destination write fails while the goroutine is still reading.
+//
+// Compact does not wait for the compaction goroutine before returning. When
+// WriteLTXFile fails first, Compact returns and closes every source stream
+// while the goroutine may still be reading them. Each subsequent read then
+// fails, which the resumable reader treats as a transient connection failure:
+// it reopens the source from the current offset and keeps going until the
+// goroutine finally hits the closed pipe. Nothing ever closes those reopened
+// streams. Against an S3 replica each one pins an HTTP response body, so a
+// wedged compaction leaks one connection per source file on every attempt.
+func TestCompactor_CompactDoesNotReopenSourcesAfterClose(t *testing.T) {
+	client := newLeakCheckCompactionClient(t.TempDir())
+	compactor := litestream.NewCompactor(client, slog.Default())
+
+	createTestLTXFile(t, client, 0, 1, 1)
+	createTestLTXFile(t, client, 0, 2, 2)
+	createTestLTXFile(t, client, 0, 3, 3)
+	client.failWrites = true
+
+	if _, err := compactor.Compact(context.Background(), 1); err == nil {
+		t.Fatal("expected error")
+	}
+
+	// Compact has returned and closed the streams it opened. Wait for the
+	// detached compaction goroutine to finish before checking for leaks.
+	deadline := time.NewTimer(5 * time.Second)
+	defer deadline.Stop()
+	ticker := time.NewTicker(10 * time.Millisecond)
+	defer ticker.Stop()
+	for countCompactorPipeWriters() > 0 {
+		select {
+		case <-deadline.C:
+			t.Fatal("compactor goroutine did not exit")
+		case <-ticker.C:
+		}
+	}
+
+	if total, open, labels := client.streamStats(); open > 0 {
+		t.Fatalf("%d of %d source streams still open after Compact: %v", open, total, labels)
 	}
 }
 
@@ -730,6 +776,122 @@ func (r *disconnectingReadCloser) Read(p []byte) (int, error) {
 		return n, io.EOF
 	}
 	return n, nil
+}
+
+// leakCheckCompactionClient tracks every stream opened via OpenLTXFile and
+// whether it was closed. The first read of the first source stream parks until
+// that stream is closed, holding the compaction goroutine inside a source read
+// while WriteLTXFile fails. This pins down the ordering seen in production
+// when an upload to a remote replica fails mid-compaction.
+type leakCheckCompactionClient struct {
+	litestream.ReplicaClient
+	failWrites bool
+	parked     chan struct{} // closed once a source read is parked
+
+	mu      sync.Mutex
+	gated   bool // whether the parking stream has been handed out
+	streams []*leakCheckStream
+}
+
+func newLeakCheckCompactionClient(path string) *leakCheckCompactionClient {
+	return &leakCheckCompactionClient{
+		ReplicaClient: file.NewReplicaClient(path),
+		parked:        make(chan struct{}),
+	}
+}
+
+func (c *leakCheckCompactionClient) OpenLTXFile(ctx context.Context, level int, minTXID, maxTXID ltx.TXID, offset, size int64) (io.ReadCloser, error) {
+	rc, err := c.ReplicaClient.OpenLTXFile(ctx, level, minTXID, maxTXID, offset, size)
+	if err != nil {
+		return nil, err
+	}
+
+	c.mu.Lock()
+	defer c.mu.Unlock()
+	s := &leakCheckStream{
+		rc:    rc,
+		label: fmt.Sprintf("level=%d txid=%s-%s offset=%d", level, minTXID, maxTXID, offset),
+		done:  make(chan struct{}),
+	}
+	if !c.gated {
+		c.gated = true
+		s.parked = c.parked
+	}
+	c.streams = append(c.streams, s)
+	return s, nil
+}
+
+func (c *leakCheckCompactionClient) WriteLTXFile(ctx context.Context, level int, minTXID, maxTXID ltx.TXID, r io.Reader) (*ltx.FileInfo, error) {
+	if !c.failWrites {
+		return c.ReplicaClient.WriteLTXFile(ctx, level, minTXID, maxTXID, r)
+	}
+
+	// Fail the upload only once the compaction goroutine is inside a read of
+	// its first source stream.
+	select {
+	case <-c.parked:
+	case <-time.After(5 * time.Second):
+		return nil, fmt.Errorf("timeout waiting for compaction goroutine to park")
+	}
+	return nil, fmt.Errorf("simulated upload failure")
+}
+
+func (c *leakCheckCompactionClient) streamStats() (total, open int, labels []string) {
+	c.mu.Lock()
+	defer c.mu.Unlock()
+	for _, s := range c.streams {
+		total++
+		if !s.isClosed() {
+			open++
+			labels = append(labels, s.label)
+		}
+	}
+	return total, open, labels
+}
+
+type leakCheckStream struct {
+	rc     io.ReadCloser
+	label  string
+	parked chan struct{} // if non-nil, the first read parks until close
+	once   sync.Once
+	done   chan struct{} // closed when the stream is closed
+
+	mu     sync.Mutex
+	closed bool
+}
+
+func (s *leakCheckStream) Read(p []byte) (int, error) {
+	if s.parked != nil {
+		s.once.Do(func() { close(s.parked) })
+		<-s.done
+	}
+
+	s.mu.Lock()
+	closed := s.closed
+	s.mu.Unlock()
+	if closed {
+		return 0, os.ErrClosed
+	}
+	return s.rc.Read(p)
+}
+
+func (s *leakCheckStream) Close() error {
+	s.mu.Lock()
+	if s.closed {
+		s.mu.Unlock()
+		return nil
+	}
+	s.closed = true
+	s.mu.Unlock()
+
+	close(s.done)
+	return s.rc.Close()
+}
+
+func (s *leakCheckStream) isClosed() bool {
+	s.mu.Lock()
+	defer s.mu.Unlock()
+	return s.closed
 }
 
 func countCompactorPipeWriters() int {
--- a/internal/resumable_reader.go
+++ b/internal/resumable_reader.go
@@ -7,6 +7,7 @@ import (
 	"io"
 	"log/slog"
 	"os"
+	"sync"
 	"time"
 
 	"github.com/superfly/ltx"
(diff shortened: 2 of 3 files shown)


corylanou:
Thanks for the detailed investigation and the deterministic compactor regression test. The sticky closed state addresses the stream leak while preserving retry behavior.

During review, I found two follow-up items and pushed commit 00422ba to the PR branch to address them:

- ResumableReader.Read checked a cached terminal error before the closed state. After retry exhaustion followed by Close, later reads could return the old retry error instead of os.ErrClosed. The follow-up gives the closed state precedence and adds regression coverage.
- The in-flight reopen/Close path did not have direct deterministic coverage. The follow-up adds a blocked-opener test confirming that the newly opened stream is closed, the read returns os.ErrClosed, and no additional reopen occurs.

With those changes applied, the synchronization and no-reopen behavior look sound. The full race test suite, pre-commit checks, and CI are green.

PR #1500, opened 2026-09-02
fix(replica): reconcile resumable reader recovery

### LINKED ISSUE #1492: Compaction source streams reopened after Compact returns are never closed, leaking one connection per source file when an upload fails mid-compaction

corylanou:
## Description

Reconciles the resumable reader's retry and lifecycle policies so large transfers can recover from repeated transient disconnects without restoring the unbounded connection churn fixed by #1360. It also makes `Close` terminal and concurrency-safe: later reads return `os.ErrClosed`, no stream is reopened after close, and a stream returned by an in-flight reopen is closed immediately.

This PR incorporates the complete lifecycle work from #1493 by @akeemjenkins, including its deterministic compactor leak regression, and the close-priority follow-up by @corylanou. It addresses the restore-liveness failure reported and initially fixed in #1472 by @darkgnotic while preserving #1360's protection for #1354. The carried commits retain their original authors and source-SHA trailers.

The retry budget is renewed only after a positive underlying read completely satisfies the caller's request. Bytes returned with an error or as a short read still advance the resume offset, but do not clear the connection-churn protection.

## Motivation and Context

The existing lifetime retry limit protects replication from a pathological backend that declares the remaining object length but returns only one byte per connection. Resetting the complete budget on any `n > 0`, as proposed in #1472, lets that case open one connection per byte and prevents replication backoff from engaging.

A pure lifetime limit causes a different failure: #1472 observed disconnects around offsets 1.2 MB and 25 MB in the same file. Despite roughly 24 MB of healthy progress, the fourth lifetime disconnect aborted the restore, causing it to restart without converging.

Completing a caller read is the meaningful-progress invariant. The LTX decoder uses `io.ReadFull` for file headers, page headers, size prefixes, and payloads, so a healthy resumed stream completes these requests. A tiny truncated response does not. This avoids adding an arbitrary byte or time threshold.

| Scenario | Previous behavior | Reconciled behavior |
|---|---|---|
| Five disconnects with 64 KiB progress between each | Retry limit at the fourth disconnect | Completes all 393,216 bytes |
| One-byte truncated responses | `n > 0` reset opens one connection per byte | Retry-limit error within four connections |
| Close during an in-flight reopen | Raced stream could be installed or leaked | Raced stream is closed; read returns `os.ErrClosed` |
| Failed compaction closes source readers | Detached reader could reopen and leak streams | Close remains terminal; no reopen or leak |

Fixes #1492

Supersedes #1472 and incorporates #1493. Related to #1354 and #1360.

## How Has This Been Tested?

Tested on macOS arm64 with the repository-declared Go 1.25.14 toolchain.

```bash
GOTOOLCHAIN=go1.25.14 go test -race -count=1 \
  -run '^(TestResumableReader.*|TestCompactor_Compact(DoesNotReopenSourcesAfterClose|ClosesPipeOnWriteError|ResumesRemoteSourceAfterDisconnect))$' \
  ./internal .

GOTOOLCHAIN=go1.25.14 go test -race -count=1 ./...
GOTOOLCHAIN=go1.25.14 go vet ./...
GOTOOLCHAIN=go1.25.14 staticcheck ./...
GOTOOLCHAIN=go1.25.14 pre-commit run --all-files

golang_toolchain=go1.25.14
env GOTOOLCHAIN="$golang_toolchain" go build -o bin/litestream ./cmd/litestream
env GOTOOLCHAIN="$golang_toolchain" go build -o bin/litestream-test ./cmd/litestream-test
LOG_LEVEL=DEBUG GOTOOLCHAIN="$golang_toolchain" \
  go test -tags 'integration,soak' \
  -run '^TestLTXBehavior$' \
  -short -v -timeout 15m \
  ./tests/integration/
```

All commands pass. The LTX behavioral gate completed in 504.95 seconds; low-volume, high-volume, and burst-volume all passed their behavioral assertions, restoration, and integrity checks.

The progress regression first failed on the #1493 baseline at the fourth disconnect and offset 262,144, then passed with the reconciled policy. The preserved one-byte truncation regression still returns `max retries exceeded` with no more than the initial connection plus three retries.

## Types of changes

## Checklist

- [x] I have updated the documentation accordingly (no documentation change is needed for this internal retry/lifecycle policy)

### DIFF
--- a/compactor_test.go
+++ b/compactor_test.go
@@ -6,8 +6,10 @@ import (
 	"fmt"
 	"io"
 	"log/slog"
+	"os"
 	"runtime"
 	"strings"
+	"sync"
 	"testing"
 	"time"
 
@@ -116,6 +118,50 @@ func TestCompactor_CompactClosesPipeOnWriteError(t *testing.T) {
 			t.Fatalf("compactor goroutine leaked: got %d, want <= %d", countCompactorPipeWriters(), before)
 		case <-ticker.C:
 		}
+	}
+}
+
+// TestCompactor_CompactDoesNotReopenSourcesAfterClose verifies that no source
+// stream is left open once Compact and its compaction goroutine have finished,
+// even when the destination write fails while the goroutine is still reading.
+//
+// Compact does not wait for the compaction goroutine before returning. When
+// WriteLTXFile fails first, Compact returns and closes every source stream
+// while the goroutine may still be reading them. Each subsequent read then
+// fails, which the resumable reader treats as a transient connection failure:
+// it reopens the source from the current offset and keeps going until the
+// goroutine finally hits the closed pipe. Nothing ever closes those reopened
+// streams. Against an S3 replica each one pins an HTTP response body, so a
+// wedged compaction leaks one connection per source file on every attempt.
+func TestCompactor_CompactDoesNotReopenSourcesAfterClose(t *testing.T) {
+	client := newLeakCheckCompactionClient(t.TempDir())
+	compactor := litestream.NewCompactor(client, slog.Default())
+
+	createTestLTXFile(t, client, 0, 1, 1)
+	createTestLTXFile(t, client, 0, 2, 2)
+	createTestLTXFile(t, client, 0, 3, 3)
+	client.failWrites = true
+
+	if _, err := compactor.Compact(context.Background(), 1); err == nil {
+		t.Fatal("expected error")
+	}
+
+	// Compact has returned and closed the streams it opened. Wait for the
+	// detached compaction goroutine to finish before checking for leaks.
+	deadline := time.NewTimer(5 * time.Second)
+	defer deadline.Stop()
+	ticker := time.NewTicker(10 * time.Millisecond)
+	defer ticker.Stop()
+	for countCompactorPipeWriters() > 0 {
+		select {
+		case <-deadline.C:
+			t.Fatal("compactor goroutine did not exit")
+		case <-ticker.C:
+		}
+	}
+
+	if total, open, labels := client.streamStats(); open > 0 {
+		t.Fatalf("%d of %d source streams still open after Compact: %v", open, total, labels)
 	}
 }
 
@@ -730,6 +776,122 @@ func (r *disconnectingReadCloser) Read(p []byte) (int, error) {
 		return n, io.EOF
 	}
 	return n, nil
+}
+
+// leakCheckCompactionClient tracks every stream opened via OpenLTXFile and
+// whether it was closed. The first read of the first source stream parks until
+// that stream is closed, holding the compaction goroutine inside a source read
+// while WriteLTXFile fails. This pins down the ordering seen in production
+// when an upload to a remote replica fails mid-compaction.
+type leakCheckCompactionClient struct {
+	litestream.ReplicaClient
+	failWrites bool
+	parked     chan struct{} // closed once a source read is parked
+
+	mu      sync.Mutex
+	gated   bool // whether the parking stream has been handed out
+	streams []*leakCheckStream
+}
+
+func newLeakCheckCompactionClient(path string) *leakCheckCompactionClient {
+	return &leakCheckCompactionClient{
+		ReplicaClient: file.NewReplicaClient(path),
+		parked:        make(chan struct{}),
+	}
+}
+
+func (c *leakCheckCompactionClient) OpenLTXFile(ctx context.Context, level int, minTXID, maxTXID ltx.TXID, offset, size int64) (io.ReadCloser, error) {
+	rc, err := c.ReplicaClient.OpenLTXFile(ctx, level, minTXID, maxTXID, offset, size)
+	if err != nil {
+		return nil, err
+	}
+
+	c.mu.Lock()
+	defer c.mu.Unlock()
+	s := &leakCheckStream{
+		rc:    rc,
+		label: fmt.Sprintf("level=%d txid=%s-%s offset=%d", level, minTXID, maxTXID, offset),
+		done:  make(chan struct{}),
+	}
+	if !c.gated {
+		c.gated = true
+		s.parked = c.parked
+	}
+	c.streams = append(c.streams, s)
+	return s, nil
+}
+
+func (c *leakCheckCompactionClient) WriteLTXFile(ctx context.Context, level int, minTXID, maxTXID ltx.TXID, r io.Reader) (*ltx.FileInfo, error) {
+	if !c.failWrites {
+		return c.ReplicaClient.WriteLTXFile(ctx, level, minTXID, maxTXID, r)
+	}
+
+	// Fail the upload only once the compaction goroutine is inside a read of
+	// its first source stream.
+	select {
+	case <-c.parked:
+	case <-time.After(5 * time.Second):
+		return nil, fmt.Errorf("timeout waiting for compaction goroutine to park")
+	}
+	return nil, fmt.Errorf("simulated upload failure")
+}
+
+func (c *leakCheckCompactionClient) streamStats() (total, open int, labels []string) {
+	c.mu.Lock()
+	defer c.mu.Unlock()
+	for _, s := range c.streams {
+		total++
+		if !s.isClosed() {
+			open++
+			labels = append(labels, s.label)
+		}
+	}
+	return total, open, labels
+}
+
+type leakCheckStream struct {
+	rc     io.ReadCloser
+	label  string
+	parked chan struct{} // if non-nil, the first read parks until close
+	once   sync.Once
+	done   chan struct{} // closed when the stream is closed
+
+	mu     sync.Mutex
+	closed bool
+}
+
+func (s *leakCheckStream) Read(p []byte) (int, error) {
+	if s.parked != nil {
+		s.once.Do(func() { close(s.parked) })
+		<-s.done
+	}
+
+	s.mu.Lock()
+	closed := s.closed
+	s.mu.Unlock()
+	if closed {
+		return 0, os.ErrClosed
+	}
+	return s.rc.Read(p)
+}
+
+func (s *leakCheckStream) Close() error {
+	s.mu.Lock()
+	if s.closed {
+		s.mu.Unlock()
+		return nil
+	}
+	s.closed = true
+	s.mu.Unlock()
+
+	close(s.done)
+	return s.rc.Close()
+}
+
+func (s *leakCheckStream) isClosed() bool {
+	s.mu.Lock()
+	defer s.mu.Unlock()
+	return s.closed
 }
 
 func countCompactorPipeWriters() int {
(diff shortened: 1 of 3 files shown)


emschwartz:
I ran into this problem as well and would love to see this merged in a future release.

PR #1509, opened 2026-09-04
fix(s3lease): support Hetzner lease semantics

hannesortmeier:
## Summary

Make the S3 lease implementation in #1317 compatible with Hetzner Object Storage while preserving the existing AWS S3 behavior.

This PR is intentionally stacked on #1317 so it can be reviewed or absorbed into that feature branch as one small follow-up commit.

## Problem

A live compatibility test against a disposable prefix in a Hetzner test bucket showed two differences from AWS S3:

| Request | Hetzner response |
| --- | --- |
| `PUT If-None-Match: *` against an existing lock | `412 PreconditionFailed`; object preserved |
| `PUT If-Match` with the quoted ETag returned by S3 | `412 PreconditionFailed` |
| `PUT If-Match` with the same ETag unquoted | `200`; stale unquoted ETags still return `412` |
| `DELETE If-Match` with a deliberately wrong ETag | `204`; object was deleted |

On the unmodified #1317 head (`07202a4`), the exact `s3lease` binary acquired its lock but failed on the first renewal:

```text
Error: renew lease: lease not held
```

The unconditional delete behavior is also unsafe for fencing: a stale owner could delete a newer owner's lock.

## Solution

- Propagate the configured S3 endpoint to the leaser.
- For Hetzner endpoints, remove the surrounding quotes from ETags used by conditional PUTs.
- Release a Hetzner lease by conditionally replacing it with an expired lease instead of using conditional DELETE.
- Re-read after a failed release write to preserve the distinction between an already released lease and a lease held by another owner.
- Document the provider-specific behavior.

AWS S3 and other providers continue using their existing quoted ETags and conditional DELETE release path.

## Scope

**In scope:**
- Hetzner-specific S3 lease renewal, takeover, and release behavior.
- Unit coverage for renewal, safe release, stale owners, and already-released leases.
- Provider compatibility documentation.

**Not in scope:**
- General Litestream replication behavior.
- Changes to lease timing or the `s3lease` process lifecycle.
- Compatibility changes for other S3 providers.

## Test Plan

```bash
go test ./...
go test -race ./...
pre-commit run --all-files
```

All commands pass locally, including `go vet`, `goimports`, and `staticcheck` through pre-commit.

The patched `s3lease` binary was also tested end-to-end against a unique disposable Hetzner object prefix with a 4-second TTL and 1-second heartbeat:

- the first owner remained alive across multiple renewals;
- a second owner timed out while the lease was held;
- the second owner acquired immediately after the first owner's PUT-based release;
- the disposable lock object was removed after the test.

## Related

- Stacked on #1317
- Related to #1301

### DIFF
--- a/docs/PROVIDER_COMPATIBILITY.md
+++ b/docs/PROVIDER_COMPATIBILITY.md
@@ -154,6 +154,8 @@ Related issues: #912
 - `InvalidArgument` errors with default AWS SDK settings
 - Does not support `aws-chunked` content encoding
 - Requires signed payloads for all requests
+- Conditional PUT requests require unquoted ETags
+- Conditional DELETE requests ignore `If-Match`; lease release uses a conditional PUT
 
 **Configuration**:
 
--- a/s3/leaser.go
+++ b/s3/leaser.go
@@ -43,10 +43,11 @@ type Leaser struct {
 	s3     S3API
 	logger *slog.Logger
 
-	Bucket string
-	Path   string
-	TTL    time.Duration
-	Owner  string
+	Bucket   string
+	Path     string
+	Endpoint string
+	TTL      time.Duration
+	Owner    string
 }
 
 func NewLeaser() *Leaser {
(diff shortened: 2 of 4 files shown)


corylanou:
@hannesortmeier — I pushed one commit to this branch (dab765e7) while reviewing. Two things came out of it.

The commit adds `TestLeaser_RenewLease_HetznerExplicitPort`, a copy of your existing Hetzner test with an explicit `:443` on the endpoint. It's skipped, because it fails: `extractEndpointHost` in `replica_url.go` returns `url.URL.Host`, which keeps the port, so the suffix match against `.your-objectstorage.com` misses and the lease falls back to the generic S3 path with a quoted If-Match. That's pre-existing on main rather than anything this PR introduced — the same miss hits Tigris, DigitalOcean and Backblaze detection too — but it's worth covering here because your change is what makes the wrong path a lease-safety problem instead of a cosmetic one. I filed it separately as #1523 with the fix (`u.Hostname()`); unskip the test once that lands.

The second thing I couldn't settle on my own. Hetzner's supported-actions page lists "Conditional PUT/DELETE operations on versioned Buckets" as not supported, and the leaser relies on If-Match for both the renew and the delete path. Versioning is opt-in there and isn't even exposed in their UI, so a normal bucket is unaffected — but for someone who turned it on with `mc`, I can't tell from the docs whether Hetzner rejects the conditional request or accepts it while ignoring the condition. Those are very different outcomes: one is a loud failure, the other is silent lease loss. I don't have a Hetzner account to test against.

If you have a bucket handy, a versioned-bucket PUT with a deliberately stale If-Match would answer it in one request. If not, we'll take the docs at face value and note the limitation rather than hold the PR on it.

PR #1514, opened 2026-09-09
feat(replicate): upload snapshots directly to L9, reduce compaction overhead

### LINKED ISSUE #1510: Compaction from L0 snapshot writes DB-sized LTX's across all levels
## Bug Description

The first compaction into each destination level re-compacts the entire TXID history from TXID 1 — including a DB-sized base snapshot — so a freshly-replicated database is rewritten and re-uploaded in full, once per compaction level. On a large database each of these is an O(database size) read + merge + upload, failing to take the advantage of the incremental benefit of relying on the L9 snapshot.

### Where it happens

`Compactor.Compact` chooses its source range from the **destination** level's high-water mark:

```go
// compactor.go
prevMaxInfo, err := c.MaxLTXFileInfo(ctx, dstLevel)   // dst level's max
seekTXID := prevMaxInfo.MaxTXID + 1                    // == 1 when dst is empty
itr, err := c.client.LTXFiles(ctx, srcLevel, seekTXID, false)
```

When a destination level is empty, `prevMaxInfo.MaxTXID == 0`, so `seekTXID == 1` and the compaction pulls every source-level file from TXID 1 onward — which, on a fresh generation, begins with the DB-sized base snapshot L0. The resulting L1 file therefore spans `[1 .. N]` and is itself DB-sized. When L2 first compacts, it pulls L1 from TXID 1 (again DB-sized) → L2 `[1 .. M]`; then L3 likewise. The base propagates up the entire ladder, one full-DB rewrite per level.

## Environment

**Litestream version:**

```text
main @ v0.5.17-1-g4ed7a30
```

**Operating system & version:** Linux (container), arm64
**Installation method:** built from source
**Storage backend:** S3-compatible (reproduces on any backend; the cost is read+merge+upload of the whole DB per level)

## Steps to Reproduce

1. Replicate a large database (ours is ~65 GB) from empty, so the base snapshot L0 is DB-sized.
2. Keep the DB under steady write load so incremental L0 files accumulate.
3. Watch the remote store (or in-progress multipart uploads). As each ladder level (L1, then L2, then L3) is first populated, observe a DB-sized object whose key spans `MinTXID = 1` — i.e. the base is being re-written into that level.

**Expected:** the incremental ladder compacts only the increments *above the latest snapshot*. The DB-sized base is written once (to the snapshot level) and never re-materialized into L1/L2/L3. Per-level compaction cost is O(new data), not O(database size).

**Actual:** each level's first compaction re-reads/re-writes/re-uploads the entire database (spanning from TXID 1), because an empty destination seeks from TXID 1.

## Evidence

From one ~65 GB generation (object key = `<level>/<minTXID>-<maxTXID>.ltx`, size in bytes):

```text
# Base capture — the DB written twice at TXID 1 (level 0 and the snapshot level):
0000/0000000000000001-0000000000000001.ltx   65,390,548,046   # base snapshot L0 (~60.9 GiB)
0009/0000000000000001-0000000000000001.ltx   65,390,548,046   # snapshot level, identical size

# Then each ladder level's FIRST compaction re-includes the base (note MinTXID = 1),
# observed as successive DB-sized multipart uploads:
0001/0000000000000001-0000000000000039.ltx   ~65 GB   # L1  first compaction  (base + increments)
0002/0000000000000001-00000000000000b3.ltx   ~65 GB   # L2  first compaction  (pulls L1 from TXID 1)
0003/0000000000000001-00000000000000b3.ltx   ~65 GB   # L3  first compaction  (pulls L2 from TXID 1)
```

Timeline on our deployment (build includes serialized compactions, similar to what will land in https://github.com/benbjohnson/litestream/pull/1479):

| step | key range | window | duration |
|---|---|---|---|
| snapshot (level 9) | `1–1` | 18:56:12 → 19:13:18 | ~17 min |
| L1 first compaction | `1–0x39` | 19:14:01 → ~19:33 | ~19 min |
| L2 first compaction | `1–0xb3` | ~19:40 → 19:57 | ~17 min |
| L3 first compaction | `1–0xb3` | 20:00 → … | ~17 min |

Because of compaction serialization, these large db-sized writes result in incremental L0 files accumulating from 55 to 172+ over the same ~25 minutes. Net per fresh generation: ~4–5 sequential full-DB rewrites (≈ an hour of monopolized compaction).

## Additional Context

**Suggested fix.** When compacting into an empty destination level, seek from the latest snapshot watermark + 1 instead of TXID 1, so the ladder never re-includes the base:

```go
seekTXID := prevMaxInfo.MaxTXID + 1
if prevMaxInfo.MaxTXID == 0 {              // destination level never populated
    seekTXID = latestSnapshotMaxTXID + 1   // skip everything already in the snapshot
}
```

This is restore-safe with no other change, because restore already reconstructs from the snapshot and layers increments on top:

- `CalcRestorePlan` (replica.go) starts from the latest snapshot ≤ target ("Start with latest snapshot…", `infos = append(infos, snapshot)`), then greedily extends the contiguous TXID range from the other levels.
- The extension requires contiguity (`info.MinTXID > currentMax+1` disqualifies a candidate), which is preserved exactly: the first ladder file now begins at `snapshotMax + 1`, so it joins the snapshot with no gap. Today's plan merely *overlaps* the base (snapshot `[..1]` + L1 `[1..]`); the fix removes the redundant re-coverage.

Guard for the empty-destination case only (not `max(snapMax, dstMax)`): a non-empty level must keep seeking from `dstMax + 1` so a periodic snapshot that leapfrogs a lagging level cannot open an intra-level gap that trips `VerifyLevelConsistency` (the `prev.MaxTXID + 1 == curr.MinTXID` check).

### Separate but related

**The base is also written twice — concurrently — at TXID 1.** On a fresh generation the full database is materialized as both an L0 file (`0000/0001`, the streaming initial capture) and a snapshot-level file (`0009/0001`). These are produced by two independent code paths — the sync/replication path and the snapshot-level compaction — under two different locks, so on a large DB they run at the same time: two full-DB reads, two DB-sized in-memory page indexes, and two DB-sized uploads. In our trace both `0000/0001` and `0009/0001` were actively uploading in the same window (~18:56:12 → ~18:57:04+), and the snapshot level is an independent full re-read rather than a promotion of the L0 file.

We suspect this independence is deliberate (keeping the sync path and compaction decoupled), but the consequence scales with database size: it roughly doubles both the peak page-index memory (the very pressure #1479 targets, via a path that mutex doesn't cover) and the initial upload volume. It's a slightly separate issue from the cascade, but has the same "the DB-sized base gets processed more times than necessary" character, so we're flagging it here.

**A more complete alternative — we'd value your read on it.** Beyond the localized fix above, we're considering writing the base once, directly to the snapshot level (local + remote), and starting the L0 stream at the first increment (TXID = `snapshotMax + 1`) rather than writing a DB-sized L0 base at all. Because restore already seeds from the snapshot and layers increments on top, this would address **both** problems at once: no DB-sized L0 base (so the ladder — which pulls from L0 — never sees the base, making the cascade fix unnecessary), and no concurrent second capture (the snapshot-level monitor becomes a no-op when the snapshot is already current, so it doesn't re-read the DB).

The cost is that the position/anchor logic — which today derives the latest replicated position from the newest **L0** file — must become **level-aware**, so it can read the position from the snapshot level when no L0 increment exists yet, while preserving the invariant that the file holding the current max TXID is locally present. We're inclined toward this as our end state, but we'd very much appreciate the maintainers weighing in: is there something that fundamentally requires a level-0 file at the base TXID, or another pitfall we've missed? Happy to open a separate discussion/PR if that's a better venue.

**Relationship to other issues.**
- Complementary to #1479: serializes the concurrent DB-sized page indexes; this issue (and the alternative above) removes the *reason* several of them are DB-sized in the first place. With both, compaction is neither concurrent-and-OOM nor sequential-and-monopolizing.
- On our large base, the first incremental sync after the base also scans the whole base file in `lastPageMatch` (filed separately, #1508) — different code path, same "O(DB size) on a hot path" flavor.

darkgnotic:
# Upload sync snapshots directly to L9 to eliminate unnecessary Compaction overhead

## Summary

Sync snapshots — the initial base at TXID 1 and mid-stream boundary snapshots
taken at checkpoint boundaries — are now uploaded directly to the L9 snapshot
level, instead of landing in L0 and being pulled up through the compaction
ladder. This keeps DB-sized full images out of the L0→L1→L2→L3 ladder entirely.

## Background

Every write, including a snapshot, previously landed in L0, and compaction
re-materialized that whole DB-sized image into each ladder level in turn:

- On a fresh generation, a ~120 GB database gets written roughly 5× — the L0
  base, the L9 snapshot, and again into L1, L2, and L3, each spanning
  `MinTXID=1`.
- Large compactions are resource intensive, resulting in memory issues such
  as #1477 and necessitating mitigations such as serializing them (#1479).
  This in aggregate results in monopolizing the DB for long stretches (order
  of an hour per generation on large DBs) while L0 files pile up.
- This repeats on every restart that mints a new generation.

Fixes #1510

## Approach

Let the sync path own snapshot capture and route the bytes straight to the L9
snapshot level, so the compaction ladder never sees a full-DB image. Snapshots
remain visible to restore (at L9) and continue to serve as the local resume
anchor, but they no longer participate in the ladder.

Supporting components are updated to understand the new semantics:
- **replication** distinguishes between snapshots (MinLTX = 1) and increments
  (MinLTX = MaxLTX) and uploads them to the appropriate level
- sync snapshots are locally deleted after their next file has been uploaded,
  since they are otherwise no longer tracked by L0 retention
- **compaction** avoids doing the initial `[1,1]` snapshot so as to eliminate
  redundant snapshotting when a generation begins. All levels instead wait for
  an L9 snapshot to exist and then compact LTX files after it.
- **restore following** checks the L9 level when encountering a gap
  (it previously only consulted up to L8)

## Reviewer's guide

The diff is ~255 lines of production change across 4 files and ~429 lines of
test change across 6. The tests are mechanical fallout (see Testing), so reading
the production code first is enough to understand the change. Suggested order —
it follows a snapshot's life from capture to reclamation:

1. **`db.go` — `sync()`** (the origin). A snapshotting write is tagged
   `[1, txID]` and leaves `l0FileInfo` nil. _Check:_ the `[1, k]` marker is
   unambiguous (no increment is ever `[1, x]`) and snapshots no longer advance
   the L0 ladder max.
2. **`db.go` — `uploadTarget`, `openLevel0Anchor`, `currentSnapshotMaxTXID`**
   (resolution). How a `[1, k]` file is recognized and how the anchor is
   resolved when the name isn't `[txID, txID]`. _Check:_ the ordering in
   `uploadTarget` (snapshot name tried first — the two collide at `fromTXID==1`)
   and the cold-resume fallback to the live snapshot.
3. **`replica.go` — `syncOnce`, `uploadLTXFile`** (routing). `MinTXID==1` →
   remote L9, increments → remote L0; the loop advances by the file's actual
   `MaxTXID`. _Check:_ re-encountering a snapshot TXID is idempotent.
4. **`compactor.go` — `Compact`, `VerifyLevelConsistency`** (the ladder skips
   snapshots). _Check:_ `seekTXID = max(dstMax, snapMax) + 1`, the require-a-
   snapshot guard, and that a snapshot-spanned gap is accepted (loaded lazily).
5. **`store.go` — `CompactDB`** (monitor stand-down). _Check:_ the snapshot
   monitor stands down at `pos <= 1` so it can't race the base capture.
6. **`replica.go`/`db.go` — `calcRemotePos` vs `calcPos`, `SyncStatus`**
   (reporting). _Check:_ status uses `max(L0, L9)` while the resume cursor stays
   L0-only.
7. **`db.go` — `deleteSupersededLocalSnapshot`** (reclamation). _Check:_ the
   single-`[1, prevTXID]`-delete argument and the single-replica assumption.

For the test churn, the two files worth actually reading are `compactor_test.go`
(the gap-spanned-by-snapshot verification and the seek behavior) and
`store_test.go` (the snapshot-monitor stand-down and base auto-delivery). The
rest is find-and-replace: removing now-redundant manual snapshot seeds and
updating expected TXID ranges / remote-level counts.

### Snapshot capture and tagging (`sync`)

- Snapshotting writes are tagged `[1, txID]` (`MinTXID=1`) in the local L0
  directory; increments stay `[txID, txID]`. The `[1, k]` tag is a durable,
  stateless marker of "this is a snapshot" that survives restart — no increment
  is ever `[1, x]`, because the first write is always the base.
- Snapshotting writes no longer set `l0FileInfo`, so they don't advance the L0
  max the compaction ladder reads. The local file is still written to the L0
  directory so `MaxLTX`/`Pos` continue to find the anchor.

### Upload routing (`Replica.syncOnce`, `uploadTarget`)

- The upload loop resolves each file through `DB.uploadTarget`: files tagged
  `MinTXID=1` upload to remote L9, increments upload to remote L0.
- `openLevel0Anchor` / `currentSnapshotMaxTXID` resolve the local `[1, k]`
  anchor for the verify, WAL-offset, and cold-resume paths that previously
  assumed a `[txID, txID]` filename.

### Compaction (`Compactor.Compact`, `VerifyLevelConsistency`)

- `Compact` seeks past the newest snapshot: `max(dstMax, snapMax) + 1`. A
  snapshot at L9 leaves a "hole" in the lower levels; seeking past it means the
  hole is skipped rather than mistaken for a gap.
- `Compact` requires an initial snapshot to exist before it will compact — a
  ladder with no snapshot to anchor it is unrestorable.
- `VerifyLevelConsistency` accepts a TXID gap when an L9 snapshot spans it
  (loaded lazily, so the common gapless path makes no extra request).

### Snapshot monitor stand-down (`Store.CompactDB`)

- The snapshot-level `CompactDB` stands down at `pos <= 1`. The base is the sync
  path's responsibility; standing down avoids a concurrent second full-DB read
  racing the sync path's base capture.

### Position reporting (`calcRemotePos` vs `calcPos`)

- Snapshots live only at L9, so a replica that has uploaded the base but no
  increments is fully caught up. `SyncStatus` now reports remote position via
  `calcRemotePos` = `max(L0 max, L9 max)`, so a boundary snapshot with no
  trailing increment doesn't read as lag.
- `calcPos` stays L0-only: it is the sync loop's resume cursor, and keeping it
  on L0 preserves point-in-time-restore granularity (every increment is still
  uploaded to L0 even when a snapshot at L9 already covers a higher TXID).
  Re-encountering a snapshot TXID is harmless — `uploadTarget` re-routes it to
  L9 idempotently.

### Local disk reclamation (`deleteSupersededLocalSnapshot`)

- Because a sync-path snapshot uploads to L9, remote-listing-driven L0 retention
  never sees it, so it would otherwise never be reclaimed locally. Once the
  replica has advanced past a snapshot (its bytes durable at L9 and its
  succeeding file uploaded), its local file is deleted to reclaim disk.
- Deleting only `[1, prevTXID]` suffices because the contiguous walk visits
  every superseded snapshot as `prevTXID` exactly once. Like all local
  retention, this assumes a single replica walks the L0 directory.

## Invariants and assumptions

- **Single replica per DB.** All local-retention paths already assume one
  consumer walks the local L0 directory (`DB.Replica` is a singular field with
  no fan-out). `deleteSupersededLocalSnapshot` makes the same assumption; a
  second independent replica sharing these files is not a supported
  configuration.
- **`[1, k]` as a stateless snapshot marker.** No increment is ever `[1, x]`, so
  the filename alone durably distinguishes a snapshot from an increment across
  restarts, with no side metadata to keep in sync.

## Testing

- `go test ./...` passes.
- `go test -race -count=5 .` passes.
- The test diff is largely mechanical: snapshots are now auto-delivered to L9 by
  the sync path, so tests that previously seeded a snapshot by hand or asserted
  `[txID, txID]` names / L0-only remote counts were updated accordingly.

## Out of scope

While running the suite under `-race`, we hit a pre-existing, unrelated data
race in `FuzzRestoreWithMissingCompactedFile`. `git diff` confirms this change
does not touch the affected code. `Store.Close()` closes each `DB` before it
cancels its own context and waits for the compaction-monitor goroutines to stop,
so a monitor can still be mid-`db.Snapshot()` (reading `db.f`) at the moment
`db.Close()` nils it out under a separate lock. It's a shutdown-ordering bug
worth a separate issue/PR, but it is independent of this change and not
addressed here.

### DIFF
--- a/_examples/library/library_example_test.go
+++ b/_examples/library/library_example_test.go
@@ -87,6 +87,23 @@ func TestLibraryExampleFileBackend(t *testing.T) {
 	if err := restoreReplica.Restore(ctx, opt); err != nil {
 		t.Fatalf("restore: %v", err)
 	}
+
+	// Verify the restored database contains the replicated row. The write lived
+	// only in the base snapshot (ltx/9), so this exercises the snapshot-level
+	// restore path end to end.
+	restoredDB, err := sql.Open("sqlite", restorePath)
+	if err != nil {
+		t.Fatalf("open restored db: %v", err)
+	}
+	defer restoredDB.Close()
+
+	var message string
+	if err := restoredDB.QueryRowContext(ctx, `SELECT message FROM events WHERE id = 1`).Scan(&message); err != nil {
+		t.Fatalf("query restored row: %v", err)
+	}
+	if message != "hello" {
+		t.Fatalf("restored message = %q, want %q", message, "hello")
+	}
 }
 
 func openAppDB(ctx context.Context, path string) (*sql.DB, error) {
@@ -108,7 +125,10 @@ func openAppDB(ctx context.Context, path string) (*sql.DB, error) {
 func waitForLTXFiles(replicaPath string, timeout time.Duration) error {
 	deadline := time.Now().Add(timeout)
 	for time.Now().Before(deadline) {
-		matches, err := filepath.Glob(filepath.Join(replicaPath, "ltx", "0", "*.ltx"))
+		// Match any level: the initial database state is captured as a base
+		// snapshot uploaded directly to the snapshot level (ltx/9), while later
+		// increments land at ltx/0. Either one means replication has occurred.
+		matches, err := filepath.Glob(filepath.Join(replicaPath, "ltx", "*", "*.ltx"))
 		if err != nil {
 			return fmt.Errorf("glob ltx files: %w", err)
 		}
--- a/compactor.go
+++ b/compactor.go
@@ -108,7 +108,20 @@ func (c *Compactor) Compact(ctx context.Context, dstLevel int) (*ltx.FileInfo, e
 	if err != nil {
 		return nil, fmt.Errorf("cannot determine max ltx file for destination level: %w", err)
 	}
-	seekTXID := prevMaxInfo.MaxTXID + 1
+
+	snapInfo, err := c.MaxLTXFileInfo(ctx, SnapshotLevel)
+	if err != nil {
+		return nil, fmt.Errorf("cannot determine snapshot max ltx file: %w", err)
+	}
+
+	// An initial snapshot is required before compaction can begin.
+	if snapInfo.MaxTXID == 0 {
+		return nil, ErrNoCompaction
+	}
+
+	// Snapshots are uploaded directly to the L9 SnapshotLevel, creating a "hole"
+	// in lower levels. Always seek past the newest snapshot.
+	seekTXID := max(prevMaxInfo.MaxTXID, snapInfo.MaxTXID) + 1
 
 	itr, err := c.client.LTXFiles(ctx, srcLevel, seekTXID, false)
 	if err != nil {
@@ -201,6 +214,35 @@ func (c *Compactor) VerifyLevelConsistency(ctx context.Context, level int) error
 	}
 	defer itr.Close()
 
+	// A ladder level skips TXIDs covered by a snapshot (Compact seeks past the
+	// latest snapshot), so a gap is valid when a SnapshotLevel file spans the
+	// missing range. Loaded lazily on the first gap so the common (gapless)
+	// path makes no extra request.
+	var snapshots []*ltx.FileInfo
+	snapshotsLoaded := false
+	gapCoveredBySnapshot := func(g0, g1 ltx.TXID) (bool, error) {
+		if !snapshotsLoaded {
+			sitr, err := c.client.LTXFiles(ctx, SnapshotLevel, 0, false)
+			if err != nil {
+				return false, fmt.Errorf("fetch snapshot ltx files: %w", err)
+			}
+			for sitr.Next() {
+				s := *sitr.Item()
+				snapshots = append(snapshots, &s)
+			}
+			if err := sitr.Close(); err != nil {
+				return false, fmt.Errorf("close snapshot iterator: %w", err)
+			}
+			snapshotsLoaded = true
+		}
+		for _, s := range snapshots {
+			if s.MinTXID <= g0 && s.MaxTXID >= g1 {
+				return true, nil
+			}
+		}
+		return false, nil
+	}
+
 	var prevInfo *ltx.FileInfo
 	for itr.Next() {
 		info := itr.Item()
@@ -215,11 +257,20 @@ func (c *Compactor) VerifyLevelConsistency(ctx context.Context, level int) error
 		expectedMinTXID := prevInfo.MaxTXID + 1
 		if info.MinTXID != expectedMinTXID {
 			if info.MinTXID > expectedMinTXID {
-				return fmt.Errorf("TXID gap detected: prev.MaxTXID=%s, next.MinTXID=%s (expected %s)",
-					prevInfo.MaxTXID, info.MinTXID, expectedMinTXID)
+				covered, err := gapCoveredBySnapshot(expectedMinTXID, info.MinTXID-1)
+				if err != nil {
+					return err
+				}
+				if !covered {
+					return fmt.Errorf("TXID gap detected: prev.MaxTXID=%s, next.MinTXID=%s (expected %s)",
+						prevInfo.MaxTXID, info.MinTXID, expectedMinTXID)
+				}
+				// Gap is spanned by a snapshot; restore seeds from it and
+				// extends contiguously. Accept and continue.
+			} else {
+				return fmt.Errorf("TXID overlap detected: prev.MaxTXID=%s, next.MinTXID=%s",
+					prevInfo.MaxTXID, info.MinTXID)
 			}
-			return fmt.Errorf("TXID overlap detected: prev.MaxTXID=%s, next.MinTXID=%s",
-				prevInfo.MaxTXID, info.MinTXID)
 		}
 
 		prevInfo = info
--- a/compactor_test.go
+++ b/compactor_test.go
@@ -22,9 +22,11 @@ func TestCompactor_Compact(t *testing.T) {
 		client := file.NewReplicaClient(t.TempDir())
 		compactor := litestream.NewCompactor(client, slog.Default())
 
-		// Create test L0 files
-		createTestLTXFile(t, client, 0, 1, 1)
+		// The base is captured in the snapshot at TXID 1; L0 holds increments
+		// from TXID 2 onward. Compaction into the empty L1 seeks from snapMax+1.
+		createTestLTXFile(t, client, litestream.SnapshotLevel, 1, 1)
 		createTestLTXFile(t, client, 0, 2, 2)
+		createTestLTXFile(t, client, 0, 3, 3)
 
 		info, err := compactor.Compact(context.Background(), 1)
 		if err != nil {
@@ -33,8 +35,8 @@ func TestCompactor_Compact(t *testing.T) {
 		if info.Level != 1 {
 			t.Errorf("Level=%d, want 1", info.Level)
 		}
-		if info.MinTXID != 1 || info.MaxTXID != 2 {
-			t.Errorf("TXID range=%d-%d, want 1-2", info.MinTXID, info.MaxTXID)
+		if info.MinTXID != 2 || info.MaxTXID != 3 {
+			t.Errorf("TXID range=%d-%d, want 2-3", info.MinTXID, info.MaxTXID)
 		}
 	})
 
@@ -52,29 +54,30 @@ func TestCompactor_Compact(t *testing.T) {
 		client := file.NewReplicaClient(t.TempDir())
 		compactor := litestream.NewCompactor(client, slog.Default())
 
-		// Create L0 files
-		createTestLTXFile(t, client, 0, 1, 1)
+		// Base in the snapshot at TXID 1; L0 increments from TXID 2.
+		createTestLTXFile(t, client, litestream.SnapshotLevel, 1, 1)
 		createTestLTXFile(t, client, 0, 2, 2)
+		createTestLTXFile(t, client, 0, 3, 3)
 
-		// Compact to L1
+		// Compact to L1 (seeks from snapMax+1 = 2)
 		_, err := compactor.Compact(context.Background(), 1)
 		if err != nil {
 			t.Fatal(err)
 		}
 
 		// Create more L0 files
-		createTestLTXFile(t, client, 0, 3, 3)
+		createTestLTXFile(t, client, 0, 4, 4)
 
-		// Compact to L1 again (should only include TXID 3)
+		// Compact to L1 again (should only include TXID 4)
 		info, err := compactor.Compact(context.Background(), 1)
 		if err != nil {
 			t.Fatal(err)
 		}
-		if info.MinTXID != 3 || info.MaxTXID != 3 {
-			t.Errorf("TXID range=%d-%d, want 3-3", info.MinTXID, info.MaxTXID)
+		if info.MinTXID != 4 || info.MaxTXID != 4 {
+			t.Errorf("TXID range=%d-%d, want 4-4", info.MinTXID, info.MaxTXID)
 		}
 
-		// Now compact L1 to L2 (should include all from 1-3)
+		// Now compact L1 to L2 (empty L2 seeks from snapMax+1 = 2, pulling all L1)
 		info, err = compactor.Compact(context.Background(), 2)
 		if err != nil {
 			t.Fatal(err)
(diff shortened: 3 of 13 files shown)


darkgnotic:
FYI, here is an example of the replicated files from our staging environment (running this change):

```
$ aws s3 ls --recursive s3://my-bucket/2026-08-27T22:00:21.312Z/1788975115363/
2026-09-09 10:34:13     1153692 2026-08-27T22:00:21.312Z/1788975115363/0000/0000000000000002-0000000000000002.ltx
2026-09-09 10:34:13      441165 2026-08-27T22:00:21.312Z/1788975115363/0000/0000000000000003-0000000000000003.ltx
2026-09-09 10:34:14      208516 2026-08-27T22:00:21.312Z/1788975115363/0000/0000000000000004-0000000000000004.ltx
2026-09-09 10:34:14         194 2026-08-27T22:00:21.312Z/1788975115363/0000/0000000000000005-0000000000000005.ltx
2026-09-09 10:34:18     1154722 2026-08-27T22:00:21.312Z/1788975115363/0000/0000000000000006-0000000000000006.ltx
2026-09-09 10:34:18     1155407 2026-08-27T22:00:21.312Z/1788975115363/0000/0000000000000007-0000000000000007.ltx
2026-09-09 10:34:18     1154624 2026-08-27T22:00:21.312Z/1788975115363/0000/0000000000000008-0000000000000008.ltx
2026-09-09 10:34:18     1154589 2026-08-27T22:00:21.312Z/1788975115363/0000/0000000000000009-0000000000000009.ltx
2026-09-09 10:34:18      427349 2026-08-27T22:00:21.312Z/1788975115363/0000/000000000000000a-000000000000000a.ltx
2026-09-09 10:34:18         194 2026-08-27T22:00:21.312Z/1788975115363/0000/000000000000000b-000000000000000b.ltx
2026-09-09 10:34:29    30825646 2026-08-27T22:00:21.312Z/1788975115363/0000/000000000000000c-000000000000000c.ltx
2026-09-09 10:34:44    45667509 2026-08-27T22:00:21.312Z/1788975115363/0000/000000000000000d-000000000000000d.ltx
2026-09-09 10:34:45     4737910 2026-08-27T22:00:21.312Z/1788975115363/0000/000000000000000e-000000000000000e.ltx
... ellided ...
2026-09-09 10:57:44     1153882 2026-08-27T22:00:21.312Z/1788975115363/0000/00000000000000d7-00000000000000d7.ltx
2026-09-09 10:57:44     1153469 2026-08-27T22:00:21.312Z/1788975115363/0000/00000000000000d8-00000000000000d8.ltx
2026-09-09 10:57:44         194 2026-08-27T22:00:21.312Z/1788975115363/0000/00000000000000d9-00000000000000d9.ltx
2026-09-09 10:57:59     1155567 2026-08-27T22:00:21.312Z/1788975115363/0000/00000000000000da-00000000000000da.ltx
2026-09-09 10:57:59         194 2026-08-27T22:00:21.312Z/1788975115363/0000/00000000000000db-00000000000000db.ltx
2026-09-09 10:58:28     1155146 2026-08-27T22:00:21.312Z/1788975115363/0000/00000000000000dc-00000000000000dc.ltx
2026-09-09 10:58:29     1155161 2026-08-27T22:00:21.312Z/1788975115363/0000/00000000000000dd-00000000000000dd.ltx

2026-09-09 10:36:01   135897184 2026-08-27T22:00:21.312Z/1788975115363/0001/0000000000000002-000000000000002c.ltx
2026-09-09 10:38:03     1155910 2026-08-27T22:00:21.312Z/1788975115363/0001/000000000000002d-000000000000003d.ltx
2026-09-09 10:42:01     1155470 2026-08-27T22:00:21.312Z/1788975115363/0001/000000000000003e-000000000000005d.ltx
2026-09-09 10:44:01     1155448 2026-08-27T22:00:21.312Z/1788975115363/0001/000000000000005e-000000000000006e.ltx
2026-09-09 10:46:01     1156069 2026-08-27T22:00:21.312Z/1788975115363/0001/000000000000006f-0000000000000079.ltx
2026-09-09 10:48:01    78351686 2026-08-27T22:00:21.312Z/1788975115363/0001/000000000000007a-0000000000000080.ltx
2026-09-09 10:50:02    64412954 2026-08-27T22:00:21.312Z/1788975115363/0001/0000000000000081-000000000000009a.ltx
2026-09-09 10:52:02     1155587 2026-08-27T22:00:21.312Z/1788975115363/0001/000000000000009b-00000000000000aa.ltx
2026-09-09 10:54:01     1154850 2026-08-27T22:00:21.312Z/1788975115363/0001/00000000000000ab-00000000000000ba.ltx
2026-09-09 10:56:01     1156104 2026-08-27T22:00:21.312Z/1788975115363/0001/00000000000000bb-00000000000000cb.ltx
2026-09-09 10:58:01     1155632 2026-08-27T22:00:21.312Z/1788975115363/0001/00000000000000cc-00000000000000db.ltx

2026-09-09 10:40:01   135897543 2026-08-27T22:00:21.312Z/1788975115363/0002/0000000000000002-000000000000003d.ltx
2026-09-09 11:00:01   140546672 2026-08-27T22:00:21.312Z/1788975115363/0002/000000000000003e-00000000000000db.ltx

2026-09-09 10:33:31 11137455622 2026-08-27T22:00:21.312Z/1788975115363/0009/0000000000000001-0000000000000001.ltx
```

bitfliq:
I was testing this branch for an unrelated reason and hit something involving follow mode. It depends on an ordering that may not be representative, and you have context I don't, so this is a question rather than a report.

Running the same script against `main` and 9e491f5, writing, a boundary snapshot, an L1 compaction, then L0 retention, with nothing deleted by hand:

| | L1 after compaction | follower parked at TXID 3 |
|---|---|---|
| `main` (4ed7a308) | `[1,12]` | advances to 12 |
| this branch, L1 empty when the snapshot-spanning compaction ran | `[6,12]` | stays at 3 |
| this branch, L1 already covering `[2,5]` | `[2,5] [6,12]` | advances to 12 |

The third row is the one that corrected my first reading, and it is why I am not claiming anything general: when L1 has already compacted over the snapshot's range the ladder stays contiguous and the follower recovers normally. The middle row only strands it because the seek moved past TXIDs no lower level had recorded, and L0 retention then pruned them (its guard is `info.MaxTXID <= maxL1TXID`, which is satisfied even where L1 has a hole).

In that middle case the follower returned no error and logged no ERROR or WARN while the replica advanced to 12. My reading is that `fillFollowGap` walks `for level := 1; level < SnapshotLevel; level++`, so the follow path does not consult L9, whereas `CalcRestorePlan` does start from the snapshot level, which would match the note added to `VerifyLevelConsistency`. If there is an equivalent path on the follow side that I have missed, that would explain it and I would rather know.

The follow-side behaviour is already on `main`, so it is not something this PR introduces; I have filed it separately as #1515 with a fix in #1516 that only makes the condition visible. The open question I could not answer from outside is whether that ordering is one worth designing around here, or whether it is rare enough not to matter. I have not measured how often it occurs.

darkgnotic:
Thank you for the report @bitfliq! You are right in that the follow logic needed to be updated to understand the new flow of data from L0. I've fixed this and also added WARN-level diagnostics that you mentioned were missing. Please let me know if this is still an issue.

bitfliq:
Confirmed fixed on my side, thank you for turning that around so quickly.

I re-ran the same scenario against d2e542c: the follower now advances 3 --> 5 off the L9 snapshot and then 5 --> 12 back on the ladder, and ends caught up. The two-step handoff in `fillFollowGapFromSnapshot` does exactly what the comment says it does, and it covers the case I hit.

One small thing, and it may be deliberate: the new WARN has no per-position deduplication, so it fires once per poll rather than once on entering the state. In a case the snapshot fallback cannot rescue I measured 60 lines in ~60 polls:

```
level=WARN msg="follow: unable to bridge gap; follower stalled" current_txid=0000000000000005 gap_min_txid=0000000000000007
```

Worth saying that the case I had to build to reach it is narrow now that the fallback exists: the follower has to sit exactly at the newest snapshot's `MaxTXID`, with a hole above it that no ladder level covers, so the snapshot cannot advance it either. I have not tried to work out how often that occurs, and it may be rare enough not to matter to you.

If you do want it, deduplicating needs a little state carried across polls, since `fillFollowGap` is re-entered from scratch each time. I threaded a position through as an out-parameter in #1516, but a field on `Replica` would work as well and might sit better in your structure. Happy to send it as a PR against your branch, or to just leave the approach here for you to take, whichever is less friction.

On that: #1516 now overlaps your change, and yours is the more complete fix since it addresses the bridging rather than only the reporting. If the maintainers would rather take yours whole, I am glad to close mine; it only really matters that `main` ends up with one of them.

Issue #1270, opened 2026-05-12
Restore: TOCTOU race between CalcRestorePlan and OpenLTXFile when L0 retention deletes a planned file mid-restore


esqandil:
## Bug Description

`litestream restore` (latest-restore mode, no `-txid` / `-timestamp`) fails intermittently with:

```
level=ERROR msg="failed to run" error="open ltx file: file does not exist"
```

when the running `litestream replicate` daemon's L0 retention monitor concurrently deletes an L0 LTX file that the restore plan has already picked. The data is not lost — Azure (or any backend) returns `BlobNotFound` because retention has deleted the file after it was compacted into L1+. The data is still available at L1+, but `Restore()` does not re-plan when `OpenLTXFile` returns `os.ErrNotExist`; it bails out.

In our setup the race is reproducible against high-write DBs with short sync intervals (5s); we hit it on attempt 1 of a clean `litestream restore` against a live ABS replica.

## Root cause (v0.5.11 source trace)

`Replica.Restore()` calls `CalcRestorePlan()` at the start ([`replica.go:611`](https://github.com/benbjohnson/litestream/blob/v0.5.11/replica.go#L611)), then iterates the resulting `[]*ltx.FileInfo` and calls `OpenLTXFile` for each ([`replica.go:638`](https://github.com/benbjohnson/litestream/blob/v0.5.11/replica.go#L638)):

```go
infos, err := CalcRestorePlan(ctx, r.Client, opt.TXID, opt.Timestamp, r.Logger())
// ...
for _, info := range infos {
    // ...
    f, err := r.Client.OpenLTXFile(ctx, info.Level, info.MinTXID, info.MaxTXID, 0, 0)
    if err != nil {
        return fmt.Errorf("open ltx file: %w", err)   // <-- this is our error
    }
    // ...
}
```

Between `CalcRestorePlan` building the list and the fetch loop opening each file, the running `replicate` daemon's L0 retention monitor (default `interval=15s retention=5m0s`, configurable via top-level `l0-retention` / `l0-retention-check-interval`) can delete an L0 file that satisfies its retention criteria (already compacted into L1+ AND older than the retention window). The restore is now reading a stale plan.

`OpenLTXFile` on the ABS backend ([`abs/replica_client.go:280`](https://github.com/benbjohnson/litestream/blob/v0.5.11/abs/replica_client.go#L280)) maps `BlobNotFound` → `os.ErrNotExist`:

```go
if isNotExists(err) {
[...]

Issue #235, opened 2021-09-30
Replica.monitor logs context.Canceled


colin-sitehost:
`Replica.monitor` is `context.Context` aware, and passes that into `Replica.Sync` calls within. unfortunately, if the context is cancelled during the call to `Replica.Sync`, because all `err != nil` `errors` are logged, we get a spurious `log.Printf`. this is also non-fatal, since we only log these `error`s from `Replica.Sync`.

spit balling solutions, we could ignore `context.Canceled` (and possibly `context.DeadlineExceeded`):
```diff
diff --git a/replica.go b/replica.go
index 0cfc21d..aa48091 100644
--- a/replica.go
+++ b/replica.go
@@ -3,6 +3,7 @@ package litestream
 import (
        "context"
        "encoding/binary"
+       "errors"
        "fmt"
        "hash/crc64"
        "io"
@@ -627,7 +628,7 @@ func (r *Replica) monitor(ctx context.Context) {
                notify = r.db.Notify()

                // Synchronize the shadow wal into the replication directory.
-               if err := r.Sync(ctx); err != nil {
+               if err := r.Sync(ctx); err != nil && !errors.Is(err, context.Canceled) {
                        log.Printf("%s(%s): monitor error: %s", r.db.Path(), r.Name(), err)
                        continue
                }
```
but it may also be better to ignore any errors if `ctx.Done()` is closed, or punt the issue upstream, since I do not know the canonical solution to this problem.

hifi:
This would indeed be nice. We get a lot of these logs during clean shutdown. There's more places where Litestream logs context canceled.

PR #1516, opened 2026-09-10
fix(follow): warn when follow mode cannot advance past a gap

### LINKED ISSUE #1515: `restore -f` stops making progress permanently and logs nothing when it cannot bridge a gap
## Bug Description

A `restore -f` follower can stop advancing permanently while emitting no error and no warning. It stays alive, keeps polling, and keeps serving reads from a database that never changes again. A restart resumes from the `-txid` sidecar and stops at the same TXID. A plain `litestream restore` from the same replica, at the same moment, restores the database completely, so the data is reachable; the follow path just cannot reach it.

Not a duplicate of the other silent-stall reports: #1037 and #1083 were the primary-side WAL change-detection stall (fixed), and #1310 was initial sync wedging on a full disk. This is the follower after it is already running. Open PR #1385 relaxes the sibling "ahead of latest snapshot" resume guard and says it leaves the "behind the earliest snapshot" check alone, so it does not cover this either.

## Environment

**Litestream version:**
```text
0.5.17
```

Also reproduced on 0.5.14 and 0.5.16. `replica.go` is byte-identical on v0.5.17 and `main` (checked at 4ed7a308f627), and `fillFollowGap` is unchanged since v0.5.14.

**Operating system & version:** Debian bookworm-slim container, Linux x86_64

**Installation method:** official .deb from the GitHub release

**Storage backend:** S3 (MinIO in the reproduction; also seen on a hosted S3-compatible store)

## Steps to Reproduce

Self-contained, no credentials, ~3 minutes: `docker compose run --rm repro` (the four files are inline below so they can be read before running; MinIO runs in the compose project). They are also in a gist if cloning is easier: https://gist.github.com/bitfliq/8b3c982599a7af5fc1b48420d37a7c86

```bash
git clone https://gist.github.com/bitfliq/8b3c982599a7af5fc1b48420d37a7c86.git repro && cd repro && docker compose run --rm repro
```

1. Replicate a WAL-mode SQLite database, writing ~24 transactions.
2. Delete **one** L0 object from the replica, standing in for an upload that never landed. The script asserts L0 is contiguous first, and picks a TXID above the L1 watermark, so the hole is real. (If L1 has already absorbed that TXID, `fillFollowGap` bridges via L1 and the follower correctly recovers; that is the control.)
3. Keep writing, so the frontier advances.
4. Park a follower one TXID below the hole and run `restore -f`.
5. Run a plain `litestream restore` from the same replica as a control.

<details><summary>reproduction files (4 files, self-contained)</summary>

**`docker-compose.yml`**

```yaml
services:
  minio:
    image: minio/minio:latest
    command: server /data
    environment:
      MINIO_ROOT_USER: minioadmin
      MINIO_ROOT_PASSWORD: minioadmin
    healthcheck:
      test: ["CMD", "mc", "ready", "local"]
      interval: 3s
      timeout: 3s
      retries: 20

  createbucket:
    image: minio/mc:latest
    depends_on:
      minio: {condition: service_healthy}
    entrypoint: >
      /bin/sh -c "mc alias set local http://minio:9000 minioadmin minioadmin &&
                  mc mb --ignore-existing local/repro"

  repro:
    build:
      context: .
      args:
        LITESTREAM_VERSION: ${LITESTREAM_VERSION:-0.5.17}
    depends_on:
      createbucket: {condition: service_completed_successfully}
    environment:
      AWS_ACCESS_KEY_ID: minioadmin
      AWS_SECRET_ACCESS_KEY: minioadmin
```

**`Dockerfile`**

```dockerfile
FROM debian:bookworm-slim
ARG LITESTREAM_VERSION=0.5.17
# Populated automatically by BuildKit (amd64 / arm64), so the reproduction runs
# on an Apple Silicon laptop as well as an x86_64 server.
ARG TARGETARCH
RUN apt-get update && apt-get install -y --no-install-recommends \
      ca-certificates curl sqlite3 python3 && rm -rf /var/lib/apt/lists/*
RUN set -eu; \
    case "${TARGETARCH:-amd64}" in \
      amd64) LS_ARCH=x86_64 ;; \
      arm64) LS_ARCH=arm64  ;; \
      *) echo "unsupported TARGETARCH: ${TARGETARCH}" >&2; exit 1 ;; \
    esac; \
    curl -fsSL -o /tmp/ls.deb \
      "https://github.com/benbjohnson/litestream/releases/download/v${LITESTREAM_VERSION}/litestream-${LITESTREAM_VERSION}-linux-${LS_ARCH}.deb"; \
    dpkg -i /tmp/ls.deb; rm /tmp/ls.deb
RUN set -eu; \
    curl -fsSL -o /usr/local/bin/mc \
      "https://dl.min.io/client/mc/release/linux-${TARGETARCH:-amd64}/mc"; \
    chmod +x /usr/local/bin/mc
COPY litestream.yml /etc/litestream.yml
COPY repro.sh /repro.sh
RUN chmod +x /repro.sh
ENTRYPOINT ["/repro.sh"]
```

**`litestream.yml`**

```yaml
# Intervals compressed so the scenario completes in ~3 minutes.
# Defaults (L1 30s / L9 24h / l0-retention 5m) behave the same, just slower.
levels:
  - interval: 45s
  - interval: 5m
  - interval: 1h

snapshot:
  # Snapshots must track the frontier, as they do on any long-running database.
  # With only a txid-1 snapshot the follower instead hits the documented
  # "saved TXID is ahead of latest snapshot" refusal, which is a different path.
  interval: 20s
  retention: 24h

l0-retention: 300s
l0-retention-check-interval: 5s

dbs:
  - path: /data/primary/db.sqlite
    replicas:
      - type: s3
        bucket: repro
        path: db
        endpoint: http://minio:9000
        region: us-east-1
        force-path-style: true
        sync-interval: 1s
```

**`repro.sh`**

```bash
#!/usr/bin/env bash
# Follow mode stops making progress, forever, without logging anything.
#
# Setup: one L0 object is removed from the replica, standing in for an upload
# that never landed. Everything after that is Litestream's own behaviour.
#
# Expected: the follower reports that it cannot proceed (some error, or a
#   recovery), the way `restore` does.
# Observed: the follower logs nothing after "entering follow mode", never
#   advances, survives restart -- while `litestream restore` from the SAME
#   replica at the SAME moment restores the database completely.
set -uo pipefail
DB=/data/primary/db.sqlite
mkdir -p /data/primary /data/follower /data/control
log(){ printf '\n=== %s\n' "$*"; }
fail(){ printf '\nPRECONDITION FAILED: %s\n' "$*"; exit 2; }

# Defaults target the bundled MinIO; override to point at any S3-compatible store.
S3_ENDPOINT="${S3_ENDPOINT:-http://minio:9000}"
S3_BUCKET="${S3_BUCKET:-repro}"
S3_PREFIX="${S3_PREFIX:-db}"
mc alias set L "$S3_ENDPOINT" "${AWS_ACCESS_KEY_ID}" "${AWS_SECRET_ACCESS_KEY}" >/dev/null
# Start from an empty replica. Without this, a re-run writes a fresh database
# (txid 1..N) into a bucket that still holds the previous run's objects, and the
# leftovers look exactly like naturally-occurring holes.
mc rm --recursive --force "L/$S3_BUCKET/$S3_PREFIX" >/dev/null 2>&1 || true
# TXIDs are 16-hex in the object names; print them as decimal to avoid confusion.
lvl(){ mc ls --recursive "L/$S3_BUCKET" 2>/dev/null | awk '{print $NF}' | grep "^$S3_PREFIX/$1/" \
       | sed -E 's|.*/([0-9a-f]{16})-([0-9a-f]{16})\.ltx|\1 \2|' \
       | python3 -c 'import sys
for line in sys.stdin:
    a,b=line.split()
    print(f"{int(a,16)}-{int(b,16)}")'; }

litestream version
sqlite3 "$DB" "pragma journal_mode=WAL; create table t(id integer primary key, v text);" >/dev/null

litestream replicate > /tmp/primary.log 2>&1 &
sleep 3
for i in $(seq 1 12); do sqlite3 "$DB" "insert into t(v) values('a$i');" >/dev/null; sleep 0.6; done
sleep 3

log "replica contents"
echo "L0: $(lvl 0000 | tr '\n' ' ')"
echo "L1: $(lvl 0001 | tr '\n' ' ')"

# Assert L0 is contiguous BEFORE we remove anything, so the hole we demonstrate
# is unambiguously the one we made.
N0=$(lvl 0000 | wc -l)
(( N0 >= 8 )) || fail "expected >=8 L0 objects before removal, saw $N0 (listing broken or nothing replicated)"
GAPS=$(lvl 0000 | awk -F- 'NR>1 && $1 != prev+1 {print prev"->"$1} {prev=$2}')
[[ -n "$GAPS" ]] && fail "L0 already has gaps before removal: $GAPS"
echo "L0 is contiguous before removal: OK ($N0 objects)"

# Choose a victim that L1 has NOT already absorbed. If L1 already covers the TXID,
# removing its L0 is a no-op: fillFollowGap bridges via L1 and the follower correctly
# catches up. That is a useful positive control but it is not the scenario under test,
# and picking blindly makes the reproduction flaky.
MAXL1=$(lvl 0001 | awk -F- '{print $2}' | sort -n | tail -1); MAXL1=${MAXL1:-0}
echo "highest TXID already compacted into L1: $MAXL1"

VICTIM=""; VLO=0
while read -r key; do
  hex=$(sed -E 's|.*/([0-9a-f]{16})-[0-9a-f]{16}\.ltx|\1|' <<<"$key")
  lo=$((16#$hex))
  if (( lo > MAXL1 + 1 )); then VICTIM="$key"; VLO=$lo; break; fi
done < <(mc ls --recursive "L/$S3_BUCKET" | awk '{print $NF}' | grep "^$S3_PREFIX/0000/" | sort)

[[ -n "$VICTIM" ]] || fail "no L0 object above the L1 watermark ($MAXL1) available to remove"
PARK=$(( VLO - 1 ))
echo "victim TXID $VLO is above the L1 watermark; follower will park at $PARK"

if [[ "${NEGATIVE_CONTROL:-0}" == "1" ]]; then
  log "NEGATIVE CONTROL: removing nothing. The follower must catch up."
else
  log "removing ONE L0 object (simulating an upload that never landed): $VICTIM"
  mc rm "L/$S3_BUCKET/$VICTIM" >/dev/null
fi

for i in $(seq 13 24); do sqlite3 "$DB" "insert into t(v) values('b$i');" >/dev/null; sleep 0.6; done

log "waiting for L1 compaction to run (45s interval)"
sleep 60
echo "L1 is now: $(lvl 0001 | tr '\n' ' ')      <-- frozen; see the errors below"
grep -m2 'compaction failed' /tmp/primary.log

COVER=""
for L in 0000 0001 0002 0003; do
  c=$(lvl "$L" | awk -F- -v n="$VLO" -v lab="$L" '$1<=n && n<=$2 {print lab" "$1"-"$2}')
  [[ -n "$c" ]] && COVER="$COVER $c"
done
if [[ "${NEGATIVE_CONTROL:-0}" == "1" ]]; then
  echo "negative control: TXID $VLO coverage is:${COVER:- none}"
else
  [[ -z "$COVER" ]] || fail "TXID $VLO is still covered by:$COVER -- no hole exists, nothing to test"
  echo "no incremental coverage for TXID $VLO: confirmed"
fi

log "park a follower just below the removed txid, then follow"
# restore to an exact txid, then write the sidecar exactly as litestream does
# (fmt.Fprintln => 16 hex digits + newline), so `restore -f` resumes from there
# precisely as it would after a process restart.
PARKHEX=$(printf '%016x' "$PARK")
litestream restore -txid "$PARKHEX" -o /data/follower/db.sqlite "$DB" >/dev/null 2>&1
printf '%s\n' "$PARKHEX" > /data/follower/db.sqlite-txid
# LOG_LEVEL is read in the shared config path (cmd/litestream/main.go), so it
# applies to `restore` as well -- `restore` has no -log-level flag of its own.
LOG_LEVEL=trace litestream restore -f -follow-interval 1s -o /data/follower/db.sqlite "$DB" \
  > /tmp/follower.log 2>&1 &
FPID=$!
for s in 15 30 45 60; do
  sleep 15
  printf '  t+%-3s txid=%s rows=%s alive=%s\n' "${s}s" \
    "$(tr -d '\n' < /data/follower/db.sqlite-txid)" \
    "$(sqlite3 "file:/data/follower/db.sqlite?mode=ro" 'select count(*) from t;')" \
    "$(kill -0 $FPID 2>/dev/null && echo yes || echo no)"
done
kill $FPID 2>/dev/null; wait $FPID 2>/dev/null

log "follower log, in full (LOG_LEVEL=trace)"
cat /tmp/follower.log
echo "error/warning lines: $(grep -ciE 'error|warn' /tmp/follower.log)"

log "control: plain restore from the SAME replica"
litestream restore -o /data/control/db.sqlite "$DB" 2>&1 | tail -2
echo "rows restored: $(sqlite3 "file:/data/control/db.sqlite?mode=ro" 'select count(*) from t;')"
echo "rows in primary: $(sqlite3 "file:$DB?mode=ro" 'select count(*) from t;')"
echo "highest txid in replica: $( { lvl 0000; lvl 0001; lvl 0009; } | awk -F- '{print $2}' | sort -n | tail -1)"
```

</details>

**Expected behavior:** the follower reports that it cannot proceed, via an error or the "delete and re-restore" refusal it already emits for other unrecoverable states.

**Actual behavior:** it advances no further, logs nothing about it, and stays that way. 11 of 11 runs where the hole was asserted present; 2 of 2 controls with nothing deleted caught up fully.

## Configuration

```yaml
levels:
  - interval: 45s
  - interval: 5m
  - interval: 1h
snapshot:
  interval: 20s        # must track the frontier; see note below
  retention: 24h
l0-retention: 300s
l0-retention-check-interval: 5s
dbs:
  - path: /data/primary/db.sqlite
    replicas:
      - type: s3
        bucket: repro
        path: db
        endpoint: http://minio:9000
        region: us-east-1
        force-path-style: true
        sync-interval: 1s
```

</details>

A snapshot must exist above the follower's position, otherwise the documented `saved TXID ... is ahead of latest snapshot` refusal fires instead and the failure is loud. Any long-running database satisfies that; a 3-minute test needs the short interval.

## Logs

<details><summary>one run, unedited</summary>

```
=== park a follower just below the removed txid, then follow
  t+15s txid=0000000000000004 rows=4 alive=yes
  t+30s txid=0000000000000004 rows=4 alive=yes
  t+45s txid=0000000000000004 rows=4 alive=yes
  t+60s txid=0000000000000004 rows=4 alive=yes

=== follower log, in full
time=2026-08-31T04:08:46.335Z level=INFO msg="resuming follow mode from crash recovery" db=db.sqlite replica=s3 txid=0000000000000004 output=/data/follower/db.sqlite
time=2026-08-31T04:08:46.335Z level=INFO msg="entering follow mode" db=db.sqlite replica=s3 output=/data/follower/db.sqlite txid=0000000000000004 interval=1s
time=2026-08-31T04:09:46.331Z level=INFO msg="follow mode stopped" db=db.sqlite replica=s3
error/warning lines: 0
```

</details>

At `LOG_LEVEL=trace` the follower is not literally silent (it logs routine S3 traffic), but there is no ERROR or WARN in the entire stall window, and the output is indistinguishable from an idle follower with nothing to do. It issues a LIST against `db/0000/` through `db/0008/` on every poll: nine requests per interval, forever. At `-follow-interval 1s` that is roughly 780k LIST requests per day per stalled database, which may be of interest alongside #1390 and #1468.

### Scope

**What this reports:** follow mode makes no progress and emits no error or warning when it cannot bridge a gap, and a restart resumes into the same state.

**What it does not claim:** that L0 objects go missing in practice. The reproduction removes one by hand, and so do the primary-side experiments below. They show what Litestream does *given* a missing object, not how often that happens. The one case I can show arising with nothing removed is the interaction with open PR #1514, below.

**What it proposes:** surfacing the condition (avenue 1), for which there is a PR. It takes no position on whether the gap-fill bound itself should change (avenue 2). That is a separate decision with consequences I cannot evaluate from outside.

### Root cause

`applyNewLTXFiles` (`replica.go:865`) polls level 0 and calls `fillFollowGap` (`replica.go:1007`) when it sees a gap. That bridges levels 1..8:

```go
for level := 1; level < SnapshotLevel; level++ {
```

`SnapshotLevel` is 9, so snapshots are never consulted. If no level can bridge, it returns the unchanged TXID and a **nil** error. Back in `follow()` (`replica.go:804`), a nil error with `newTXID == lastTXID` takes neither the error branch nor the `newTXID > lastTXID` branch, so nothing is logged on that iteration, or any subsequent one, since nothing in the loop changes.

Restart does not help. The resume guard at `replica.go:653` is:

```go
"cannot resume follow mode: saved TXID %s is behind the earliest snapshot (min TXID %s); replica history has been pruned -- delete %s and %s-txid to re-restore"
```

It names the right remedy, but it compares against `latestSnapshot.MinTXID`, and every snapshot object is `0009/0000000000000001-<max>`. A trace-level LIST of that prefix during the reproduction returns

```
[0000000000000001-0000000000000001.ltx 0000000000000001-0000000000000002.ltx
 0000000000000001-0000000000000011.ltx 0000000000000001-0000000000000012.ltx]
```

so `MinTXID` is 1 and the check cannot fire for any saved TXID >= 1.

This is the shape `AI_PR_GUIDE.md`'s own checklist warns about: *"Does the code return errors to callers? Watch for `log.Printf(err)` followed by `continue` or `return nil`: this silently swallows failures."* Here there is not even a log line.

### How the gap can arise on the primary

The reproduction above removes an L0 object by hand. Separately, I looked at what Litestream itself does when an L0 object is missing, because that determines whether this state is reachable without anyone deleting anything.

`Compact()` seeks from `W+1` (`W` = highest TXID in L1) and hands whatever `LTXFiles` returns to `ltx.NewCompactor`, which validates the inputs it is given. So the outcome depends on whether the **surviving inputs are still contiguous from the seek point**, not on where the missing file sits. Four conditions, three runs each, all twelve consistent:

| deleted | survivors | compaction | L1 |
|---|---|---|---|
| `W+1` | `W+2,W+3,…` contiguous | advances past the hole (1 transient failure) | permanent gap `W → W+2` |
| `W+1,W+2` | `W+3,…` contiguous | advances past the hole (1 transient failure) | permanent gap `W → W+3` |
| `W+2` | `W+1,W+3,…` broken | fails repeatedly, never recovers | frozen at `W` |
| `W+1,W+3` | `W+2,W+4,…` broken | fails repeatedly, never recovers | frozen at `W` |

Rows 2 and 4 both delete at the window start and behave oppositely, which is what rules out position as the cause.

**The quiet outcome is the one that strands a follower.** In the first case the gap is permanent (held across a four-minute window while L1 advanced from 15 to 33), and a follower parked at `W` stayed there for the whole window, alive and polling at 1s, with **zero error or warning lines**, while the missing TXID was covered only by a level-9 snapshot. In the loud case the follower actually advances one step, because the L0 file it needs is still there; compaction is wedged but the follow path is not blocked by an unbridgeable gap.

`verify-compaction: true` reports the quiet case by name (`TXID gap detected: prev.MaxTXID=…, next.MinTXID=… (expected …)`) and stays silent in the loud one, since it only runs after a successful compaction. It is off by default.

**Two limits on this.** The initiating L0 is still removed by hand, so this shows what happens *given* a lost upload, not that uploads are lost in practice. And it is a `file://` replica in one configuration.

### Interaction with open PR #1514

**This is the part I would most like a second opinion on, because it changes how reachable the stall is.** Everything above reaches the stall by removing an L0 object by hand. Open PR #1514 (*"upload snapshots directly to L9, reduce compaction overhead"*) can reach the same stall with nothing deleted, though only under a particular ordering, which the third row of the table below pins down.

That PR routes snapshots straight to L9 and changes the compaction seek to `seekTXID := max(prevMaxInfo.MaxTXID, snapInfo.MaxTXID) + 1`, described in its own comment as

> Snapshots are uploaded directly to the L9 SnapshotLevel, creating a "hole" in lower levels.
> Always seek past the newest snapshot.

It then relaxes `VerifyLevelConsistency` to accept a gap when a snapshot spans it, on the grounds that *"restore seeds from it and extends contiguously."* That holds for `restore`, which consults the snapshot level. It does not hold for `follow`, which does not: `fillFollowGap` stops at `SnapshotLevel - 1`.

I ran the same script against both, doing nothing but writing, snapshotting, compacting, and letting L0 retention age files out, with **no object removed by hand at any point**:

| | L1 after compaction | L9 | follower parked at TXID 3 |
|---|---|---|---|
| `main` (4ed7a308) | `[1,12]` | `[1,5]` | advances to 12 |
| PR #1514, L1 empty when the snapshot-spanning compaction first ran | `[6,12]` | `[1,1] [1,5]` | **stays at 3, silently** |
| PR #1514, L1 already covering `[2,5]` before the snapshot | `[2,5] [6,12]` | `[1,1] [1,5]` | advances to 12 |

All three end with L0 pruned to `[12,12]` by retention (10-11 files), and in all three the follow call returned a nil error rather than refusing.

**What the third row means, and how narrow this is.** My first reading of this was too broad and the third row is what corrected it. I assumed any follower positioned below the newest snapshot would be stranded, because the seek moves past the whole snapshot span. That is wrong: if L1 has *already* compacted over that range, its existing files still cover it, the ladder stays contiguous, and the follower recovers exactly as it does on `main`. The hole only becomes permanent when the snapshot-spanned range was never compacted into L1 in the first place, so the seek jumps past TXIDs no lower level ever recorded. In my runs that meant a boundary snapshot landing before L1's first compaction over the same range.

So the trigger is an ordering, not a general property: a boundary snapshot covering `[1, k]`, a compaction that then seeks from `k+1` while L1 holds nothing below it, and L0 retention subsequently pruning that range. Retention does prune it, because its guard is the scalar `info.MaxTXID <= maxL1TXID` rather than a check that L1 contiguously covers the TXID. I have not measured how often that ordering arises in a real deployment, and I have not tried to reason about it from the compaction and checkpoint intervals; it may be rare or it may be common on a fresh generation, and you are far better placed to judge that than I am.

The runs used an explicit `db.Snapshot` and a very short `l0-retention` to reach the state in one second rather than one retention window. Both are ordinary configuration, but it is one synthetic sequence, not an observation from production.

Under the PR, TXIDs 2–5 exist only in the L9 snapshot. L1 legitimately starts at 6, because that is what the new seek is designed to do. A follower sitting at TXID 3 when L0 retention runs then has no reachable path forward in that configuration, and takes the silent branch described above: the replica advanced to TXID 12 while the follower stayed at 3, logging nothing but `entering follow mode` and, on cancellation, `follow mode stopped`.

I am not arguing against #1514. It is addressing a real cost (#1510, #1477, #1479), the hole in the lower levels is a deliberate part of that, and on the evidence above it is not a general property of the change. What I can say is narrower: there is at least one ordering under which the ladder ends up with a hole that only L9 fills, and the follow path is the one reader that does not consult L9. Whether that ordering is worth designing around is a judgement I cannot make from outside.

A correction to my own first guess, in case it saves someone the detour: I initially expected the boundary snapshot to punch a hole in **L0** at its own TXID. It does not; L0 stays contiguous from TXID 2. The hole is in **L1**, from the seek change, and it only strands a follower once L0 retention has aged out the L0 files that covered the same range.

### Related work, and how this differs

Searched 2026-09-10: I did not find an existing bug report for this failure mode, and no issue or PR text mentions `fillFollowGap` or `applyNewLTXFiles` outside #1102, which introduced them. The nearby work is related by remedy or symptom, **not** by mechanism:

- **#1270**: *different root cause, adjacent fix.* That is a TOCTOU race: the file is present when `CalcRestorePlan` lists it and is deleted before `OpenLTXFile` fetches it, so it is transient, recoverable by re-planning, and fails **loudly**. This is a file that is **durably absent at list time**, with no race, and the result is permanent and silent. The overlap is the suggested direction, not the cause. (That race is separately reachable from follow mode; I saw it once against a hosted backend as `open ltx file: file does not exist … consecutive_errors=1`. Different report.)
- **#1458**: similar symptom class, different component (the replica/sync side), which it notes was "a silent permanent compaction stall" before the gap-heal work.
- **#1460, #1262, #1263**: plausible *sources* of a durable absence, not duplicates of it.
- **#1268**: the exclusive-lock concern relevant to avenue (2) above.
- **#1514**: open, and the reason this is worth settling now; see the section above.
- **#1510, #1477, #1479**: the compaction cost that #1514 is addressing. Context, not duplicates.

### Possible avenues, cheapest first

I have implemented (1) and left (2) alone deliberately; the choice on (2) has consequences I cannot evaluate from outside. Scope notes so the cost is visible:

**1. Surface the condition.** The caller already has the machinery: `follow()` increments `consecutiveErrors`, logs, and continues, so nothing new is needed there. Two things worth knowing before touching it:

- The silent return happens at **two** exits in `applyNewLTXFiles`, not one: the gap-seen-in-the-listing path and the no-level-0-files-at-all path. A fix that handles only the first still leaves the second silent, and the reproduction drives the first.
- The open question is **cadence, not detection**. That branch runs every poll, so a naive error would log once per `-follow-interval` forever, trading silence for a permanent ERROR stream. Logging on transition into the state, or a distinguished error the loop handles once, seems more useful than logging every iteration.
- `TestReplica_ContextCancellationNoLogs` exists to assert that nothing logs during shutdown (issue #235), so whatever gets added has to stay quiet on cancellation.

**2. The bound that decides it.** `fillFollowGap` stops at `SnapshotLevel - 1` (`replica.go:1007`), so whether a follower can recover here is decided by that one comparison. I am deliberately not recommending a value for it: changing it admits level-9 snapshots to the gap-fill path, which means a full-database file could be applied where an incremental one would have served, on the normal path, under `applyLTXFile`'s exclusive lock, on databases far larger than any fixture, and that interacts with the lock concern in open PR #1268. Pointing at the line, not prescribing the change.

### A fix, and a regression test

I have opened a PR against this issue implementing avenue (1): it reports the condition once per position rather than once per poll, covers both silent exits, and stays quiet when the context is already cancelled so shutdown is still silent (#235). It does not change which gaps can be bridged: `fillFollowGap` still searches levels 1..8 and still does not consult snapshots, so it is strictly a visibility change and leaves avenue (2) open as a separate decision. Against PR #1514's scenario above it turns the silent stall into one line:

```text
level=WARN msg="follow: no ltx file available at current position, follower is not advancing"
  txid=0000000000000003 next_txid=0000000000000006
```

It is +32/−5 in `replica.go` with test coverage for both exits and the cancellation case. Blast radius is narrow: `fillFollowGap` has two call sites, both inside `applyNewLTXFiles`, which has one caller in `follow()`. Full package suite passes under `-race` on current main.

```bash
go test -race -count=1 -run 'TestReplica_ApplyNewLTXFiles' ./
```

If the out-parameter used to carry the stall position across polls is not to taste, I have an alternative that keeps the state on the replica instead; say the word and I will swap it.

Separately, and independent of that PR, I have a specification for an end-to-end regression test in the project's existing style (`TEST-SPEC.md` in the same gist: https://gist.github.com/bitfliq/8b3c982599a7af5fc1b48420d37a7c86). It is `TestReplica_Restore_Follow_CrashRecovery` (`replica_test.go:2362`) with one level-0 file removed via `c.LTXFilePath(0, parked+1, parked+1)` while follow is stopped, and the opposite assertion at the end: the follower never advances, `Restore` never returns, and a non-follow `Restore` from the same replica does reach the data. It includes the presence assertions needed to stop it silently becoming a no-op, and a positive control that omits the deletion. Note that PR #1514 rewrites the setup of all three `TestReplica_Restore_Follow_*` tests, so that spec wants rebasing onto whichever lands first.

### What I verified vs. inferred

Verified by running: everything under Logs, the 11/2 run counts, the trace behaviour, and the #1514 comparison table (both columns run from the same script, on `main` and on the PR branch). Verified by reading source at `main` (4ed7a308f627): the level bounds, the nil return, the dead-branch analysis, and the guard string. Inferred: that a lost L0 upload is a realistic way to reach this state. The reproduction creates the hole by deleting an object, and I have not observed one arise unaided. The silence seems worth reporting regardless of how the gap appears.

This was written with heavy usage of Claude in response to a problem I saw in the wild.

bitfliq:
## Summary

`restore -f` can stop making progress permanently while logging nothing. This makes the condition visible. It does not change which gaps can be bridged.

## Problem

When follow mode cannot bridge a gap from its current position, `applyNewLTXFiles` returns the unchanged TXID and a `nil` error. Back in `follow()`, a nil error with `newTXID == lastTXID` takes neither the error branch nor the `newTXID > lastTXID` branch, so nothing is logged on that iteration, or any later one, since nothing in the loop changes. The follower stays alive, keeps polling, and keeps serving reads from a database that never changes again. A restart resumes from the `-txid` sidecar and stops at the same TXID, while a plain `litestream restore` from the same replica succeeds.

A follower stalled this way, at `-follow-interval 1s`, with `LOG_LEVEL=trace`:

```
level=INFO msg="resuming follow mode from crash recovery" txid=0000000000000004
level=INFO msg="entering follow mode" txid=0000000000000004 interval=1s
level=INFO msg="follow mode stopped"
error/warning lines: 0
```

Sixty seconds of polling, frontier advancing the whole time, zero ERROR or WARN lines. Full investigation, a self-contained reproduction, and the level-by-level analysis are in #1515.

## Solution

Report the condition once per position rather than once per poll.

The position is monotonic, so keying the report on it logs when the state is *entered* rather than on every tick. Both silent exits are covered: the gap seen in the level 0 listing, and the no-level-0-files-at-all path. A fix to only the first leaves the second silent.

**Why a WARN and not a returned error.** `follow()` already increments `consecutiveErrors`, logs, and continues, but nothing consumes that counter: there is no threshold and no bail-out. So returning an error here would not change behaviour, it would only produce an ERROR line every `-follow-interval` forever, trading silence for a permanent error stream. `docs/PATTERNS.md` warns against "just logging" where continuing risks corrupting state; that is not this case, since a stalled follower is idle rather than advancing on bad data. If you would rather this terminate the follower or trip a threshold, that is a behaviour change I am happy to make instead; it just seemed like your call, not mine.

Reporting stays silent when the context is already cancelled, so shutdown remains quiet (#235, `TestReplica_ContextCancellationNoLogs`).

## Scope

**In scope:**
- Surfacing the unbridgeable-gap condition in follow mode, once per position.
- Test coverage for both silent exits and for the cancelled-context case.

**Not in scope:**
- Changing which gaps can be bridged. `fillFollowGap` still searches levels 1..8 and still does not consult snapshots. Whether that bound should change is a separate decision with real consequences (it would admit a full-database file where an incremental one would have served, under `applyLTXFile`'s exclusive lock, interacting with #1268), and I have deliberately not touched it.
- Any change to restore, compaction, or the replica sync path.

## Test Plan

```bash
go test -race -count=1 -run 'TestReplica_ApplyNewLTXFiles' ./
go test -race ./...
```

New: `TestReplica_ApplyNewLTXFiles_UnreachableGapWarnsOnce`, covering
- `GapInLevel0`: gap visible in the level 0 listing; warns once across repeated polls
- `NoLevel0Files`: level 0 compacted away entirely; the other silent exit
- `CancelledContext`: stays silent during shutdown

Verified on this branch: `gofmt`/`goimports` clean, `go vet ./...` clean, `staticcheck` clean, `go test -race ./...` passing.

## Related

- Fixes #1515
- Related to #1514. That PR makes the compaction ladder skip TXIDs a snapshot covers and relaxes `VerifyLevelConsistency` to accept the resulting hole, on the grounds that restore seeds from the snapshot. `follow` does not read the snapshot level, so on that branch a follower can reach this stall with nothing missing from the replica. Details in #1515.
- #1270 is a different root cause (a TOCTOU race that fails loudly), adjacent in remedy only.

---

This contribution was AI-assisted. I have reviewed the diff and am responsible for it.

### DIFF
--- a/replica.go
+++ b/replica.go
@@ -831,13 +831,14 @@ func (r *Replica) follow(ctx context.Context, outputPath string, lastTXID ltx.TX
 	defer ticker.Stop()
 
 	var consecutiveErrors int
+	var stallTXID ltx.TXID // position at which a follow gap was last reported
 	for {
 		select {
 		case <-ctx.Done():
 			r.Logger().Info("follow mode stopped")
 			return nil
 		case <-ticker.C:
-			newTXID, err := r.applyNewLTXFiles(ctx, f, lastTXID, pageSize)
+			newTXID, err := r.applyNewLTXFiles(ctx, f, lastTXID, pageSize, &stallTXID)
 			if err != nil {
 				if ctx.Err() != nil {
 					r.Logger().Info("follow mode stopped")
@@ -862,7 +863,7 @@ func (r *Replica) follow(ctx context.Context, outputPath string, lastTXID ltx.TX
 // applyNewLTXFiles polls for new LTX files and applies them to the database.
 // It starts from level 0 and falls back to higher levels if there are gaps
 // (e.g., level 0 files were compacted away).
-func (r *Replica) applyNewLTXFiles(ctx context.Context, f *os.File, afterTXID ltx.TXID, pageSize uint32) (ltx.TXID, error) {
+func (r *Replica) applyNewLTXFiles(ctx context.Context, f *os.File, afterTXID ltx.TXID, pageSize uint32, stallTXID *ltx.TXID) (ltx.TXID, error) {
 	currentTXID := afterTXID
 
 	// Poll level 0 for the most recent incremental files.
@@ -888,7 +889,7 @@ func (r *Replica) applyNewLTXFiles(ctx context.Context, f *os.File, afterTXID lt
 
 		// If there's a gap, try to fill it from higher compaction levels.
 		if info.MinTXID > currentTXID+1 {
-			bridgedTXID, err := r.fillFollowGap(ctx, f, currentTXID, info.MinTXID, pageSize)
+			bridgedTXID, err := r.fillFollowGap(ctx, f, currentTXID, info.MinTXID, pageSize, stallTXID)
 			if err != nil {
 				return closeLevel0(err)
 			}
@@ -899,6 +900,7 @@ func (r *Replica) applyNewLTXFiles(ctx context.Context, f *os.File, afterTXID lt
 				continue
 			}
 			if info.MinTXID > currentTXID+1 {
+				r.reportFollowGap(ctx, stallTXID, currentTXID, info.MinTXID)
 				return closeLevel0(nil)
 			}
 		}
@@ -925,7 +927,7 @@ func (r *Replica) applyNewLTXFiles(ctx context.Context, f *os.File, afterTXID lt
 	}
 
 	if !sawLevel0 {
-		bridgedTXID, err := r.fillFollowGap(ctx, f, currentTXID, currentTXID+1, pageSize)
+		bridgedTXID, err := r.fillFollowGap(ctx, f, currentTXID, currentTXID+1, pageSize, stallTXID)
 		if err != nil {
 			return currentTXID, err
 		}
@@ -1001,8 +1003,9 @@ func (r *Replica) applyLTXFile(ctx context.Context, f *os.File, info *ltx.FileIn
 
 // fillFollowGap attempts to bridge a gap in level 0 files by searching
 // higher compaction levels for a file that covers the missing TXID range.
-func (r *Replica) fillFollowGap(ctx context.Context, f *os.File, afterTXID ltx.TXID, gapMinTXID ltx.TXID, pageSize uint32) (ltx.TXID, error) {
+func (r *Replica) fillFollowGap(ctx context.Context, f *os.File, afterTXID ltx.TXID, gapMinTXID ltx.TXID, pageSize uint32, stallTXID *ltx.TXID) (ltx.TXID, error) {
 	currentTXID := afterTXID
+	var unreachableTXID ltx.TXID
 
 	for level := 1; level < SnapshotLevel; level++ {
 		itr, err := r.Client.LTXFiles(ctx, level, 0, false)
@@ -1025,6 +1028,9 @@ func (r *Replica) fillFollowGap(ctx context.Context, f *os.File, afterTXID ltx.T
 
 			// Skip if there's a gap at this level too.
 			if info.MinTXID > currentTXID+1 {
+				if unreachableTXID == 0 {
+					unreachableTXID = info.MinTXID
+				}
 				break
 			}
 
@@ -1060,7 +1066,28 @@ func (r *Replica) fillFollowGap(ctx context.Context, f *os.File, afterTXID ltx.T
 		}
 	}
 
+	// No level bridged the gap but files exist beyond our position, so the
+	// follower cannot advance without operator intervention.
+	if unreachableTXID != 0 {
+		r.reportFollowGap(ctx, stallTXID, currentTXID, unreachableTXID)
+	}
+
 	return currentTXID, nil
+}
+
+// reportFollowGap warns that follow mode has no path from its current position
+// to the next available LTX file. stallTXID holds the position last reported by
+// the caller's follow loop; the follower retries every interval and its position
+// is monotonic, so reporting per position logs the condition once when it is
+// entered rather than on every poll. Stays silent during shutdown (see #235).
+func (r *Replica) reportFollowGap(ctx context.Context, stallTXID *ltx.TXID, currentTXID, nextTXID ltx.TXID) {
+	if ctx.Err() != nil || *stallTXID == currentTXID {
+		return
+	}
+	*stallTXID = currentTXID
+
+	r.Logger().Warn("follow: no ltx file available at current position, follower is not advancing",
+		"txid", currentTXID, "next_txid", nextTXID)
 }
 
 // RestoreV3 restores from a v0.3.x format backup.
--- a/replica_internal_test.go
+++ b/replica_internal_test.go
@@ -11,6 +11,7 @@ import (
 	"log/slog"
 	"os"
 	"path/filepath"
+	"strings"
 	"testing"
 	"time"
 
@@ -63,7 +64,7 @@ func TestReplica_ApplyNewLTXFiles_FillGapWithOverlappingCompactedFile(t *testing
 	f := mustCreateWritableDBFile(t)
 	defer func() { _ = f.Close() }()
 
-	got, err := r.applyNewLTXFiles(context.Background(), f, 150, pageSize)
+	got, err := r.applyNewLTXFiles(context.Background(), f, 150, pageSize, new(ltx.TXID))
 	if err != nil {
 		t.Fatalf("apply new ltx files: %v", err)
 	}
@@ -107,7 +108,7 @@ func TestReplica_ApplyNewLTXFiles_LevelZeroEmptyFallsBackToCompaction(t *testing
 	f := mustCreateWritableDBFile(t)
 	defer func() { _ = f.Close() }()
 
-	got, err := r.applyNewLTXFiles(context.Background(), f, 10, pageSize)
+	got, err := r.applyNewLTXFiles(context.Background(), f, 10, pageSize, new(ltx.TXID))
 	if err != nil {
 		t.Fatalf("apply new ltx files: %v", err)
 	}
@@ -132,7 +133,7 @@ func TestReplica_ApplyNewLTXFiles_IteratorCloseError(t *testing.T) {
 	f := mustCreateWritableDBFile(t)
 	defer func() { _ = f.Close() }()
 
-	_, err := r.applyNewLTXFiles(context.Background(), f, 10, 4096)
+	_, err := r.applyNewLTXFiles(context.Background(), f, 10, 4096, new(ltx.TXID))
 	if err == nil {
 		t.Fatal("expected error")
 	}
@@ -541,5 +542,95 @@ func TestApplyLTXFile_MultiplePages(t *testing.T) {
 		if i > 0 && !bytes.Equal(readBuf, pages[i]) {
 			t.Fatalf("page %d data mismatch", i+1)
 		}
+	}
+}
+
+func TestReplica_ApplyNewLTXFiles_UnreachableGapWarnsOnce(t *testing.T) {
+	const pageSize = 4096
+
+	for _, tt := range []struct {
+		name         string
+		level0       []*ltx.FileInfo
+		level1       []*ltx.FileInfo
+		cancelled    bool
+		wantWarnings int
+	}{
+		// The gap is visible in the level 0 listing itself.
+		{
+			name:         "GapInLevel0",
+			level0:       []*ltx.FileInfo{{Level: 0, MinTXID: 200, MaxTXID: 200}},
+			wantWarnings: 1,
+		},
+		// Level 0 has been compacted away entirely; the gap is only visible at
+		// a higher level.
+		{
+			name:         "NoLevel0Files",
+			level1:       []*ltx.FileInfo{{Level: 1, MinTXID: 200, MaxTXID: 250}},
+			wantWarnings: 1,
+		},
+		// Shutdown must stay quiet (see TestReplica_ContextCancellationNoLogs).
+		{
+			name:      "CancelledContext",
+			level0:    []*ltx.FileInfo{{Level: 0, MinTXID: 200, MaxTXID: 200}},
+			cancelled: true,
+		},
+	} {
+		t.Run(tt.name, func(t *testing.T) {
+			client := &followTestReplicaClient{}
+			client.LTXFilesFunc = func(_ context.Context, level int, seek ltx.TXID, _ bool) (ltx.FileIterator, error) {
+				var all []*ltx.FileInfo
+				switch level {
+				case 0:
+					all = tt.level0
+				case 1:
+					all = tt.level1
+				}
+
+				infos := make([]*ltx.FileInfo, 0, len(all))
+				for _, info := range all {
+					if info.MinTXID >= seek {
+						infos = append(infos, info)
+					}
+				}
+				return ltx.NewFileInfoSliceIterator(infos), nil
+			}
+
+			var logBuf bytes.Buffer
+			db := NewDB(filepath.Join(t.TempDir(), "follower.db"))
+			db.Logger = slog.New(slog.NewTextHandler(&logBuf, &slog.HandlerOptions{Level: slog.LevelDebug}))
+
+			r := NewReplicaWithClient(db, client)
+			f := mustCreateWritableDBFile(t)
+			defer func() { _ = f.Close() }()
+
+			ctx, cancel := context.WithCancel(context.Background())
+			defer cancel()
+			if tt.cancelled {
+				cancel()
+			}
+
+			// Poll several times from the same position, as the follow loop does
+			// on every interval, carrying the same stall state it carries.
+			// Nothing bridges the gap, so the position never advances and the
+			// condition must be reported exactly once.
+			stallTXID := new(ltx.TXID)
+			for i := 0; i < 3; i++ {
+				got, err := r.applyNewLTXFiles(ctx, f, 100, pageSize, stallTXID)
+				if err != nil {
+					t.Fatalf("apply new ltx files: %v", err)
+				}
+				if got != 100 {
+					t.Fatalf("txid=%s, want %s", got, ltx.TXID(100))
+				}
+			}
+
+			logs := logBuf.String()
+			if n := strings.Count(logs, "level=WARN"); n != tt.wantWarnings {
+				t.Fatalf("warnings=%d, want %d, logs=%q", n, tt.wantWarnings, logs)
+			}
+			if tt.wantWarnings > 0 && !strings.Contains(logs, "next_txid=00000000000000c8") {
+				t.Fatalf("expected next txid in warning, logs=%q", logs)
+			}
+		})
 	}
 }


bitfliq:
Please Note that if https://github.com/benbjohnson/litestream/pull/1514 gets merged, this is likely unnecessary and should be closed.

